PR #4, opened 2019-05-16
Update wgpu-rs to reflect latest wgpu API

kyren:
Not sure if everything here is correct, especially `RenderPassColorAttachmentDescriptor` and the `resolve_target` field.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,13 +14,13 @@ pub use wgn::{
     BlendDescriptor,
     BlendFactor,
     BlendOperation,
-    BorderColor,
+    BufferAddress,
     BufferDescriptor,
     BufferMapAsyncStatus,
-    BufferUsageFlags,
+    BufferUsage,
     Color,
     ColorStateDescriptor,
-    ColorWriteFlags,
+    ColorWrite,
     CommandEncoderDescriptor,
     CompareFunction,
     CullMode,
@@ -32,17 +32,17 @@ pub use wgn::{
     FrontFace,
     IndexFormat,
     InputStepMode,
+    Limits,
     LoadOp,
     Origin3d,
     PowerPreference,
     PrimitiveTopology,
     RasterizationStateDescriptor,
-    RenderPassColorAttachmentDescriptor,
     RenderPassDepthStencilAttachmentDescriptor,
     SamplerDescriptor,
-    ShaderAttributeIndex,
+    ShaderLocation,
     ShaderModuleDescriptor,
-    ShaderStageFlags,
+    ShaderStage,
     StencilOperation,
     StencilStateFaceDescriptor,
     StoreOp,
@@ -51,7 +51,7 @@ pub use wgn::{
     TextureDescriptor,
     TextureDimension,
     TextureFormat,
-    TextureUsageFlags,
+    TextureUsage,
     TextureViewDescriptor,
     TextureViewDimension,
     VertexAttributeDescriptor,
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -161,7 +160,7 @@ pub struct Queue<'a> {
 pub enum BindingResource<'a> {
     Buffer {
         buffer: &'a Buffer,
-        range: Range<u32>,
+        range: Range<u64>,
```
kvark: we should use the [BufferAddress](https://github.com/gfx-rs/wgpu/blob/a056d2478a3e34fa20071957af89fc6eda7a313d/wgpu-native/src/lib.rs#L53) alias for these

src/lib.rs
```diff
@@ -726,8 +733,9 @@ impl CommandEncoder {
         let colors = desc
             .color_attachments
             .iter()
-            .map(|ca| RenderPassColorAttachmentDescriptor {
+            .map(|ca| wgn::RenderPassColorAttachmentDescriptor {
                 attachment: ca.attachment.id,
+                resolve_target: ca.resolve_target.map(|v| &v.id as *const _).unwrap_or(ptr::null()),
```
kvark: nit: map_or()

src/lib.rs
```diff
@@ -216,11 +215,19 @@ pub struct ComputePipelineDescriptor<'a> {
...
 pub struct RenderPassDescriptor<'a> {
-    pub color_attachments: &'a [RenderPassColorAttachmentDescriptor<&'a TextureView>],
+    pub color_attachments: &'a [RenderPassColorAttachmentDescriptor<'a, &'a TextureView>],
     pub depth_stencil_attachment:
         Option<RenderPassDepthStencilAttachmentDescriptor<&'a TextureView>>,
 }
 
+pub struct RenderPassColorAttachmentDescriptor<'a, T> {
+    pub attachment: T,
+    pub resolve_target: Option<&'a TextureView>,
```
kvark: let's use `Option<T>` here

src/lib.rs
```diff
@@ -441,7 +448,7 @@ impl Device {
                         module: desc.vertex_stage.module.id,
                         entry_point: vertex_entry_point.as_ptr(),
                     },
-                    fragment_stage: wgn::PipelineStageDescriptor {
+                    fragment_stage: &wgn::PipelineStageDescriptor {
                         module: desc.fragment_stage.module.id,
```
kvark: we should make `desc.fragment_stage` to also be an `Option` now

src/lib.rs
kyren: yep, derp.

src/lib.rs
kyren: Are there other instances besides this one that should use the BufferAddress alias?

src/lib.rs
kyren: Yeah that should have been obvious, sorry!

src/lib.rs
```diff
@@ -418,7 +426,20 @@ impl Device {
...
+        let (_fragment_entry_point, fragment_stage) = if let Some(fragment_stage) = &desc.fragment_stage {
+            let fragment_entry_point = CString::new(fragment_stage.entry_point).unwrap();
+            let fragment_stage = wgn::PipelineStageDescriptor {
+                module: fragment_stage.module.id,
+                entry_point: fragment_entry_point.as_ptr(),
+            };
+            (fragment_entry_point, Some(fragment_stage))
+        } else {
+            (CString::default(), None)
+        };
```
kyren: I'm uncomfortable with how not-obviously-safe this is, but I couldn't think of a better way :/

src/lib.rs
kyren: Yeah okay, it's a lot more, it would have been obvious had I actually looked at wgpu more closely!

kyren:
> sorry about the inconvenience

Not inconvenient at all!

src/lib.rs
kyren: Actually just %s/u64/BufferAddress/g was correct, I made it harder than it had to be :P

PR #16, opened 2019-05-28
Fix tests and run in CI

rukai:
Fixes tests and enables running tests with vulkan on windows and linux.
If there are other platforms + backends I should add let me know.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,3 +8,8 @@ os:
 branches:
   except:
   - staging.tmp
+
+script:
+  - cargo test
+  - if [[ $TRAVIS_OS_NAME == "linux" ]]; then cargo test --no-run --features vulkan; fi
+  - if [[ $TRAVIS_OS_NAME == "windows" ]]; then cargo test --no-run --features vulkan; fi
--- a/tests/multithreaded_compute.rs
+++ b/tests/multithreaded_compute.rs
@@ -13,16 +13,17 @@ fn multithreaded_compute() {
         thread::spawn(move || {
             let numbers = vec!(100, 100, 100);
 
-            let size = (numbers.len() * std::mem::size_of::<u32>()) as u32;
+            let size = (numbers.len() * std::mem::size_of::<u32>()) as wgpu::BufferAddress;
 
             let instance = wgpu::Instance::new();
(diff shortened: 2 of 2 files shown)


rukai:
Oh huh, vulkano runs tests on CI so I assumed I could run run this on CI as well.
I guess vulkanos dont actually touch vulkan then.

aloucks:
Vulkano's tests all bail out early if they can't enumerate a physical device. I thought vulkano had done something magical until I actually looked at how the tests worked:

https://github.com/vulkano-rs/vulkano/blob/master/vulkano/src/device/mod.rs#L822

The VMs they run on don't have GPU. If you install the vulkan sdk, `vulkaninfo` will bail out with an initialization failed message (tested on azure and assuming the same on travis, given other reports on the matter). I think I saw somewhere that travis has a limited amount of GPU for private repos, but I haven't confirmed it nor have I found any additional evidence other than a single forum posting.

I'm also struggling to find a way to test vulkan/gpu code in CI. From what I can tell, circle CI and azure-pipelines allow you to bring-your-own-agent (azure) or provision VMs on demand (in the case of Circle). I haven't tried either of these options yet, but from what I can tell you need a GPU enabled VM from [AWS](https://aws.amazon.com/ec2/instance-types/#Accelerated_Computing) or N-series from azure. 

I'd happily eat my words if you find an easier way :)

rukai:
Thanks for the information!
Sounds like the best thing to do is to use `cargo test --no-run`. I think silently doing nothing in the tests is likely to trick people into thinking they are passing for real.

azriel91:
Hiya, I've been ~~slaving~~ chipping away at trying to get CI to work with Vulkan for the past 9 days, and thought my investigations would be useful to leave here.

* Windows: It "just worked<TM>" with my Windows laptop with nvidia GPU as a CI agent on gitlab.
* Linux: Haven't gotten this to work yet. Here's a (crazy) gist containing my investigation to get Xorg to run in the background on my laptop (with nvidia GPU): https://gist.github.com/azriel91/a236daafa4e5be3b122745239d03596c

rukai:
Bump! Would be nice to get this merged before it breaks again.

kvark:
bors retry

PR #25, opened 2019-06-18
Add MSAA example

rukai:
This example is the same as the triangle example except:
*    The render pipeline is created with a sample_count of 4
*    A new texture with a sample_count of 4 is created and set as the color_attachment instead of the swapchain
*    The swapchain is now specified as a resolve_target

It doesn't work yet as wgpu-native doesn't handle MSAA yet.

### DIFF
--- a/examples/msaa-linelist/main.rs
+++ b/examples/msaa-linelist/main.rs
@@ -1,0 +1,212 @@
+//! The parts of this example enabling MSAA are:
+//! *    The render pipeline is created with a sample_count > 1.
+//! *    A new texture with a sample_count > 1 is created and set as the color_attachment instead of the swapchain.
+//! *    The swapchain is now specified as a resolve_target.
+//!
+//! The parts of this example enabling LineList are:
+//! *   Set the primitive_topology to PrimitiveTopology::LineList.
+//! *   Vertices and Indices describe the two points that make up a line.
+
+#[path = "../framework.rs"]
+mod framework;
+
+#[derive(Clone, Copy)]
+struct Vertex {
+    _pos:   [f32; 2],
+    _color: [f32; 4],
+}
+
+struct Example {
+    vs_module: wgpu::ShaderModule,
+    fs_module: wgpu::ShaderModule,
+    pipeline_layout: wgpu::PipelineLayout,
+    pipeline: wgpu::RenderPipeline,
+    multisampled_framebuffer: wgpu::TextureView,
+    vertex_buffer: wgpu::Buffer,
+    vertex_count: u32,
+    rebuild_pipeline: bool,
+    sample_count: u32,
+    sc_desc: wgpu::SwapChainDescriptor,
+}
+
+impl Example {
+    fn create_pipeline(device: &wgpu::Device, sc_desc: &wgpu::SwapChainDescriptor, vs_module: &wgpu::ShaderModule, fs_module: &wgpu::ShaderModule, pipeline_layout: &wgpu::PipelineLayout, sample_count: u32) -> wgpu::RenderPipeline {
+        println!("sample_count: {}", sample_count);
+        device.create_render_pipeline(&wgpu::RenderPipelineDescriptor {
+            layout: &pipeline_layout,
+            vertex_stage: wgpu::PipelineStageDescriptor {
+                module: vs_module,
+                entry_point: "main",
+            },
+            fragment_stage: Some(wgpu::PipelineStageDescriptor {
+                module: fs_module,
+                entry_point: "main",
+            }),
+            rasterization_state: wgpu::RasterizationStateDescriptor {
+                front_face: wgpu::FrontFace::Ccw,
+                cull_mode: wgpu::CullMode::None,
+                depth_bias: 0,
+                depth_bias_slope_scale: 0.0,
+                depth_bias_clamp: 0.0,
+            },
+            primitive_topology: wgpu::PrimitiveTopology::LineList,
+            color_states: &[wgpu::ColorStateDescriptor {
+                format: sc_desc.format,
+                color_blend: wgpu::BlendDescriptor::REPLACE,
+                alpha_blend: wgpu::BlendDescriptor::REPLACE,
+                write_mask: wgpu::ColorWrite::ALL,
+            }],
+            depth_stencil_state: None,
+            index_format: wgpu::IndexFormat::Uint16,
+            vertex_buffers: &[wgpu::VertexBufferDescriptor {
+                stride: std::mem::size_of::<Vertex>() as wgpu::BufferAddress,
+                step_mode: wgpu::InputStepMode::Vertex,
+                attributes: &[
+                    wgpu::VertexAttributeDescriptor {
+                        format: wgpu::VertexFormat::Float2,
+                        offset: 0,
+                        shader_location: 0,
+                    },
+                    wgpu::VertexAttributeDescriptor {
+                        format: wgpu::VertexFormat::Float4,
+                        offset: 2 * 4,
+                        shader_location: 1,
+                    },
+                ],
+            }],
+            sample_count,
+        })
+    }
+
+    fn create_multisampled_framebuffer(device: &wgpu::Device, sc_desc: &wgpu::SwapChainDescriptor, sample_count: u32) -> wgpu::TextureView {
+        let multisampled_texture_extent = wgpu::Extent3d {
+            width: sc_desc.width,
+            height: sc_desc.height,
+            depth: 1,
+        };
+        let multisampled_frame_descriptor = &wgpu::TextureDescriptor {
+            size: multisampled_texture_extent,
+            array_layer_count: 1,
+            mip_level_count: 1,
+            sample_count: sample_count,
+            dimension: wgpu::TextureDimension::D2,
+            format: sc_desc.format,
+            usage: wgpu::TextureUsage::OUTPUT_ATTACHMENT,
+        };
+
+        device.create_texture(multisampled_frame_descriptor).create_default_view()
+    }
+}
+
+impl framework::Example for Example {
+    fn init(sc_desc: &wgpu::SwapChainDescriptor, device: &mut wgpu::Device) -> Self {
+        println!("Press left/right arrow keys to change sample_count.");
+        let sample_count = 2;
+
+        let vs_bytes = framework::load_glsl(include_str!("shader.vert"), framework::ShaderStage::Vertex);
+        let fs_bytes = framework::load_glsl(include_str!("shader.frag"), framework::ShaderStage::Fragment);
+        let vs_module = device.create_shader_module(&vs_bytes);
+        let fs_module = device.create_shader_module(&fs_bytes);
+
+        let pipeline_layout = device.create_pipeline_layout(&wgpu::PipelineLayoutDescriptor {
+            bind_group_layouts: &[],
+        });
+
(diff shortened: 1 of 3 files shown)


kvark:
This could be a good start! We need to make sure the examples don't repeat themselves too much and try to cover more ground. For this case, perhaps we could think of some workload involving a bunch of lines to draw? That would show MSAA as well as non-triangular primitive topologies.

kvark:
Also, let's use the framework module to avoid boilerplate repetition, like cube and shadow examples do. I don't think there needs to be more than hello-triangle and hello-compute for barebone setup.

rukai:
Ah right, I missed the framework.rs. I would appreciate it if I could get feedback as is, about if I am correctly using the wgpu API for MSAA. Then I'll go ahead and make the changes you have requested.

examples/msaa/main.rs
```diff
@@ -0,0 +1,181 @@
...
+            usage: wgpu::TextureUsage::OUTPUT_ATTACHMENT,
+        };
+
+        let multisampled_frame = device.create_texture(multisampled_frame_descriptor);
+        let mut encoder =
+            device.create_command_encoder(&wgpu::CommandEncoderDescriptor { todo: 0 });
+        {
+            let mut rpass = encoder.begin_render_pass(&wgpu::RenderPassDescriptor {
+                color_attachments: &[wgpu::RenderPassColorAttachmentDescriptor {
+                    attachment: &multisampled_frame.create_default_view(),
```
aloucks: Should the texture and view be created every frame or only when the window size changes?

kvark: It is a good question, and it's currently unclear because Google is pushing for caching all the things by the implementation, i.e. the device would remember if there is already a good view for a given texture and return it again.
I'd prefer if we didn't rely on this hypothetical behavior :)

kvark:
> I would appreciate it if I could get feedback as is, about if I am correctly using the wgpu API for MSAA

I think it's correct. We might be missing some implementation bits (like https://github.com/gfx-rs/gfx/pull/2813), but it would be easy to fix them given this test case :)

rukai:
I changed the example to use PrimitiveTopology::LineList and to utilize the framework module.

I also added runtime configurable sample_count.

Screenshot from 2019-06-20 00-06-32

If you want to test the rendering then you will need to disable the MSAA logic by setting the swapchain as the color attachment rather then the resolve_target and you will need to set the sample_count to 1.

kvark:
@rukai this is a great test case, thank you!

examples/msaa-linelist/main.rs
```diff
@@ -0,0 +1,235 @@
...
+                0 + i * 2,
+                1 + i * 2,
+            ]);
+        }
+
+        let vertex_buffer = device
+            .create_buffer_mapped(vertex_data.len(), wgpu::BufferUsage::VERTEX)
+            .fill_from_slice(&vertex_data);
+        let index_buffer = device
+            .create_buffer_mapped(index_data.len(), wgpu::BufferUsage::VERTEX)
```
kvark: should be `INDEX`

examples/msaa-linelist/main.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    fn update(&mut self, event: wgpu::winit::WindowEvent, device: &mut wgpu::Device) {
+        match event {
+            wgpu::winit::WindowEvent::KeyboardInput { input, .. } => {
+                if let wgpu::winit::ElementState::Pressed = input.state {
+                    match input.virtual_keycode {
+                        Some(wgpu::winit::VirtualKeyCode::Left) => {
+                            if self.sample_count >= 2 {
+                                self.sample_count = self.sample_count >> 1;
+                                self.pipeline = Example::pipeline(device, &self.vs_module, &self.fs_module, &self.pipeline_layout, self.sample_count);
```
kvark: could we keep `update` function signature unchanged (i.e. not getting the `Device` parameter) by just storing the `requested_sample_count` variable and updating the actual `pipeline` at the start of `render`?

examples/msaa-linelist/main.rs
```diff
@@ -0,0 +1,235 @@
...
+                            }
+                        }
+                        Some(wgpu::winit::VirtualKeyCode::Right) => {
+                            if self.sample_count <= 32 {
```
kvark: I don't think we want 64 samples supported (D3D12/Metal limitations...). Let's change this to `<32`

examples/msaa-linelist/main.rs
```diff
@@ -0,0 +1,235 @@
...
+        self.multisampled_framebuffer = Example::multisampled_framebuffer(device, sc_desc.width, sc_desc.height, self.sample_count);
+    }
+
+    fn render(&mut self, frame: &wgpu::SwapChainOutput, device: &mut wgpu::Device) {
+        let mut encoder =
+            device.create_command_encoder(&wgpu::CommandEncoderDescriptor { todo: 0 });
+        {
+            let mut rpass = encoder.begin_render_pass(&wgpu::RenderPassDescriptor {
+                color_attachments: &[wgpu::RenderPassColorAttachmentDescriptor {
+                    attachment: &self.multisampled_framebuffer.create_default_view(),
```
kvark: let's keep this view cached in `self`

examples/msaa-linelist/main.rs
```diff
@@ -0,0 +1,235 @@
...
+                    resolve_target: Some(&frame.view),
+                    load_op: wgpu::LoadOp::Clear,
+                    store_op: wgpu::StoreOp::Store,
+                    clear_color: wgpu::Color {
```
kvark: let's use `Color::BLACK`

examples/msaa-linelist/main.rs
```diff
@@ -0,0 +1,235 @@
...
+            rasterization_state: wgpu::RasterizationStateDescriptor {
+                front_face: wgpu::FrontFace::Ccw,
+                cull_mode: wgpu::CullMode::None,
+                depth_bias: 0,
+                depth_bias_slope_scale: 0.0,
+                depth_bias_clamp: 0.0,
+            },
+            primitive_topology: wgpu::PrimitiveTopology::LineList,
+            color_states: &[wgpu::ColorStateDescriptor {
+                format: wgpu::TextureFormat::Bgra8Unorm,
```
kvark: let's use the `format` provided to use in `SwapChainDescriptor`

examples/msaa-linelist/main.rs
```diff
@@ -0,0 +1,235 @@
...
+    vertex_buffer: wgpu::Buffer,
+    index_buffer: wgpu::Buffer,
+    index_count: u32,
+    sample_count: u32,
+    width: u32,
+    height: u32,
+}
+
+impl Example {
+    fn pipeline(device: &mut wgpu::Device, vs_module: &wgpu::ShaderModule, fs_module: &wgpu::ShaderModule, pipeline_layout: &wgpu::PipelineLayout, sample_count: u32) -> wgpu::RenderPipeline {
```
kvark: let's rename this to start with a verb, i.e. `create_pipeline`

examples/msaa-linelist/main.rs
```diff
@@ -0,0 +1,235 @@
...
+                write_mask: wgpu::ColorWrite::ALL,
+            }],
+            depth_stencil_state: None,
+            index_format: wgpu::IndexFormat::Uint16,
+            vertex_buffers: &[wgpu::VertexBufferDescriptor {
+                stride: std::mem::size_of::<Vertex>() as wgpu::BufferAddress,
+                step_mode: wgpu::InputStepMode::Vertex,
+                attributes: &[
+                    wgpu::VertexAttributeDescriptor {
+                        format: wgpu::VertexFormat::Float4,
```
kvark: should be `Float2`

examples/msaa-linelist/main.rs
```diff
@@ -0,0 +1,235 @@
...
+                        offset: 2 * 4,
+                        shader_location: 1,
+                    },
+                ],
+            }],
+            sample_count,
+        })
+    }
+
+    fn multisampled_framebuffer(device: &mut wgpu::Device, width: u32, height: u32, sample_count: u32) -> wgpu::Texture {
```
kvark: similarly, let's rename this to start with a verb

examples/msaa-linelist/main.rs
kvark: here and in the other function we need to only borrow `&Device` immutably

examples/msaa-linelist/main.rs
```diff
@@ -0,0 +1,235 @@
...
+            height,
+            depth: 1
+        };
+        let multisampled_frame_descriptor = &wgpu::TextureDescriptor {
+            size: multisampled_texture_extent,
+            array_layer_count: 1,
+            mip_level_count: 1,
+            sample_count: sample_count,
+            dimension: wgpu::TextureDimension::D2,
+            format: wgpu::TextureFormat::Bgra8Unorm,
```
kvark: same here - need to use format given to us in the swapchain descriptor

examples/msaa-linelist/main.rs
```diff
@@ -0,0 +1,235 @@
...
+    fn init(sc_desc: &wgpu::SwapChainDescriptor, device: &mut wgpu::Device) -> Self {
+        println!("Press left/right arrow keys to change sample_count.");
+        let sample_count = 2;
+
+        let vs_bytes = framework::load_glsl(include_str!("shader.vert"), framework::ShaderStage::Vertex);
+        let fs_bytes = framework::load_glsl(include_str!("shader.frag"), framework::ShaderStage::Fragment);
+        let vs_module = device.create_shader_module(&vs_bytes);
+        let fs_module = device.create_shader_module(&fs_bytes);
+
+        let bind_group_layout =
```
kvark: let's not have any bindings or bind groups at all, since they are not needed! That is also a good property to show off in this example

examples/msaa-linelist/main.rs
```diff
@@ -0,0 +1,235 @@
...
+            vertex_data.push(Vertex {
+                _pos: [0.0, 0.0],
+                _color: [1.0, -sin, cos, 1.0],
+            });
+            vertex_data.push(Vertex {
+                _pos: [1.0 * cos, 1.0 * sin],
+                _color: [sin, -cos, 1.0, 1.0],
+            });
+
+            index_data.extend(&[
```
kvark: let's not have index buffer at all, since it's not saving us anything here

examples/msaa-linelist/shader.vert
```diff
@@ -0,0 +1,10 @@
+#version 450
+
+layout(location = 0) in vec2 a_Pos;
+layout(location = 1) in vec4 a_color;
```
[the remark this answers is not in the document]
rukai: Are you just asking for the capitalization? I'm not sure what the first character is supposed to be.

rukai:
Thanks for the fantastic feedback!
I even learned something to apply to my own project.

I've made a few clarification questions and addressed everything else.

examples/msaa-linelist/shader.vert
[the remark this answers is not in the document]
kvark: yes, capitalization of the letter after `_`. The first character here stands for "Attribute", fyi

examples/msaa-linelist/main.rs
kvark: Let's leave it there. The logic here is: having no bind groups in a pipeline layout is an *edge* case. So it needs to be shown somewhere, but not necessarily in the hello-triangle example.

rukai: Sounds good to me. I reverted the hello-triangle change.

PR #28, opened 2019-06-22
msaa-line example fixes

rukai:
This PR fixes the msaa-line example in addition to the fixes in https://github.com/gfx-rs/wgpu/pull/235

If https://github.com/gfx-rs/gfx/pull/2853 is merged first we can remove the crates.io patch.


rukai:
Updated the commit, should be good to go.

Issue #7, opened 2019-05-19
device should implement debug


m4b:
e.g., 
```
        let mut device = adapter.request_device(&wgpu::DeviceDescriptor {
            extensions: wgpu::Extensions {
                anisotropic_filtering: false,
            },
            limits: wgpu::Limits::default(),
        });
info!("Got device: {:?}", device)
```
should do something (as opposed to not compiling); ideally, if possible, print the vendor info of the device that was obtained?

E.g., something like:
```
Device: Intel(R) HD Graphics 620 (Kaby Lake GT2) (type: IntegratedGpu)
```

kvark:
Yes, that would be good.

GabrielMajeri:
Is this still an issue since `Adapter::get_info` exists?

kvark:
indeed, I think we can close this

PR #29, opened 2019-06-23
impl Debug for Device

paulkernfeld:
Currently this only prints the device ID

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ gl = ["wgn/gfx-backend-gl"]
 
 [dependencies]
 #TODO: only depend on the published version
-wgn = { package = "wgpu-native", version = "0.2.6", features = ["local", "window-winit"], git = "https://github.com/gfx-rs/wgpu", rev = "a667d50d01c3df03b8540c523278e111da7fc82a" }
+wgn = { package = "wgpu-native", version = "0.2.6", features = ["local", "window-winit"], git = "https://github.com/gfx-rs/wgpu", rev = "dbef9f397eda87b82ae1691b2f7e8ba0f3e2e5d2" }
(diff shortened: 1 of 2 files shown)


paulkernfeld:
This was requested in #7. Unfortunately, I'm having some trouble figuring out how to print out more useful information, such as the make and model of the video card. If someone could point me in the right direction, I'd be happy to give it a shot.

kvark:
As with any wgpu-rs object, they are simply indices into the tables owned by wgpu-native's `HUB`. If the implementation of `Debug` is meant to work equally well with wgpu-native and under WASM in the browser, it needs to only rely on the WebGPU C API and nothing more. That C API is most of the time asynchronous - you never expect the result right away, unless it's some opaque handle produced by the clinet...

We should probably just add a method returning [adapter info](https://github.com/gfx-rs/wgpu/blob/f8df8d40326670213008318d87baa62319ad463f/wgpu-native/src/instance.rs#L230) to `wgpu-native` (in [this repo](https://github.com/gfx-rs/wgpu)) and use it here.

paulkernfeld:
Okay, thanks for the explanation!

Is it bad to call an async method inside of the `Debug` implementation? My intuition is that this could cause some unexpected behavior for end users. If this is a legitimate concern, perhaps it could be solved by retrieving the device info when the adapter is created and caching it.

kvark:
I think we shouldn't try making `Debug` nice for the case of WASM target. We should at least make sure it's meaningful when building on wgpu-native, which would be achieved by providing an extra public function there that gets the adapter info (to be used here in wgpu-rs in `Debug` implementation).

paulkernfeld:
Closing in favor of #89

Issue #19, opened 2019-06-04
BufferMapAsyncResult should be bound to a gfx-rs Pod like trait.


rukai:
gfx-rs prevented UB via limiting types to the Pod (plain old datatype) trait https://docs.rs/gfx/0.18.1/gfx/traits/trait.Pod.html
We should create a similar trait bounds for BufferMapAsyncResult.

paulkernfeld:
This can probably already be closed, but #46 addresses this in a more rigorous way using the `zerocopy` crate.

PR #30, opened 2019-06-24
Add Pod trait bound for all BufferMapAsyncResult data

paulkernfeld:


paulkernfeld:
This is for #19. I was originally hoping to do like `pub struct BufferAsyncMapping<T: Pod> { ... }`, but `Pod` is not implemented for `&[T]`. If there's a better solution, please let me know!

kvark:
Thank you for the PR!
@grovesNL noted that we shouldn't be exposing `Pod` in gfx-rs at all, since the API and the backends don't depend on it (it's purely a thing for the user). So we may need to move `Pod` out and put it inside `wgpu-rs` itself at some point.
In the meantime, let's proceed :)
bors r+

PR #32, opened 2019-07-05
Add conversion matrix from OpenGL to wgpu

dragly:
The matrices in the examples are given in an OpenGL-like coordinate system,
while a Vulkan-like coordinate system is used by wgpu. This was previously
partially corrected in the shader and by flipping the up axis of the camera,
but left the x-axis mirrored in the final result.

This change adds a conversion matrix to framework.rs that can be used
to convert from OpenGL to wgpu. This also allows us to set the
winding-order to counter-clockwise, which matches the ordering in the data.

### DIFF
--- a/examples/framework.rs
+++ b/examples/framework.rs
@@ -1,5 +1,13 @@
 use log::info;
 
+#[cfg_attr(rustfmt, rustfmt_skip)]
+pub const OPENGL_TO_WGPU_MATRIX: cgmath::Matrix4<f32> = cgmath::Matrix4::new(
+    1.0, 0.0, 0.0, 0.0,
+    0.0, -1.0, 0.0, 0.0,
+    0.0, 0.0, 0.5, 0.0,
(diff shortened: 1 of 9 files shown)


examples/framework.rs
```diff
@@ -30,6 +30,14 @@ pub fn load_glsl(code: &str, stage: ShaderStage) -> Vec<u8> {
     spv
 }
 
+pub fn opengl_to_wgpu_matrix() -> cgmath::Matrix4<f32> {
+    // converts from -1,1 Z to 0,1 Z and flips Y
+    cgmath::Matrix4::new(1.0, 0.0, 0.0, 0.0,
```
kvark: can we have that as a constant instead of a function?

dragly: Definitely! Updated now.

dragly:
Thanks! I completely missed that the view-projection matrix was also used by the lights in the forward pass.

PR #38, opened 2019-07-21
Add an example showing how to capture a render

paulkernfeld:
I had wanted to use wgpu-rs to generate `.png` images rather than displaying to the screen, so I coded up this prototype as an example to others who might want to do something similar.

Possible changes:
- Create an actual `.png` or other image
- Convert this into a unit test

I'm happy to make other modifications to this as well.


examples/capture/main.rs
```diff
@@ -0,0 +1,98 @@
...
+        (std::mem::size_of::<u32>() as u32 * size * size) as u64,
+        |result: wgpu::BufferMapAsyncResult<&[u8]>| {
+            for pixel in result.unwrap().data.chunks(std::mem::size_of::<u32>()) {
+                let expected_b = 0;
+                let expected_g = 0;
+                let expected_r = 255;
+                let expected_a = 255;
+                assert_eq!(pixel, [expected_b, expected_g, expected_r, expected_a]);
+            }
+        },
```
rukai: I think we need a `device.poll(true);` here to ensure the map_read_async runs.

examples/capture/main.rs
```diff
@@ -0,0 +1,98 @@
...
+        limits: wgpu::Limits::default(),
+    });
+
+    // Rendered image is 256×256 with 32-bit BGRA color
+    let size = 256u32;
+
+    // The output buffer lets us retrieve the data as an array
+    let output_buffer = device.create_buffer(&wgpu::BufferDescriptor {
+        size: (size * size * std::mem::size_of::<u32>() as u32) as u64,
+        usage: wgpu::BufferUsage::MAP_READ,
```
rukai: I think we also need `BufferUsage::TRANSFER_DEST` as we are performing a `copy_texture_to_buffer` on this buffer.

examples/capture/main.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    // The render pipeline renders data into this texture
+    let texture = device.create_texture(&wgpu::TextureDescriptor {
+        size: texture_extent,
+        array_layer_count: 1,
+        mip_level_count: 1,
+        sample_count: 1,
+        dimension: wgpu::TextureDimension::D2,
+        format: wgpu::TextureFormat::Bgra8Unorm,
+        usage: wgpu::TextureUsage::NONE,
```
rukai: This needs `TextureUsage::OUTPUT_ATTACHMENT | TextureUsage::TRANSFER_SRC`.
OUTPUT_ATTACHMENT is needed because we are rendering to this texture, TRANSFER_SRC is needed as it is used in the `copy_texture_buffer`

rukai:
Hi, this is something I have done before, you might want to refer to https://github.com/rukai/brawllib_rs/blob/master/src/renderer.rs
I used it to make these gifs https://github.com/gfx-rs/wgpu-rs/blob/master/etc/brawl-attack.gif

I think the example should really draw to an actual image, using e.g. the image crate.

Given that we have two people who required this functionality, I think this example will be useful!

examples/capture/main.rs
paulkernfeld: Okay, I added this, hopefully in the correct place.

examples/capture/main.rs
paulkernfeld: Done

examples/capture/main.rs
paulkernfeld: Done

paulkernfeld:
Okay, sounds good! I just set this up to write out to a PNG image and confirmed that it works correctly on OSX.

I noticed that `tests/multithreaded_compute` was also failing to call `device.poll`, so I added it there as well. FWIW this caused the test to start passing on my system.

examples/capture/main.rs
```diff
@@ -1,72 +1,74 @@
 /// This example shows how to capture an image by rendering it to a texture, copying the texture to
 /// a buffer, and retrieving it from the buffer. This could be used for "taking a screenshot," with
 /// the added benefit that this method doesn't require a window to be created.
+use wgpu::*;
```
kvark: Global imports have very limited uses (prelude and tests). Let's not use them here.

examples/capture/main.rs
```diff
@@ -0,0 +1,99 @@
...
+            BufferCopyView {
+                buffer: &output_buffer,
+                offset: 0,
+                row_pitch: std::mem::size_of::<u32>() as u32 * size,
+                image_height: size,
+            },
+            texture_extent,
+        );
+
+        encoder
```
kvark: let's move `finish` here

examples/capture/main.rs
kvark: @rukai we wait for all the callbacks on device destruction, so this shouldn't be required. But I think it's still useful to `poll` explicitly here for demonstration only.

examples/capture/main.rs
```diff
@@ -0,0 +1,99 @@
...
+    .finish();
+
+    device.get_queue().submit(&[command_buffer]);
+
+    // Write the buffer as a PNG
+    output_buffer.map_read_async(
+        0,
+        (std::mem::size_of::<u32>() as u32 * size * size) as u64,
+        move |result: BufferMapAsyncResult<&[u8]>| {
+            let png_encoder = image::png::PNGEncoder::new(std::fs::File::create("red.png").unwrap());
```
kvark: could we only use the `png` crate directly (instead of `image`) in order to reduce the dev build times?

tests/clear_screen.rs
```diff
@@ -0,0 +1,100 @@
+use wgpu::*;
+
+#[test]
+#[cfg(any(feature = "vulkan", feature = "metal", feature = "dx12"))]
+fn clear_screen() {
```
kvark: This is a good test to have, but I'm concerned that maybe it's not the shape we want to have a test in. Let's limit the PR to just an example, for now.

To clarify, we are going to use the test suite from upstream WebGPU, plus likely to roll out something akin Warden test framework (in gfx-rs today). It would help reducing the example boilerplate - we don't want to write all the same code for the variations of rendering tests we need at the end of the day.

tests/multithreaded_compute.rs
```diff
@@ -94,6 +94,9 @@ fn multithreaded_compute() {
             staging_buffer.map_read_async(0, size, |result: wgpu::BufferMapAsyncResult<&[u32]>| {
                 assert_eq!(result.unwrap().data, [25, 25, 25]);
             });
+
+            device.poll(true);
```
kvark: let's remove this as well

examples/capture/main.rs
rukai: Oh! Must be a bug then, because I have always had to include this to ensure the map_read_async runs. Created https://github.com/gfx-rs/wgpu-rs/issues/41

tests/clear_screen.rs
paulkernfeld: Okay, that makes sense. Will remove.

examples/capture/main.rs
paulkernfeld: Okay, sounds good. I have switched to explicit imports, but I would also be open to only importing `wgpu` if you think that's better.

tests/multithreaded_compute.rs
paulkernfeld: Good point, this is pretty unrelated.

examples/capture/main.rs
paulkernfeld: Yes, no problem

examples/capture/main.rs
paulkernfeld: 👍

paulkernfeld:
@kvark thanks for the review! I think that I have now addressed all of your comments.

examples/capture/main.rs
kvark: Please follow the examples and use `wgpu` only import. If you don't think it's good, let's talk this over (on gitter, preferably) in the context of all the examples - no need to break consistency before there is any dialog on the problem.

paulkernfeld: Okay, that makes sense. I just pushed a commit to maintain consistency with the other examples.

paulkernfeld:
I squashed into a single commit since this is just adding a single file, but I could squash differently if it's useful to preserve any aspects of the commit history.

Issue #36, opened 2019-07-11
Uniform buffer's field value is chaotic in compute shader


grenlight:
I have defined a ```uniform``` struct named ```FluidUniform```, in the compute shader, , use ```buffer``` to record the uniform value in order, then, use  ```copy_buffer_to_buffer``` copy buffer's value to ```staging_buffer``` and print it, the value of the FluidUniform's ```e``` field is chaotic, and some values are for other fields (such as ```lattice_num```).

```rust
#[repr(C)]
#[derive(Copy, Clone)]
pub struct FluidUniform {
    // e: [[f32; 2]; 9],
    e: [f32; 18],
    lattice_size: [f32; 2],
    lattice_num: [f32; 2],
    weight: [f32; 9],
    swap: i32,
}

// compute shader
layout(set = 0, binding = 0) uniform FluidUniform {
    // vec2 e[9];
    float e[18];
    vec2 lattice_size;
    vec2 lattice_num;
    float weight[9];
    int swap;
};
...
layout (set = 0, binding = 3) buffer FluidBuffer { 
    FluidCell fluidCells[];    
};

// write e's value to buffer
for (int j = 0; j < 6; j ++) {
    fluidCells[j].color[0] = e[j * 3];
    fluidCells[j].color[1] = e[j * 3 + 1];
    fluidCells[j].color[2] = e[j * 3 + 2];
}
```
Print ```staging_buffer```'s value

```rust
encoder.copy_buffer_to_buffer(&self.fluid_buffer, 0, &self.staging_buffer, 0, fluid_buf_range);
self.app_view.device.get_queue().submit(&[encoder.finish()]);
self.staging_buffer.map_read_async(0, fluid_buf_range, |result: wgpu::BufferMapAsyncResult<&[FluidCell]>| {
    println!("{:?}", result.unwrap().data);
});
```

grenlight:
I uploaded an example, only tested on macOS: [https://github.com/grenlight/idroid_demo](https://github.com/grenlight/idroid_demo)
related rust file: [src/fluid/lbm_flow.rs](https://github.com/grenlight/idroid_demo/blob/master/src/fluid/lbm_flow.rs)
compute shader file: [shader/fluid/collide.comp](https://github.com/grenlight/idroid_demo/blob/master/shader/fluid/collide.comp)

kvark:
[...]

PR #44, opened 2019-07-25
Replace the has been deprecated glsl-to-spirv with shaderc-rs crate

grenlight:
Since ```glsl-to-spirv``` has been deprecated, [wgpu-rs/issues/36](https://github.com/gfx-rs/wgpu-rs/issues/36) is related to glsl spec, and [glslc](https://github.com/google/shaderc/tree/master/glslc) is still iterating fast, use ```shaderc-rs``` to follow up on future updates from ```glslc```  .

### Extra information
compile ```shaderc-rs``` need to install ```Pathon``` 3, and maybe  need to remove local computer’s ```CMakeCache.txt``` file to let cmake use latest Pathon version.

If the following error message appears when compiling on macOS: 
```
CMake Error at .../CMakeTestCXXCompiler.cmake:53 (message):
  The C++ compiler
  ...
  is not able to compile a simple test program.
  ```
Make sure have XCode installed as well as command line tools: ```xcode-select --install```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,6 @@ arrayvec = "0.4"
 [dev-dependencies]
 cgmath = "0.17"
 env_logger = "0.6"
-glsl-to-spirv = "0.1"
+shaderc = "0.6"
 log = "0.4"
 png = "0.15"
--- a/examples/framework.rs
+++ b/examples/framework.rs
@@ -1,4 +1,5 @@
 use log::info;
+use shaderc::ShaderKind;
 
 #[cfg_attr(rustfmt, rustfmt_skip)]
 pub const OPENGL_TO_WGPU_MATRIX: cgmath::Matrix4<f32> = cgmath::Matrix4::new(
@@ -24,17 +25,18 @@ pub enum ShaderStage {
 }
 
 pub fn load_glsl(code: &str, stage: ShaderStage) -> Vec<u8> {
-    use std::io::Read;
-
     let ty = match stage {
-        ShaderStage::Vertex => glsl_to_spirv::ShaderType::Vertex,
-        ShaderStage::Fragment => glsl_to_spirv::ShaderType::Fragment,
-        ShaderStage::Compute => glsl_to_spirv::ShaderType::Compute,
+        ShaderStage::Vertex => ShaderKind::Vertex,
+        ShaderStage::Fragment => ShaderKind::Fragment,
+        ShaderStage::Compute => ShaderKind::Compute,
     };
 
-    let mut output = glsl_to_spirv::compile(&code, ty).unwrap();
-    let mut spv = Vec::new();
-    output.read_to_end(&mut spv).unwrap();
+    let mut compiler = shaderc::Compiler::new().unwrap();
+    let binary_result = compiler
+        .compile_into_spirv(&code, ty, "shader.glsl", "main", None)
+        .unwrap();
+    let spv: Vec<u8> = binary_result.as_binary_u8().to_vec(); 
+
     spv
 }
 


grenlight:
#36

grenlight:
Uniform buffer's field value is chaotic #36

grenlight:
[ Uniform buffer's field value is chaotic](https://github.com/gfx-rs/wgpu-rs/issues/36) #36

grenlight:
#36

rukai:
I support this change, but we need to add explain how to build shaderc in the examples readme. You can link to or copy the relevant text from the shaderc readme https://github.com/google/shaderc-rs

rukai:
oh... and I think we should consider using https://github.com/Ralith/vk-shader-macros instead of shaderc-rs.
Its just a nice wrapper over the top, but it helps ensure that we keep our spirv as [u32] instead of [u8] which is UB due to bad alignment.

grenlight:
@rukai When build wgpu-rs examples, shaderc will be compiled automaticly.

rukai:
shaderc-rs handles compilation but your environment has to be specifically setup for it to work, depending on your OS you may need to setup a C++ compiler + ninja or maybe just install the shaderc library via your package manager, if such a package is available.

grenlight:
shaderc compile result [CompilationArtifact](https://docs.rs/shaderc/0.6.0/shaderc/struct.CompilationArtifact.html) can output [u32] spirv stream :).
If use [vk-shader-macros](https://github.com/Ralith/vk-shader-macros), perhaps it's a bit more convenient to use, but since it's just a wrapped shaderc,  can't avoid the shader compilation process.

kvark:
>shaderc-rs handles compilation but your environment has to be specifically setup for it to work, depending on your OS you may need to setup a C++ compiler + ninja or maybe just install the shaderc library via your package manager, if such a package is available.

That's a red flag for me. We don't want to require any more steps than there is. We'd even like to *reduce* the preparation by, say, eliminating the "--features <backend>" parameter.

Could you explain to me one thing. So glsl-to-spirv is deprecated, ok. Is it not doing it's job? In my book, as long as it's working for our needs, we should keep it (given that it doesn't require that much of an environment setup). Later, we'd replace it by something more Rust-ic.

rukai:
Yeah it is bad. I see shaderc-rs as the lesser evil though. I think in this case grenlight was prompted to make this pr because glsl-to-spirv was miscompiling his spirv?

This has been discussed many times before:
https://github.com/Lokathor/learn-gfx-hal/issues/4
https://github.com/gfx-rs/gfx/issues/2546
https://github.com/gfx-rs/gfx/pull/2731

The other option is resuming maintence of glsl-to-spirv under the gfx org. I can help set that up, but I am not interested in long-term maintenence I have enough projects already. Additionally only trusted people could supply prs to update the windows exe.

grenlight:
> We don't want to require any more steps than there is. We'd even like to reduce the preparation by, say, eliminating the "--features " parameter.

I agree with your opinion. Although wgpu users  may already have C++ and Pathon environments, but it is possible to generate additional configurations, that is not friendly enough to some users

aloucks:
> Additionally only trusted people could supply prs to update the windows exe.

You could remove the binary from `glsl-to-spirv` and require the Vulkan SDK to be installed with `glslc` or `glslangValidator` on the path.

seivan:
Not sure if it matters but as of today there is a path of making glslang, which glsl-to-spirv uses to work on mobile devices, like iOS. I know that MoltenVK has managed to do that. Abandoning that means effort has to be put in for making sure that shaderc and it’s rust wrapper does so as well. 

I’m not against the move but there’s an uncertainty that shaderc can compiled for and  run on iOS while glslang has been proven to do - at least the MoltenVK guys managed to do so with their build scripts at least. 

Obviously I’m ignoring any issues with the macro itself and only discussing the internal libs here so if there’s an issue with the macro that’s a different topic.

kvark:
I feel like we could skip the whole glsl-to-spirv -> shaderc transition and instead focus on generating SPIRV in Rust based off GLSL, which becomes more and more of an option.

Is there an urgency here? I mean, wgpu-rs doesn't even have GLSL anywhere in its API. So as long as our examples (that use glsl-to-spirv) still work, we are good.

grenlight:
glsl-to-spirv or shaderc they can all do their work right now, and compared to other enhancements (the use of external texture data :) ), this may not be so urgent, after all, on mobile devices, users can precompile shaders into SPIRV files for use.

kvark:
Closing this as stale. The mid-term plan for the shaders here is using WGSL language and [Naga](https://github.com/gfx-rs/naga) for translating them into SPIR-V.

PR #48, opened 2019-07-29
Exposing `set_viewport_rect()` on `RenderPass`.

seivan:
Noticed that while viewport work with floats, scissors only work with unsigned integers. 
Not sure if relevant but [`width & height`](https://www.khronos.org/registry/OpenGL-Refpages/es2.0/xhtml/glViewport.xml) might not be able to be negative.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1172,6 +1172,13 @@ impl<'a> RenderPass<'a> {
     pub fn set_scissor_rect(&mut self, x: u32, y: u32, w: u32, h: u32) {
         wgn::wgpu_render_pass_set_scissor_rect(self.id, x, y, w, h)
     }
+    
+    /// Sets the viewport region.
+    ///
+    /// Subsequent draw calls will draw any fragments in this region.
+    pub fn set_viewport(&mut self, x: f32, y: f32, w: f32, h: f32, min_depth: f32, max_depth: f32) {
+        wgn::wgpu_render_pass_set_viewport(self.id, x, y, w, h, min_depth, max_depth) 
+    }
 
     /// Sets the stencil reference.
     /// 


src/lib.rs
```diff
@@ -1172,6 +1172,13 @@ impl<'a> RenderPass<'a> {
     pub fn set_scissor_rect(&mut self, x: u32, y: u32, w: u32, h: u32) {
         wgn::wgpu_render_pass_set_scissor_rect(self.id, x, y, w, h)
     }
+    
+    /// Sets the viewport region.
+    ///
+    /// Subsequent draw calls will draw any fragments in this region.
+    pub fn set_viewport_rect(&mut self, x: f32, y: f32, w: f32, h: f32, min_depth: f32, max_depth: f32) {
```
kvark: let's rename to just `set_viewport` that better reflects the upstream function. The reason of not putting "rect" at the end is because it's more than just a rect: it also includes the depth range.

kvark: I think it would also make sense to have a simple `struct Rect<T>` to be used in both scissor and viewport commands - this we can do later though :)

kvark:
Also, please squash the changes.

src/lib.rs
seivan: @kvark I already got `Rect` - which consists of `Point` and `Size` (not generic)  in my own code that depends on wgpu. My concern is that it would just pollute the namespace and make methods look heavier than they are. Working with primitives is more intuitive when the rest of the code (in my book) is low level. 

`Rect<T>`  is fine, my concern isn't for that but for stuff like `PipelineLayoutDescriptor` and `BindGroupLayoutDescriptor` that essentially only take arrays of other classes.

kvark: we still need to rename the method though

seivan:
@kvark Is there anything else left to get this merged in?

kvark:
Bors retry

PR #57, opened 2019-08-07
Fix sampling in mipmap generation example

### LINKED ISSUE #55: Mipmap example does not downsample correctly
This bug is for the current git version.

The mipmap example currently downsamples using Nearest rather than Linear.
Changing the enum on `mag_filter` to `Linear` instead fixes it and correctly averages pixels:
https://github.com/gfx-rs/wgpu-rs/blob/master/examples/mipmap/main.rs#L137

I was under the impression that `mag_filter` refers to the filtering method used when upsampling, not when downsampling. Is this a bug in my understanding, or a bug in the code?

LaylConway:
Fixes #55

### DIFF
--- a/examples/mipmap/main.rs
+++ b/examples/mipmap/main.rs
@@ -135,7 +135,7 @@ impl Example {
             address_mode_v: wgpu::AddressMode::ClampToEdge,
             address_mode_w: wgpu::AddressMode::ClampToEdge,
             mag_filter: wgpu::FilterMode::Linear,
-            min_filter: wgpu::FilterMode::Linear,
+            min_filter: wgpu::FilterMode::Nearest,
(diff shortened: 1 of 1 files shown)


examples/mipmap/main.rs
```diff
@@ -134,7 +134,7 @@ impl Example {
             address_mode_u: wgpu::AddressMode::ClampToEdge,
             address_mode_v: wgpu::AddressMode::ClampToEdge,
             address_mode_w: wgpu::AddressMode::ClampToEdge,
-            mag_filter: wgpu::FilterMode::Nearest,
+            mag_filter: wgpu::FilterMode::Linear,
             min_filter: wgpu::FilterMode::Linear,
```
kvark: let's switch `min_filter` to `Nearest` to follow the spirit of the original code

PR #58, opened 2019-08-10
Update to latest wgpu native

rukai:
This PR is waiting on https://github.com/gfx-rs/wgpu/pull/276 but also allows us to test 276

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ gl = ["wgn/gfx-backend-gl"]
 
 [dependencies]
 #TODO: only depend on the published version
-wgn = { package = "wgpu-native", version = "0.2.6", features = ["local", "window-winit"], git = "https://github.com/gfx-rs/wgpu", rev = "5224bb812420e420485ee08ad8cb820695a3e6eb" }
+wgn = { package = "wgpu-native", version = "0.2.6", features = ["local", "window-winit"], git = "https://github.com/gfx-rs/wgpu", rev = "f72dfc9cbbcf752ea5905ff9b038f8ca3f1d3691"}
 arrayvec = "0.4"
 zerocopy = "0.2"
 
@@ -34,7 +34,3 @@ env_logger = "0.6"
 glsl-to-spirv = "0.1"
 log = "0.4"
 png = "0.15"
-
-[patch.crates-io]
-#gfx-hal = { path = "../gfx/src/hal" }
-#gfx-backend-dx12 = { path = "../gfx/src/backend/dx12" }
--- a/examples/framework.rs
+++ b/examples/framework.rs
@@ -23,7 +23,7 @@ pub enum ShaderStage {
     Compute,
 }
 
-pub fn load_glsl(code: &str, stage: ShaderStage) -> Vec<u8> {
+pub fn load_glsl(code: &str, stage: ShaderStage) -> Vec<u32> {
     use std::io::Read;
 
     let ty = match stage {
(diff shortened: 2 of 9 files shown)


rukai:
wgpu-native is now pointing at latest git master.
Should be good to merge.

examples/hello-triangle/main.rs
```diff
@@ -59,9 +61,18 @@ fn main() {
...
+    for bytes4 in vs_bytes.chunks(4) {
+        vs_words.push(u32::from_le_bytes([bytes4[0], bytes4[1], bytes4[2], bytes4[3]]));
+    }
+    let vs_module = device.create_shader_module(&vs_words);
+
     let fs_bytes = include_bytes!("shader.frag.spv");
-    let fs_module = device.create_shader_module(fs_bytes);
+    let mut fs_words = vec!();
+    for bytes4 in fs_bytes.chunks(4) {
+        fs_words.push(u32::from_le_bytes([bytes4[0], bytes4[1], bytes4[2], bytes4[3]]));
```
m4b: I don't think will this work on a big endian machine? Also, if willing to use unsafe transmute, can save a copy and allocation here by simply doing something like:
```
assert!(fs_bytes.len() % 4 == 0);
let fs_module = device.create_shader_module(transmute(fs.words));
```
you will also avoid endianness issues I believe

Cargo.toml
```diff
@@ -34,7 +34,3 @@ env_logger = "0.6"
 glsl-to-spirv = "0.1"
 log = "0.4"
 png = "0.15"
-
-[patch.crates-io]
```
kvark: I added these because I think they are useful. Whenever you want to test a local gfx override, you can see how this is done, only edit a few characters and good to go.

examples/framework.rs
```diff
@@ -33,9 +33,14 @@ pub fn load_glsl(code: &str, stage: ShaderStage) -> Vec<u8> {
...
 
     let mut output = glsl_to_spirv::compile(&code, ty).unwrap();
-    let mut spv = Vec::new();
-    output.read_to_end(&mut spv).unwrap();
-    spv
+    let mut spv_bytes = Vec::new();
+    output.read_to_end(&mut spv_bytes).unwrap();
+
+    let mut spv_words = Vec::new();
+    for bytes4 in spv_bytes.chunks(4) {
```
kvark: fyi, we could re-export [read_spirv](https://docs.rs/gfx-hal/0.3.0/gfx_hal/pso/fn.read_spirv.html) from gfx-hal and use here

examples/hello-triangle/main.rs
kvark: this wouldn't solve the alignment problem

m4b:
Alignment is just mod 4 on the pointer. I’m not sure why it would be unaligned in the first place, but you could implement the same behavior in case it’s unaligned, and save a copy in almost all cases.

That withstanding the code merged isn’t correct on big endian machines, so it shouldn’t have been merged.

kvark:
Yeah, a heap-allocated vector would be aligned already. The problem comes into play only when one deals with `&[u8]` slices, so I shouldn't have brought that up.

> That withstanding the code merged isn’t correct on big endian machines, so it shouldn’t have been merged.

I don't think we currently support any big endian machines. Do you have an example?
At the same time, it would surely be *nice* to keep the code endian-agnostic, and thus I suggest us to use gfx's `read_spirv` as a follow-up.

m4b:
I just realized they're all examples so it's not super important.

But this brings me to:

why was the shader code changed from `&[u32] -> &[u8]`? This is disastrous from an api standpoint since now:

1. every client has to update their code with likely incorrect endianness conversion
2. unnecessary heap allocations similar to example to avoid unsafe
3. `include_bytes` no longer works out of the box, and now extra boilerplate, at the client side needs to be done everywhere

Was this added to the spec? Otherwise I'm really at a loss why this extra burden would be pushed onto clients, or at least a helper function isn't exposed in the api to easily use `&[u8]` slices

rukai:
Kvark has suggested reexposing read_spirv which should address all of your points. I'll make a pr when I get home from work.

The change was done to prevent alignment issues from safe code. There is some more information here https://github.com/Ralith/vk-shader-macros/blob/master/README.md#why-u32

PR #61, opened 2019-08-12
Cleanup examples by using wgpu::read_spirv instead of manually creating a Vec<u32>

rukai:
Wait for me to update the git reference once https://github.com/gfx-rs/wgpu/pull/280 is merged.

### DIFF
--- a/examples/framework.rs
+++ b/examples/framework.rs
@@ -24,23 +24,13 @@ pub enum ShaderStage {
 }
 
 pub fn load_glsl(code: &str, stage: ShaderStage) -> Vec<u32> {
-    use std::io::Read;
-
     let ty = match stage {
         ShaderStage::Vertex => glsl_to_spirv::ShaderType::Vertex,
         ShaderStage::Fragment => glsl_to_spirv::ShaderType::Fragment,
         ShaderStage::Compute => glsl_to_spirv::ShaderType::Compute,
     };
 
-    let mut output = glsl_to_spirv::compile(&code, ty).unwrap();
-    let mut spv_bytes = Vec::new();
-    output.read_to_end(&mut spv_bytes).unwrap();
-
-    let mut spv_words = Vec::new();
-    for bytes4 in spv_bytes.chunks(4) {
-        spv_words.push(u32::from_le_bytes([bytes4[0], bytes4[1], bytes4[2], bytes4[3]]));
-    }
-    spv_words
+    wgpu::read_spirv(glsl_to_spirv::compile(&code, ty).unwrap()).unwrap()
 }
 
 pub trait Example {
--- a/examples/hello-compute/main.rs
+++ b/examples/hello-compute/main.rs
@@ -25,12 +25,8 @@ fn main() {
         limits: wgpu::Limits::default(),
     });
 
-    let cs_bytes = include_bytes!("shader.comp.spv");
-    let mut cs_words = vec!();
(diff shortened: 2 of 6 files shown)


examples/hello-compute/main.rs
```diff
@@ -25,12 +25,8 @@ fn main() {
...
     });
 
-    let cs_bytes = include_bytes!("shader.comp.spv");
-    let mut cs_words = vec!();
-    for bytes4 in cs_bytes.chunks(4) {
-        cs_words.push(u32::from_le_bytes([bytes4[0], bytes4[1], bytes4[2], bytes4[3]]));
-    }
-    let cs_module = device.create_shader_module(&cs_words);
+    let cs = include_bytes!("shader.comp.spv");
+    let cs_module = device.create_shader_module(&wgpu::read_spirv(std::io::Cursor::new(&cs[..])).unwrap());
```
kvark: we should be able to use `cs` instead of `&cs[..]` here and in other places

examples/framework.rs
```diff
@@ -24,23 +24,13 @@ pub enum ShaderStage {
...
-    let mut output = glsl_to_spirv::compile(&code, ty).unwrap();
-    let mut spv_bytes = Vec::new();
-    output.read_to_end(&mut spv_bytes).unwrap();
-
-    let mut spv_words = Vec::new();
-    for bytes4 in spv_bytes.chunks(4) {
-        spv_words.push(u32::from_le_bytes([bytes4[0], bytes4[1], bytes4[2], bytes4[3]]));
-    }
-    spv_words
+    wgpu::read_spirv(glsl_to_spirv::compile(&code, ty).unwrap()).unwrap()
```
kvark: I think this can just use transmute, since we are heap-allocating the byte vector when reading `glsl_to_spirv` output

examples/hello-compute/main.rs
rukai:
```
   Compiling wgpu v0.2.2 (/home/rubic/Projects/Crates/wgpu/wgpu-rs)
error[E0277]: the trait bound `[u8; 1576]: std::convert::AsRef<[u8]>` is not satisfied
  --> examples/hello-compute/main.rs:29:50
   |
29 |     let cs_module = device.create_shader_module(&wgpu::read_spirv(std::io::Cursor::new(cs)).unwrap());
   |                                                  ^^^^^^^^^^^^^^^^ the trait `std::convert::AsRef<[u8]>` is not implemented for `[u8; 1576]`
   |
   = help: the following implementations were found:
             <[T; 0] as std::convert::AsRef<[T]>>
             <[T; 10] as std::convert::AsRef<[T]>>
             <[T; 11] as std::convert::AsRef<[T]>>
             <[T; 12] as std::convert::AsRef<[T]>>
           and 30 others
   = note: required because of the requirements on the impl of `std::convert::AsRef<[u8]>` for `&[u8; 1576]`
   = note: required because of the requirements on the impl of `std::io::Read` for `std::io::Cursor<&[u8; 1576]>`
   = note: required by `gfx_hal::pso::read_spirv`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0277`.
error: Could not compile `wgpu`.
```

examples/framework.rs
rukai: Wouldn't that be UB due to incorrect alignment?

kvark: yeah, let's keep it safe :)

Issue #66, opened 2019-08-17
Update to latest wgpu


rukai:
I'm working on this.

PR #67, opened 2019-08-17
Update to latest wgpu-native commit

rukai:
closes https://github.com/gfx-rs/wgpu-rs/issues/66

Make sure my documentation for sample_mask and alpha_to_coverage_enabled is correct, I am not familiar with these fields.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,7 +11,7 @@ use std::slice;
 
 pub use wgn::winit;
 pub use wgn::{
-    AdapterDescriptor,
+    RequestAdapterOptions,
     AddressMode,
     BindGroupLayoutBinding,
     BindingType,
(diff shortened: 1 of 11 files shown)


src/lib.rs
```diff
@@ -333,6 +334,12 @@ pub struct RenderPipelineDescriptor<'a> {
 
     /// The number of samples calculated per pixel (for MSAA).
     pub sample_count: u32,
+
+    /// Bitmask deciding how much of the samples are kept.
```
kvark: Bitmask restricts the samples of a pixel modified by this pipeline.

src/lib.rs
```diff
@@ -333,6 +334,12 @@ pub struct RenderPipelineDescriptor<'a> {
...
+    pub sample_mask: u32,
+
+    /// When enabled, the sample_mask is ANDed with the the alpha value.
```
kvark: When enabled, produces another sample mask per pixel based on the alpha output value, that is ANDed with the `sample_mask` and the primitive coverage to restrict the set of samples affected by a primitive.

The implicit mask produced for alpha of zero is guaranteed to be zero, and for alpha of one is guaranteed to  be all 1-s.

rukai: Thanks!

PR #70, opened 2019-08-21
Wholesome update for wgpu-0.3

kvark:
~~This update is a little more opinionated, i.e. the BGL creation just accepts a slice of things instead of a descriptor. I believe the reason for descriptors in the upstream API is mostly techincal - it's more convenient to generate bindings in Googles giant codegen. Following it verbatim doesn't appear to be strictly necessary, especially if we can extract better usability from Rust features.~~

The change also disables "gl" feature for the release. We'll re-enable it once we can provide it nicely, it will not be a breaking change.

I also noticed that barriers are not working correctly on the mipmap example. Will investigate separately - the native part is already published anyway. Edit: fixed by https://github.com/gfx-rs/wgpu/pull/302

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,9 +4,12 @@ version = "0.3.0"
 authors = [
 	"Dzmitry Malyshau <kvark@mozilla.com>",
 	"Joshua Groves <josh@joshgroves.com>",
+	"Lucas Kent <rubickent@gmail.com>",
+	"kyren <kerriganw@gmail.com>",
+	"Cormac O'Brien <cormac@c-obrien.org>",
 ]
 edition = "2018"
-description = "Rusty wgpu API wrapper"
+description = "Rusty WebGPU API wrapper"
 homepage = "https://github.com/gfx-rs/wgpu-rs"
 repository = "https://github.com/gfx-rs/wgpu-rs"
 keywords = ["graphics"]
@@ -20,12 +23,13 @@ metal = ["wgn/gfx-backend-metal"]
 dx11 = ["wgn/gfx-backend-dx11"]
 dx12 = ["wgn/gfx-backend-dx12"]
 vulkan = ["wgn/gfx-backend-vulkan"]
-gl = ["wgn/gfx-backend-gl"]
+#TODO: there is no way to use glutin-0.20 with GL backend v0.3 yet
+#gl = ["wgn/gfx-backend-gl"]
 
 [dependencies]
-#TODO: only depend on the published version
-wgn = { package = "wgpu-native", version = "0.2.6", features = ["local", "window-winit"], git = "https://github.com/gfx-rs/wgpu", rev = "a47ff090bb042f1cb2ad7b13c76eca228390a311"}
+wgn = { package = "wgpu-native", version = "0.3.1", features = ["local"] }
 arrayvec = "0.4"
+raw-window-handle = "0.1"
 zerocopy = "0.2"
 
 [dev-dependencies]
@@ -34,3 +38,4 @@ env_logger = "0.6"
 glsl-to-spirv = "0.1"
 log = "0.4"
 png = "0.15"
+winit = "0.20.0-alpha3"
--- a/examples/cube/main.rs
+++ b/examples/cube/main.rs
@@ -129,26 +129,22 @@ impl framework::Example for Example {
                 wgpu::BindGroupLayoutBinding {
                     binding: 0,
                     visibility: wgpu::ShaderStage::VERTEX,
-                    ty: wgpu::BindingType::UniformBuffer,
-                    dynamic: false,
-                    multisampled: false,
-                    texture_dimension: wgpu::TextureViewDimension::D2,
+                    ty: wgpu::BindingType::UniformBuffer {
+                        dynamic: false,
+                    },
                 },
                 wgpu::BindGroupLayoutBinding {
                     binding: 1,
                     visibility: wgpu::ShaderStage::FRAGMENT,
-                    ty: wgpu::BindingType::SampledTexture,
-                    dynamic: false,
-                    multisampled: false,
-                    texture_dimension: wgpu::TextureViewDimension::D2,
+                    ty: wgpu::BindingType::SampledTexture {
+                        multisampled: false,
+                        dimension: wgpu::TextureViewDimension::D2,
+                    },
                 },
                 wgpu::BindGroupLayoutBinding {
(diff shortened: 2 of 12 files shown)


src/lib.rs
```diff
@@ -517,35 +534,6 @@ impl Instance {
...
-        display: *mut *const std::ffi::c_void,
-        window: u64,
-    ) -> Surface {
-        Surface {
-            id: wgn::wgpu_instance_create_surface_from_xlib(self.id, display, window),
-        }
-    }
-
-    #[cfg(not(feature = "gl"))]
-    pub fn create_surface_from_macos_layer(&self, layer: *mut std::ffi::c_void) -> Surface {
```
seivan: Is this being removed?

kvark: indeed, we should keep that one

seivan: Is it too late to give my feedback on the `raw-window-handle` stuff or is this something decided? It seems very unfinished as an idea, and I do mean just the idea alone sans implementation. https://github.com/rust-windowing/raw-window-handle/issues/14

examples/cube/main.rs
```diff
@@ -124,34 +124,28 @@ impl framework::Example for Example {
...
-                    texture_dimension: wgpu::TextureViewDimension::D2,
+                    dimension: wgpu::TextureViewDimension::D2,
                 },
-            ],
-        });
+            },
+            wgpu::BindGroupLayoutBinding {
+                binding: 2,
+                visibility: wgpu::ShaderStage::FRAGMENT,
+                ty: wgpu::BindingType::Sampler,
```
rukai: Uh what? What happened to these fields?

src/lib.rs
```diff
@@ -490,24 +507,24 @@ impl Instance {
...
-    /// If there are no available adapters matching `desc`, this function will return another
+    /// If there are no available adapters matching `options`, this function will return another
     /// adapter.
     ///
     /// # Panics
     ///
     /// Panics if there are no available adapters. This will occur if none of the graphics backends
     /// are enabled.
-    pub fn get_adapter(&self, options: Option<&RequestAdapterOptions>) -> Adapter {
+    pub fn request_adapter(&self, options: &RequestAdapterOptions) -> Adapter {
```
rukai: How are you supposed to call request_adapter(None) now?

src/lib.rs
```diff
@@ -554,9 +542,9 @@ impl Adapter {
     /// # Panics
     ///
     /// Panics if the extensions specified by `desc` are not supported by this adapter.
-    pub fn request_device(&self, desc: Option<&DeviceDescriptor>) -> Device {
+    pub fn request_device(&self, desc: &DeviceDescriptor) -> Device {
```
rukai: How are you supposed to call request_device(None) now?

src/lib.rs
kvark: it's equivalent to `request_adapter(&wgpu::RequestAdapterOptions::default())`

src/lib.rs
kvark: I think we can live with NSView/UIView, but the discussion about `CAMetalLayer` does still need to happen. Thanks for filing the issue!

src/lib.rs
kvark: similarly, by doing `request_device(&wgpu::DeviceDescriptor::default())`

Cargo.toml
```diff
@@ -20,12 +20,13 @@ metal = ["wgn/gfx-backend-metal"]
 dx11 = ["wgn/gfx-backend-dx11"]
 dx12 = ["wgn/gfx-backend-dx12"]
 vulkan = ["wgn/gfx-backend-vulkan"]
-gl = ["wgn/gfx-backend-gl"]
+#gl = ["wgn/gfx-backend-gl"]
 
 [dependencies]
 #TODO: only depend on the published version
```
yanchith: This TODO is actually done now?

kvark: indeed, will fix that, thanks!

src/lib.rs
rukai: Oh yeah fair enough.

I guess my only issue is one day when we are using wgpu-rs on the web, we wont be able to access that part of the API, if only for testing different browser implementations, but that seems pretty obscure.

kvark: The `Option` thing in WebIDL specification itself is defined to be equivalent of supplying a dictionary with the fields full of default values, so it's not like a browser has freedom to do anything other than this (it would be a bug!).

src/lib.rs
```diff
@@ -258,12 +255,32 @@ pub struct Binding<'a> {
     pub resource: BindingResource<'a>,
 }
 
-/// A description of a bind group layout.
-///
-/// A `BindGroupLayoutDescriptor` can be passed to [`Device::create_bind_group_layout`] to obtain a
-/// [`BindGroupLayout`].
-pub struct BindGroupLayoutDescriptor<'a> {
```
grovesNL: I think one reason the descriptor was useful is because of the inheritance chain (`GPUBindGroupLayoutDescriptor : GPUObjectDescriptorBase`). `GPUObjectDescriptorBase` has `DOMString? label` (which we haven't implemented yet), so by accepting a slice here we won't be able to provide debug labels for bind group layouts.

Cargo.toml
```diff
@@ -20,12 +20,13 @@ metal = ["wgn/gfx-backend-metal"]
 dx11 = ["wgn/gfx-backend-dx11"]
 dx12 = ["wgn/gfx-backend-dx12"]
 vulkan = ["wgn/gfx-backend-vulkan"]
-gl = ["wgn/gfx-backend-gl"]
+#gl = ["wgn/gfx-backend-gl"]
```
grovesNL: Is there an issue for the part that gl is blocked on here? It would probably be useful to clarify it in a comment for anyone using gl patches at the moment

src/lib.rs
kvark: aha, great point!

Cargo.toml
kvark: issue is that we initialize all other backends without `winit` feature
for GL, there is currently no way to use it without `glutin`. Enabling `glutin` though on the backend would conflict with `glutin` in wgpu-rs itself, since the backends are from crates.io and they rely on the old `glutin` version...
This will be unlocked if/when we back-port raw-window-handle to gfx-backend-gl-0.3

kvark:
I think everything got reviewed and addressed...
bors r=everyone

PR #82, opened 2019-09-10
Return Queue separately

kvark:
What problem is this PR trying to solve? We want `Device` to be freely accessible from multiple threads/objects and internally synchronized. `Arc<Device>` seems like a natural choice of such a sharable object, especially since all except one methods are `&self`.

That one method is `get_queue()`, and it returns a temporary object `Queue<'a>`. If we turn it into `&self`, we'd end up with multiple instances of `Queue` created at any time, which contradicts the initial design (of this Rust wrapper). If it stays `&mut` and the user wraps the device into `Arc`, they'll never be able to submit any work...

So this PR solves this by moving the `Queue` completely outside of the device.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -73,7 +73,6 @@ pub use wgn::glutin;
 struct Temp {
     //bind_group_descriptors: Vec<wgn::BindGroupDescriptor>,
     //vertex_buffers: Vec<wgn::VertexBufferDescriptor>,
-    command_buffers: Vec<wgn::CommandBufferId>,
 }
 
 /// A handle to a physical graphics and/or compute device.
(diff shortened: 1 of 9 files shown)


src/lib.rs
```diff
@@ -546,15 +545,21 @@ impl Adapter {
     }
 
     /// Requests a connection to a physical device, creating a logical device.
+    /// Returns the device together with a queue that executres command buffers.
```
grovesNL: typo in `executres`

src/lib.rs
```diff
@@ -546,15 +545,21 @@ impl Adapter {
...
     ///
     /// # Panics
     ///
     /// Panics if the extensions specified by `desc` are not supported by this adapter.
-    pub fn request_device(&self, desc: &DeviceDescriptor) -> Device {
-        Device {
+    pub fn request_device(&self, desc: &DeviceDescriptor) -> (Device, Queue) {
```
grovesNL: How would this work with multiple queues? We'd still want to provide the device in `request_device` but the queue would be separate.

kvark: We'll see when we get there, perhaps? So far the working group hasn't been super active on the topic.

kvark:
bors r=grovesNL
open to other ideas on how to make `Arc<Device>` :)

PR #83, opened 2019-09-10
Propagate making Vulkan backend available to wgpu-rs

yanchith:
`wgpu-native` has `gfx-backend-vulkan` feature which makes Vulkan available on macOS. This adds the feature to `wgpu-rs` also.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,8 @@ license = "MPL-2.0"
 
 [features]
 default = []
+# Make Vulkan backend available on platforms where it is by default not, e.g. macOS
+vulkan = ["wgn/gfx-backend-vulkan"]
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -19,6 +19,8 @@ license = "MPL-2.0"
 
 [features]
 default = []
+# Make Vulkan backend available on platforms where it is by default not, e.g. macOS
+gfx-backend-vulkan = ["wgn/gfx-backend-vulkan"]
```
kvark: This is a bit tricky, since the feature is not available on all platforms (only iOS and macOS). Need to figure out how cargo treats this.

yanchith: Isn't the feature available everywhere, just not really used except macOS?

kvark: Speaking in terms of strictly cargo features, it's only optional on iOS/macOS. On other platforms it's not optional, therefore not a feature.

PR #89, opened 2019-09-22
Adapter::get_info

paulkernfeld:
This more or less addresses #7.

If the example is too trivial, I'm happy to remove it.

### DIFF
--- a/examples/describe/main.rs
+++ b/examples/describe/main.rs
@@ -1,0 +1,11 @@
+/// This example shows how to describe the adapter in use.
+fn main() {
+    env_logger::init();
+
+    let adapter = wgpu::Adapter::request(&wgpu::RequestAdapterOptions {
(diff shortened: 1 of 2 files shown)


lachlansneff:
Would it make sense to make the `Debug` impl call the `get_info` method as well?

paulkernfeld:
My concern is that `Adapter::get_info` could conceivably panic if something goes wrong, and most users would probably expect an implementation of `Debug` to never panic. However, it's not clear that I have any idea what I'm talking about, so maybe some of this reasoning is wrong! 😊

lachlansneff:
Hmm. Maybe you could wrap the `get_info` call in [`panic::catch_unwind`](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html) and then print out something along the lines of "Failed to get adapter info" if it panics. Of course, if panic is set to abort for that particular build, you cannot catch the panic.

PR #100, opened 2019-10-24
Fix shadow sampling

kvark:


examples/shadow/forward.frag
```diff
@@ -37,13 +50,7 @@ void main() {
...
         Light light = u_Lights[i];
         // project into the light space
-        vec4 light_local = light.proj * v_Position;
-        // compute texture coordinates for shadow lookup
-        light_local.xy = (light_local.xy/light_local.w + 1.0) / 2.0;
-        light_local.w = light_local.z / light_local.w;
-        light_local.z = i;
-        // do the lookup, using HW PCF and comparison
-        float shadow = texture(sampler2DArrayShadow(t_Shadow, s_Shadow), light_local);
+        float shadow = fetch_shadow(i, light.proj * v_Position);
```
m4b: this still doesn't seem to fix some of the artifacts i've seen when the Y is moved to +10 or +15, but it might not be intended to, dunno

Issue #93, opened 2019-10-07
skybox cubemap example


m4b:
A new example for how to write an idiomatic skybox with a cubemap in wgpu-rs would be great.

Opening this issue in case there’s some superhuman out there who wants to help us all out

m4b:
Here’s an older GL reference which explains the general idea of a skybox in case someone doesn’t know what it is, or how it works: http://ogldev.atspace.co.uk/www/tutorial25/tutorial25.html

kvark:
Old gfx-rs example: https://github.com/gfx-rs/gfx/tree/pre-ll/examples/skybox

caelunshun:
I'll take a stab at this and hopefully submit a PR soon.

m4b:
ping @caelunshun any progress on this?

caelunshun:
> ping @caelunshun any progress on this?

I've been a bit busy lately, but I still plan to complete this. Should be done sometime this week.

m4b:
So I started messing around with what a skybox might look like, and i think i'm stuck trying to figure out how to get `TextureDimensions`, `TextureViewDimensions`, the backing texture, and texture views working right; i think you'd have to do something like:

```rust
let render_layout = device.create_bind_group_layout(&wgpu::BindGroupLayoutDescriptor {
    bindings: &[
        wgpu::BindGroupLayoutBinding {
            binding: 0,
            visibility: wgpu::ShaderStage::VERTEX | wgpu::ShaderStage::FRAGMENT,
            ty: wgpu::BindingType::UniformBuffer { dynamic: false },
        },
        wgpu::BindGroupLayoutBinding {
            binding: 2,
            visibility: wgpu::ShaderStage::FRAGMENT,
            ty: wgpu::BindingType::SampledTexture {
                multisampled: false,
                dimension: wgpu::TextureViewDimension::Cube,
            },
        },
    ],
});

...

// FIXME: what do we do here?
let texture_extent = wgpu::Extent3d {
    width: 100,
    height: 100,
    // width: image_width,
    // height: image_height,
    depth: 1, // do we use the extend to index into 1 of the 6 faces of the cubemap?
};

let texture = device.create_texture(&wgpu::TextureDescriptor {
    size: texture_extent,
    array_layer_count: 1,
    mip_level_count: 1,
    sample_count: 1,
    dimension: wgpu::TextureDimension::D3, // is this a d2 or a d3...?
    format: SKYBOX_FORMAT,
[...]

PR #109, opened 2019-11-01
add initial skybox example, ref #93

m4b:
1. doesn't render the images at all
2. mip map stuff needs clarification
3. will update uniform with temporal rotation so we can see the skybox better


Cargo.toml
```diff
@@ -35,3 +35,4 @@ glsl-to-spirv = "0.1"
 log = "0.4"
 png = "0.15"
 winit = "0.20.0-alpha4"
+image = "0.22"
```
kvark: would it be possible to use `png` images for the cubemap? Keeping it only to `png` dependency would mean our test builds are faster and breaking less often because of unrelated stuff in `image`, which is bigger.

examples/skybox/skybox_frag.glsl
```diff
@@ -0,0 +1,11 @@
+#version 450
+
+layout(set = 0, binding = 1) uniform textureCube t_cubemap;
```
kvark: please keep the naming consistent with the shaders in other examples

examples/skybox/skybox_vert.glsl
```diff
@@ -0,0 +1,19 @@
+#version 450
+
+layout(location = 0) in vec2 position;
```
kvark: we don't actually need vertex attributes for shaders that just draw a full-screen quad, instead we can generate the positions based on the vertex ID, like the other examples do, like the mipmap one:
```glsl
    vec2 tc = vec2(0.0);
    switch(gl_VertexIndex) {
        case 0: tc = vec2(1.0, 0.0); break;
        case 1: tc = vec2(1.0, 1.0); break;
        case 2: tc = vec2(0.0, 0.0); break;
        case 3: tc = vec2(0.0, 1.0); break;
    }
    v_TexCoord = tc;
    gl_Position = vec4(tc * 2.0 - 1.0, 0.5, 1.0);
```

examples/skybox/main.rs
```diff
@@ -0,0 +1,299 @@
...
+                module: &vs_module,
+                entry_point: "main",
+            },
+            fragment_stage: Some(wgpu::ProgrammableStageDescriptor {
+                module: &fs_module,
+                entry_point: "main",
+            }),
+            rasterization_state: Some(wgpu::RasterizationStateDescriptor {
+                front_face: wgpu::FrontFace::Cw,
+                cull_mode: wgpu::CullMode::Back,
```
kvark: we don't need culling for a full-screen pass

examples/skybox/main.rs
```diff
@@ -0,0 +1,299 @@
...
+        let texture_extent = wgpu::Extent3d {
+            width: image_width,
+            height: image_height,
+            depth: 1,
+        };
+
+        let texture = device.create_texture(&wgpu::TextureDescriptor {
+            size: texture_extent,
+            array_layer_count: 6,
+            mip_level_count: 6, // [WIP] i'm not sure this is right, but if this is 0 it segfaults
```
kvark: you only need 1 mipmap for now

examples/skybox/main.rs
```diff
@@ -0,0 +1,299 @@
...
+            sample_count: 1,
+            dimension: wgpu::TextureDimension::D2,
+            format: SKYBOX_FORMAT,
+            usage: wgpu::TextureUsage::SAMPLED
+                | wgpu::TextureUsage::COPY_DST
+                | wgpu::TextureUsage::COPY_SRC,
```
kvark: why do you need COPY_SRC?

examples/skybox/main.rs
```diff
@@ -0,0 +1,299 @@
...
+            for (i, image) in faces.iter().enumerate() {
+                let (image_width, image_height) = image.dimensions();
+                println!(
+                    "Copying skybox image {} of size {},{} to gpu",
+                    i, image_width, image_height,
+                );
+                let image_buf = device
+                    .create_buffer_mapped(
+                        image.len(),
+                        wgpu::BufferUsage::COPY_SRC | wgpu::BufferUsage::COPY_DST,
```
kvark: it shouldn't need COPY_DST

examples/skybox/main.rs
```diff
@@ -0,0 +1,299 @@
...
+                encoder.copy_buffer_to_texture(
+                    wgpu::BufferCopyView {
+                        buffer: &image_buf,
+                        offset: 0,
+                        row_pitch: 4 * image_width,
+                        image_height,
+                    },
+                    wgpu::TextureCopyView {
+                        texture: &texture,
+                        mip_level: i as u32, // [WIP] this doesn't seem right either?
```
kvark: you need 0 here

examples/skybox/main.rs
```diff
@@ -0,0 +1,299 @@
...
+                );
+            }
+        }
+
+        let texture_view = texture.create_view(&wgpu::TextureViewDescriptor {
+            format: SKYBOX_FORMAT,
+            dimension: wgpu::TextureViewDimension::Cube,
+            aspect: wgpu::TextureAspect::default(),
+            base_mip_level: 0,
+            level_count: 6, // [WIP] this doesn't seem right either
```
kvark: that should be 1

Cargo.toml
m4b: Sure, will update

examples/skybox/skybox_vert.glsl
m4b: Ok this part was also a little unclear to me, so:

1. I have to send something to fire the shader, yes ? So just a Float for vertex type and pass 3 dummy values ?
2. What specific values should I set ? Same as the ones which I send in now in the rust part?

examples/skybox/main.rs
m4b: Ok I definitely had this as 1 before and I had validation issues on Linux. I’ll set to the suggested mip values, I might now have tried with mip level 0 though when copying, will try

m4b:
I forgot to ask, but I’m not seeing anything at all rendered; correct me if wrong but the only real changes here are changing the mip values, the copy dst shouldn’t affect presentation on screen, ditto vertices in shader, so unless somehow the mip values are affecting what is rendered to screen, is there anything else I should fix up here to make it work ?

kvark:
@m4b you have validation errors, so that alone throws off all the bets.
Next, mipmap levels alone can affect rendering, since you are only providing level-0. Means other levels are not initialized (could be zeroes) but still used, which can make stuff seen as black.
Finally, please use any kind of GPU debugger to see what's wrong with rendering.

m4b:
Ok updated everything except the vertices thing, but that shouldn't have any affect on output afaik (assuming vertices are correct).
There are no validation errors, and still seeing same background without images.
Unfortunately i can't seem to get renderdoc working on my linux machine, it panics when executing the binary with "Vulkan driver does not support VK_KHR_WAYLAND_SURFACE", which doesn't make much sense since when i run it standalone, it runs fine? (and yes i'm on running wayland).

Don't see an easy way to setup Xcode environment, using external build system seems to disable gpu captures.

Anyway i'll keep messing with Xcode to see if i can capture a frame, but in meantime, do you see anything that's sticking out?

examples/skybox/main.rs
```diff
@@ -0,0 +1,303 @@
...
+            "examples/skybox/images/posx.png",
+            "examples/skybox/images/posy.png",
+            "examples/skybox/images/posz.png",
+            "examples/skybox/images/negx.png",
+            "examples/skybox/images/negy.png",
+            "examples/skybox/images/negz.png",
+        ];
+	// we set these multiple times, but whatever
+	let (mut image_width, mut image_height) = (0, 0);
+        let faces = paths
```
kvark: spacing is broken here (rogue tabs?)

examples/skybox/main.rs
```diff
@@ -0,0 +1,303 @@
...
+            mip_level_count: 1,
+            sample_count: 1,
+            dimension: wgpu::TextureDimension::D2,
+            format: SKYBOX_FORMAT,
+            usage: wgpu::TextureUsage::SAMPLED
+                | wgpu::TextureUsage::COPY_DST,
+        });
+
+        {
+            let mut encoder =
```
kvark: This encoder is dropped without being submitted. Can we just use `init_encoder` instead?

examples/skybox/skybox_vert.glsl
kvark: Doing `draw()` is what fires the shader, given that there is a pipeline set. Nothing else is required.
2) yes, same ones are fine (for 3-vertex triangle)

examples/skybox/main.rs
m4b: yes unfortunately it looks like the project hasn't adopted cargo fmt so i have to manually find these :(

examples/skybox/main.rs
m4b: good catch; so i just fixed this, and still not showing anything

m4b: the texture now shoes up in gpu profiling though, so that's a good sign

examples/skybox/main.rs
kvark: You are adding a new file, it costs you no friction to format it.

kvark:
Please add a screenshot to the readme (once ready), like ones we have for other examples.

examples/skybox/main.rs
m4b: > You are adding a new file, it costs you no friction to format it.

cargo fmt doesn't take files as arguments last i checked, unless something changed?

kvark: rustfmt does

m4b:
Ok i think this is ready, added skybox.gif, rustfmt'd, looks good on macOS, will test on linux in a bit.

README.md
```diff
@@ -9,6 +9,7 @@ This is an idiomatic Rust wrapper over [wgpu-native](https://github.com/gfx-rs/w
 
 ![Cube](etc/example-cube.png) ![Shadow](etc/example-shadow.png) ![MipMap](etc/example-mipmap.png)
 ![vange-rs](etc/vange-rs.png) ![Brawl](etc/brawl-attack.gif) ![GLX map](etc/glx-map.png)
+![Skybox](etc/skybox.gif)
```
kvark: let's move it up to the other examples, and rename according to how others are named: `etc/example-skybox.gif`

m4b: ok done

PR #113, opened 2019-11-04
build: exclude necessities for packaging, fixes #112

### LINKED ISSUE #112: etc and other non-essential build folders should be excluded
need to add an `includes` field in toml file; this will save a little bandwidth for initial download, especially not having to dl the gallery image files, etc.

m4b:
I believe this is the minimal set required as an external dep, e.g., don't need tests, etc., and other folders. I can't seem to find any explicit recommendations here though: https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields-optional

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ homepage = "https://github.com/gfx-rs/wgpu-rs"
 repository = "https://github.com/gfx-rs/wgpu-rs"
 keywords = ["graphics"]
 license = "MPL-2.0"
+exclude = ["etc/**/*", "examples/**/*", "tests/**/*", "Cargo.lock", "target/**/*"]
 
 [lib]
 


kvark:
Do other crates use `include` as well? Perhaps we should just `exclude` the stuff we clearly don't need instead.

m4b:
I use it; I’ve seen several other crates use it. Can switch to exclude too, I don’t have a major opinion, whitelists for rust libs seem to change less though, is only real motivating reason I’d say

kvark:
I believe `exclude` is mostly used in other established crates, e.g. https://github.com/alexcrichton/cc-rs/blob/1b97bc6df8eea6e817c46906db821a739d6d5c9a/Cargo.toml#L17

kvark:
@m4b could you follow-up with changes I requested?

m4b:
Whoops spaced on this. Ya might be able to get to it this weekend

kvark:
ping

PR #115, opened 2019-11-05
Handle error case for SwapChain::get_next_texture

antonok-edm:
Updates `SwapChain::get_next_texture` to account for https://github.com/gfx-rs/wgpu/pull/369. Otherwise, an invalid `Id` is returned and may cause confusing errors.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1351,14 +1351,20 @@ impl SwapChain {
     ///
     /// When the [`SwapChainOutput`] returned by this method is dropped, the swapchain will present
     /// the texture to the associated [`Surface`].
-    pub fn get_next_texture(&mut self) -> SwapChainOutput {
+    ///
+    /// Returns an `Err` if the GPU timed out when attempting to acquire the next texture.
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -1351,14 +1351,20 @@ impl SwapChain {
...
+    pub fn get_next_texture(&mut self) -> Result<SwapChainOutput, ()> {
         let output = wgn::wgpu_swap_chain_get_next_texture(self.id);
-        SwapChainOutput {
-            view: TextureView {
-                id: output.view_id,
-                owned: false,
-            },
-            swap_chain_id: &self.id,
+        if output.view_id == wgn::Id::ERROR {
+            Err(())
```
kvark: we should follow up with a better error here

antonok-edm: Since we're treating a zero-valued `Id` as an error, I'm not sure how you'd distinguish between any different errors that might occur. I figured the doc comment snippet is good enough until there's a more specific error-reporting mechanism in `wgpu-native`.

PR #120, opened 2019-11-10
prototype of futures for buffer mapping

swiftcoder:
This is largely in request-for-comment state right now, although it does work.

As written the `GpuFuture` implementation only works with types that are `Copy`, which rules out using a destructor to unmap the buffer. I'm assume it's possible to write a future type that can produce a non-Copy result, but I haven't sufficiently internalised how `Pin` works to see how.


examples/hello-compute/main.rs
```diff
@@ -90,9 +90,9 @@ fn main() {
 
     queue.submit(&[encoder.finish()]);
 
-    staging_buffer.map_read_async(0, size, |result: wgpu::BufferMapAsyncResult<&[u32]>| {
-        if let Ok(mapping) = result {
-            println!("Times: {:?}", mapping.data);
-        }
-    });
+    let result = futures::executor::block_on(staging_buffer.map_read::<u32>(0u64, size));
```
grovesNL: We'll need to find a good way for examples to switch between executors on native vs. web, and make sure other executors work too (for example, we can't use `block_on` on the web)

src/future.rs
```diff
@@ -0,0 +1,71 @@
...
+
+        // grab the device id and set the waker, but release the lock, so that the native callback can write to it
+        let device_id = {
+            let mut inner = arc.lock().unwrap();
+            inner.waker.replace(context.waker().clone());
+            inner.id
+        };
+
+        // polling the device should trigger the callback
+        wgn::wgpu_device_poll(device_id, true);
```
grovesNL: I don't know if it's possible, but ideally we should try to find a way to do this without using `wgpu_device_poll` and instead wait for the callback to be triggered outside of `poll` somehow.

`wgpu_device_poll` and especially the `force_wait` option are only available on wgpu-native at the moment and could even be removed entirely once we figure out how the event loop/callbacks will be exposed (see https://github.com/webgpu-native/webgpu-headers/issues/18)

examples/hello-compute/main.rs
swiftcoder: I was considering putting the entire example in an `async` function, and then we can use `await` throughout, and make main just an executor wrapper.

src/future.rs
swiftcoder: It seems to be tricky. As far as I can tell, polling the future is blocking the callback from firing on the current thread (it works fine if I await the future on a second thread).

We might have to sync the future-based API down into wgpu-native to make that work.

grovesNL: Yeah, I wonder if there are many examples of C callback APIs being wrapped into Rust futures out there to see how they're accomplishing it.

I'm not sure we'll be able to move many of these changes into wgpu-native, because on the web we'll have to [convert `Promise`s into `Future`s](https://rustwasm.github.io/wasm-bindgen/reference/js-promises-and-rust-futures.html) (instead of callbacks).

Cargo.toml
```diff
@@ -39,3 +39,4 @@ glsl-to-spirv = "0.1"
 log = "0.4"
 png = "0.15"
 winit = "0.20.0-alpha4"
+futures = "0.3"
```
kvark: nit: missing newline

examples/hello-compute/main.rs
```diff
@@ -29,14 +29,20 @@ fn main() {
     let cs = include_bytes!("shader.comp.spv");
     let cs_module = device.create_shader_module(&wgpu::read_spirv(std::io::Cursor::new(&cs[..])).unwrap());
 
-    let staging_buffer = device
-        .create_buffer_mapped(
-            numbers.len(),
-            wgpu::BufferUsage::MAP_READ
+    let staging_buffer = device.create_buffer(
```
kvark: why are we no longer using `create_buffer_mapped`?

examples/hello-compute/main.rs
```diff
@@ -90,9 +96,12 @@ fn main() {
...
-    });
+    if let Ok(mapping) = staging_buffer.map_read::<u32>(0u64, size).await {
+        println!("Times: {:?}", mapping.as_slice());
+    }
+    staging_buffer.unmap();
 }
+
+fn main() {
+    futures::executor::block_on(run());
+}
```
kvark: nit: missing newline

src/future.rs
```diff
@@ -0,0 +1,75 @@
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::{Arc, Mutex};
+use std::task::{Context, Poll, Waker};
+
+struct GpuFutureInner<T> {
```
kvark: We must be careful with naming here: a lot of asynchronouity/futures come from the fact that there is a browser and potential IPC boundary to the GPU process (on the CPU side), not necessarily GPU itself.

src/future.rs
```diff
@@ -0,0 +1,75 @@
...
+
+impl<T> Future for GpuFuture<T>
+where
+    T: Copy,
+{
+    type Output = T;
+
+    fn poll(self: Pin<&mut Self>, context: &mut Context) -> Poll<Self::Output> {
+        // grab a clone of the Arc
+        let arc = Pin::into_inner(self).inner.clone();
```
kvark: nit: let's use `Arc::clone(&...)` here, it's considered more explicit than just `clone()`

src/future.rs
```diff
@@ -0,0 +1,75 @@
...
+            inner.waker.replace(context.waker().clone());
+            inner.id
+        };
+
+        // polling the device should trigger the callback
+        wgn::wgpu_device_poll(device_id, true);
+
+        // now take the lock again, and check whether the future is complete
+        let inner = arc.lock().unwrap();
+        if inner.result.is_some() {
```
kvark: let's just do a `match`

src/future.rs
kvark: If we have wgpu-native exposing an event loop, then naturally `wgpu_device_poll` will go away

src/lib.rs
```diff
@@ -881,6 +954,96 @@ where
...
 impl Buffer {
+    /// Map the buffer for reading. The result is returned in a future.
+    /// The caller is responsible for unmapping the buffer once the future is completed.
+    pub fn map_read<T>(&self, start: BufferAddress, size: BufferAddress) -> GpuFuture<BufferMapReadResult<T>>
+    where
+        T: 'static + FromBytes,
+    {
+        let (future, completion) = future::new_gpu_future(self.device_id);
+
+        extern "C" fn buffer_map_read_future_wrapper<T>(
```
kvark: can this function use `T` type from the parent function?

src/lib.rs
```diff
@@ -881,6 +954,96 @@ where
...
+            start,
+            size,
+            buffer_map_write_future_wrapper::<T>,
+            Box::into_raw(user_data) as *mut u8,
+        );
+
+        future
+    }
+
     pub fn map_read_async<T, F>(&self, start: BufferAddress, size: BufferAddress, callback: F)
```
kvark: we'd need to remove these if we have futures

src/future.rs
swiftcoder: The good news for the web is that the `wasm-bindgen-futures` stuff should make converting between rust futures and JS promises pretty seamless.

examples/hello-compute/main.rs
swiftcoder: Mostly just so I could make sure both read and write worked. None of the samples have any real use-case for non-create map_write.

src/lib.rs
swiftcoder: No, see `rustc --explain E0401`.

swiftcoder:
I figured out what I was doing wrong re non-Copy types in futures, so we have the auto-unmap-on-drop behaviour back.

lachlansneff:
Might be useful to add a consuming `callback` method to `GpuFuture` that you can call instead of using a future runtime for convenience.

src/future.rs
swiftcoder: Suggestions re name? Right now it's really just Java's `CompletableFuture` plus a little bit of logic so the user doesn't have to remember to  invoke `wgpu_device_poll` manually in single-threaded scenarios.

kvark: ok, I don't have other naming candidates :)

swiftcoder:
I can pull the poll out and make the examples explicitly poll when they need to?

kvark:
Sounds in line with what @grovesNL suggests. Right, Josh?

grovesNL:
Yeah I think so, but I'm not actually sure which approach is better so I don't want to block the PR on this point :) One benefit of polling internally is that we don't have to worry about the event loop which is definitely nice.

We probably just need to investigate how other multi-threaded future-based APIs handle this (polling explicitly in the event loop vs. polling implicitly in the future vs. something else), and we can update the design later if there is an clear reason to prefer one of them.

swiftcoder:
Having an event loop do the polling is definitely what I'm used to from Python's async/await framework (and specifically, having a fairly generic event loop that can poll many different sources of futures - i.e. both network and gpu futures in the same loop).

kvark:
Ok, let's keep thinking, no need to block on that.
Bors r=grovesNL

swiftcoder:
Rebase again

PR #122, opened 2019-11-19
Wholesome update for the git master of wgpu-native

kvark:
Depends on https://github.com/gfx-rs/wgpu/pull/382

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,12 +20,19 @@ license = "MPL-2.0"
 [features]
 default = []
 # Make Vulkan backend available on platforms where it is by default not, e.g. macOS
-vulkan = ["wgn/gfx-backend-vulkan"]
+vulkan = ["wgn/vulkan-portability"]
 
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -20,12 +20,17 @@ license = "MPL-2.0"
...
 default = []
 # Make Vulkan backend available on platforms where it is by default not, e.g. macOS
-vulkan = ["wgn/gfx-backend-vulkan"]
+vulkan = ["wgn/vulkan-portability"]
 
 [dependencies.wgn]
 package = "wgpu-native"
 version = "0.4"
-features = ["local"]
+path = "/hub/moz/wgpu/wgpu-native"
```
grovesNL: Do we need to have the paths listed here?

kvark: no, this entry had to change to a git link anyway, so I left it unchanged from my local override
fixing now

kvark:
bors r=grovesNL

PR #126, opened 2019-11-21
Fix generic bounds on buffer mapping

kvark:
Smaller (and incomplete!) alternative to #119 while it's still discussed.
One missing piece here is alignment. cc @Coder-256

Also, importing `wgpu-core` as just "core" wasn't the best idea 😅 : it collides with the actual `core`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ version = "0.4"
 git = "https://github.com/gfx-rs/wgpu"
 rev = "73b33ea76e2f91b3114aa7640b1d60518d39f915"
 
-[dependencies.core]
+[dependencies.wgc]
 package = "wgpu-core"
 version = "0.1"
 git = "https://github.com/gfx-rs/wgpu"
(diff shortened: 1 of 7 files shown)


Coder-256:
I think this solves part of the issue, but why not just return a `&[u8]`? I'm not convinced that would be the best long-term solution, but it would partially fix the problem of alignment (since `[u8]` has an alignment of 1). Users can initialize their data from manual offsets with `from_ne_bytes`, which was [recently stabilized for floats](https://github.com/rust-lang/rust/pull/66002) (so it should be available on 1.40).

kvark:
@Coder-256 that would address the alignment flaw, yes, but it seems super inconvenient for the user, which I am currently being while dogfooding the library. It might be the way to proceed, I'm just not in a rush to call that while the very question of "Should we have mapping in WebGPU?" is being actively discussed by the working group.

If you want to help by making a PR that removes the generic stuff from mapping altogether (as well as drops `zerocopy` dependency), while providing some way or example for the users to operate (!), I'd be happy to merge it.

Coder-256:
@kvark I agree 100%, it's very inconvenient, but sadly I think it might be the only option. See my latest comment on #119. I'll start work on a PR. To be honest, I'm not 100% sure how it will turn out but I'll give it a shot and see how it affects the examples.

PR #127, opened 2019-11-21
Use u8 for buffer mapping

Coder-256:
cc @kvark @grovesNL

This is a temporary solution for #119, and a follow-up for #126.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,9 +1,7 @@
 //! A cross-platform graphics and compute library based on WebGPU.
 
 use arrayvec::ArrayVec;
-use zerocopy::{AsBytes, FromBytes, LayoutVerified};
 
-use std::convert::TryFrom;
 use std::ffi::CString;
 use std::ops::Range;
 use std::ptr;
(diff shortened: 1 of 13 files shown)


Coder-256:
You're on your own for testing `multithreaded_compute`... that thing butchered my poor laptop 😄

kvark:
>You're on your own for testing multithreaded_compute... that thing butchered my poor laptop

That deserves a separate bug, it's not expected!

src/lib.rs
```diff
@@ -808,32 +796,32 @@ impl Device {
...
 
-        let data = unsafe { std::slice::from_raw_parts_mut(ptr as *mut T, count) };
+        let data = unsafe { std::slice::from_raw_parts_mut(ptr as *mut u8, size) };
 
         CreateBufferMapped { id, data }
     }
 
+    /// Creates a new buffer, maps it into host-visible memory, copies data from the given slice,
+    /// and finally unmaps it, returning a [`Buffer`].
+    pub fn create_buffer_mapped_from(&self, data: &[u8], usage: BufferUsage) -> Buffer {
```
kvark: I think we should emove `mapped` from the name here, since the returned buffer is not in the mapped state. Perhaps, `create_buffer_with_data`?

src/lib.rs
```diff
@@ -883,50 +871,39 @@ impl<T> Drop for BufferAsyncMapping<T> {
...
 
-struct BufferMapReadAsyncUserData<T, F>
+struct BufferMapReadAsyncUserData<F>
 where
-    T: FromBytes,
-    F: FnOnce(BufferMapAsyncResult<&[T]>),
+    F: FnOnce(BufferMapAsyncResult<&[u8]>),
 {
-    size: BufferAddress,
+    size: usize,
```
kvark: why is this type changed?

Coder-256: To get rid of this:

https://github.com/gfx-rs/wgpu-rs/blob/a5921daeb495263b205ebe41aa8a106c3c7fb65e/src/lib.rs#L925

`BufferMapXxxAsyncUserData` is only used to pass info to the C callback in the respective `map_xxx_async`, and the input to those functions is already a `usize`. The current code converts to `BufferAddress` and back to `usize`.

src/lib.rs
Coder-256: Done.

kvark:
Bors r+

PR #128, opened 2019-11-21
Small updates

Coder-256:
Enable Travis CI cache, disable [deprecated container-based builds](https://docs.travis-ci.com/user/reference/trusty/#container-based-infrastructure), and add `red.png` to `.gitignore` (output from `cargo run --example capture`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: rust
-sudo: false
+cache: cargo
 os:
   - linux
   - osx
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ Cargo.lock
 
 # Other
 .DS_Store
+
+# Output from capture example
+red.png


.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: rust
-sudo: false
+cache: cargo
```
kvark: are you sure that's a good idea? My understanding was that the cache is huge in size and the costs of keeping it are not always justified.

.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: rust
-sudo: false
```
kvark: TIL!

.travis.yml
Coder-256: I'm not sure. I hadn't heard that, I had just assumed it was better from experience with other languages. It looks like there is a mix of projects that use `cache: cargo`, some that cache only `$HOME/.cargo/` and some that don't use caching at all, I'm not sure which is better.

kvark:
alright, let's keep an eye on it
bors r+

.travis.yml
rukai: The problem with caching is cargo never cleans up old files built from old rustc versions or old dependencies.

PR #144, opened 2019-12-28
Adds compute-boids example from gpuweb demos

masonblier:
Port of the computeBoids.ts demo from https://github.com/austinEng/webgpu-samples/blob/master/src/examples/computeBoids.ts  using separate compute and render passes with shared buffers bound by layout location. 

Screen Shot 2019-12-27 at 9 53 17 PM

I'm happy to receive any feedback related to these changes and if there's any interest in merging it to the main repository.

Thanks,
Mason


examples/compute-boids/main.rs
```diff
@@ -0,0 +1,391 @@
...
+    // buffer for the three 2d triangle vertices of each instance
+
+    let vertex_buffer_data: Vec<f32> = [-0.01, -0.02, 0.01, -0.02, 0.00, 0.02].to_vec();
+    let vertices_buffer = device.create_buffer_with_data(vertex_buffer_data.as_bytes(), 
+        wgpu::BufferUsage::VERTEX | wgpu::BufferUsage::COPY_DST);
+
+
+    // buffer for simulation parameters uniform
+
+    let sim_param_data: Vec<f32> = [
```
kvark: nit: you can instead do:
```rust
let sim_param_data = [
  0.04f32, // ...
  0.1, // ...
  0.025, ...
  ...
];
```
There is no need to make it a `Vec`, and there is no need for explicit `<f32>` if you can just make the first member `0.04f32`

examples/compute-boids/main.rs
```diff
@@ -0,0 +1,391 @@
...
+    ].to_vec();
+    let sim_param_buffer = device.create_buffer_with_data(sim_param_data.as_bytes(), 
+        wgpu::BufferUsage::UNIFORM | wgpu::BufferUsage::COPY_DST);
+
+
+    // buffer for all particles data of type [(posx,posy,velx,vely),...]
+
+    let mut initial_particle_data: Vec<f32> = vec![0.0; (4 * NUM_PARTICLES) as usize];
+    for i in 0..NUM_PARTICLES {
+        initial_particle_data[(4 * i + 0) as usize] = 2.0 * (rand::random::<f32>() - 0.5);
```
kvark: This could probably be written much shorter, something like a [f32; 4] being repeated NUM_PARTICLES times.
Something more concrete and not as extreme, we could do:
```rust
for chunk in initial_particle_data.chunks_mut(4) {
  chunk[0] = ...;
  chunk[1] = ...;
  chunk[2] = ...;
  chunk[3] = ...;
}
```

examples/compute-boids/main.rs
```diff
@@ -0,0 +1,391 @@
...
+                | wgpu::BufferUsage::COPY_DST)
+        );
+    }
+    
+    
+    // create compute bind group for each particle buffer
+
+    let mut particle_bind_groups = Vec::<wgpu::BindGroup>::new();
+    for i in 0..particle_buffers.len() {
+        particle_bind_groups.push(
```
kvark: can we do this in the same loop where we are creating the buffers?

examples/compute-boids/main.rs
```diff
@@ -0,0 +1,391 @@
...
+                // TODO dont recreate this every loop (how to borrow swap_chain?)
+                let render_pass_descriptor = wgpu::RenderPassDescriptor {
+                    color_attachments: &[wgpu::RenderPassColorAttachmentDescriptor {
+                        attachment: &frame.view,
+                        resolve_target: None,
+                        load_op: wgpu::LoadOp::Clear,
+                        store_op: wgpu::StoreOp::Store,
+                        clear_color: wgpu::Color::BLACK,
+                    }],
+                    depth_stencil_attachment: Some(wgpu::RenderPassDepthStencilAttachmentDescriptor {
```
kvark: do we need the depth for boids?

examples/compute-boids/main.rs
```diff
@@ -0,0 +1,391 @@
...
+
+                let mut command_encoder =
+                    device.create_command_encoder(&wgpu::CommandEncoderDescriptor { todo: 0 });
+                
+                {
+                    // compute pass
+                    let mut cpass = command_encoder.begin_compute_pass();
+                    cpass.set_pipeline(&compute_pipeline);
+                    cpass.set_bind_group(0, &particle_bind_groups[frame_num % 2], &[]);
+                    cpass.dispatch(NUM_PARTICLES, 1, 1);
```
kvark: is every particle being processed as a separate compute group? that seems inefficient

examples/compute-boids/main.rs
```diff
@@ -0,0 +1,391 @@
...
+                }
+
+                {
+                    // render pass
+                    let mut rpass = command_encoder.begin_render_pass(&render_pass_descriptor);
+                    rpass.set_pipeline(&render_pipeline);
+                    rpass.set_vertex_buffers(0, &[(&particle_buffers[(frame_num + 1) % 2], 0)]); // render dst particles
```
kvark: let's bind both buffers at once, i.e.
```rust
rpass.set_vertex_buffers(0, &[
  (&buffer0, offset0),
  (&buffer1, offset1),
]);
```

examples/compute-boids/main.rs
masonblier: @kvark In this case we can't because each binding group references both buffers, so they must both be created ahead of time.

examples/compute-boids/main.rs
masonblier: @kvark I'm somewhat novice to compute programming, I tried an attempt using PARTICLES_PER_INVOCATION and a for loop in the compute shader. I did these in a separate commit. I wasn't able to test the performance differences with this, rather my laptop (osx) would fail with GPU stuck errors when I increased the PARTICLES_PER_INVOCATION number too large.

examples/boids/boids.comp
```diff
@@ -0,0 +1,79 @@
+#version 450
+
+// this shader expects NUM_PARTICLES and PARTICLES_PER_INVOCATION
```
kvark: It looks like you are misunderstanding the computational model exposed by WebGPU/Vulkan. The shader `main()` function would work for a compute thread, it rarely need to loop over work items itself.
The shader also needs to define how many threads are ran per group, like:
```
layout(local_size_x = 64) in;
```
The default is 1, and this is the value used here for the lack of the explicit qualifier. This is in no way representing any practical workflow, since the threads in a group can share local data, which is what most programs try to take advantage of.

If you don't care what the thread group size needs to be, picking 64 is a safe choice, since it matches the largest wavefront length seen on the hardware.

examples/boids/main.rs
```diff
@@ -369,15 +354,17 @@ fn main() {
                     let mut cpass = command_encoder.begin_compute_pass();
                     cpass.set_pipeline(&compute_pipeline);
                     cpass.set_bind_group(0, &particle_bind_groups[frame_num % 2], &[]);
-                    cpass.dispatch(NUM_PARTICLES, 1, 1);
+                    cpass.dispatch(INVOCATION_COUNT, 1, 1);
```
kvark: this specifies the number of thread **groups**. So if each group is 64x1x1, then the group count will be NUM_PARTICLES / 64, rounded up.

examples/boids/boids.comp
masonblier: @kvark Thanks, I was definitely misunderstanding the parameters. With the change you suggested I now see the performance gains.

masonblier:
@kvark No worries, all squashed. Thanks so much for the review, it was a great learning experience!

kvark:
Great, thank you!
Bors r+

PR #145, opened 2019-12-28
Mark CommandEncoder as !Send !Sync

LaylConway:
These depend on the thread-specific command pool they're allocated from, which requires only one thread accessing them at a time, so they can't be moved to other threads without additional synchronization.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -236,6 +236,9 @@ pub struct CommandBuffer {
 #[derive(Debug)]
 pub struct CommandEncoder {
     id: wgc::id::CommandEncoderId,
+    /// This type should be !Send !Sync, because it represents an allocation on this thread's
+    /// command buffer.
+    _p: std::marker::PhantomData<*const u8>,
 }
 
 /// An in-progress recording of a render pass.
@@ -599,6 +602,7 @@ impl Device {
     pub fn create_command_encoder(&self, desc: &CommandEncoderDescriptor) -> CommandEncoder {
         CommandEncoder {
             id: wgn::wgpu_device_create_command_encoder(self.id, Some(desc)),
+            _p: Default::default(),
         }
     }
 


src/lib.rs
```diff
@@ -236,13 +236,19 @@ pub struct CommandBuffer {
...
 }
 
 /// An in-progress recording of a render pass.
 #[derive(Debug)]
 pub struct RenderPass<'a> {
     id: wgc::id::RenderPassId,
     _parent: &'a mut CommandEncoder,
+    /// This type should be !Send !Sync, because it represents an allocation on this thread's
+    /// command buffer.
+    _p: std::marker::PhantomData<*const u8>,
```
kvark: is this needed? I thought, since it borrows the encoder, it has the same restrictions

src/lib.rs
```diff
@@ -236,13 +236,19 @@ pub struct CommandBuffer {
 #[derive(Debug)]
 pub struct CommandEncoder {
     id: wgc::id::CommandEncoderId,
+    /// This type should be !Send !Sync, because it represents an allocation on this thread's
+    /// command buffer.
+    _p: std::marker::PhantomData<*const u8>,
```
kvark: I think it's more idiomatic to do
```rust
impl !Sync for CommandEncoder {}
impl !Send for CommandEncoder {}
```

LaylConway: That was my first try, but it results in the following error:
```
negative trait bounds are not yet fully implemented; use marker types for now

note: for more information, see https://github.com/rust-lang/rust/issues/13231rustc(E0658)
```

src/lib.rs
LaylConway: I'll remove that, it's not strictly necessary yea

src/lib.rs
kvark: indeed! What's confusing to me is - what "marker types" does the error mean? I wasn't able to find those.

kvark: the rustonomicon [mentions](https://doc.rust-lang.org/nomicon/send-and-sync.html) that a feature needs to be enabled:
```rust
#![feature(optin_builtin_traits)]

// I have some magic semantics for some synchronization primitive!
struct SpecialThreadToken(u8);

impl !Send for SpecialThreadToken {}
impl !Sync for SpecialThreadToken {}
```

It doesn't tell if nightly is required though

kvark: another small nit: we could have `_non_send_sync: *const ()`, which is marginally simpler than the phantom data. I believe the phantom data has semantics about stuff that needs to be logically present, so having it here implies a pointer somewhere, but there isn't one.

LaylConway: `*const ()` is 8 bytes, while `PhantomData<_>` is always 0 bytes, which is why I went with PhantomData. I think that's what they intend with "marker types" (it's in the marker module), but it could definitely be clearer on what's idiomatic Rust right now.

kvark: derp, you are right :)

PR #148, opened 2019-12-29
Change Queue::submit to take &self

LaylConway:
This small change allows multiple threads to submit buffers slightly more efficiently, by being able to do it concurrently without a lock on Queue.

In practice currently `Queue::submit` does a lot of locking internally for most of the function's duration, so it's only a small performance win right now, but it opens up the road for more improvements.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ rev = "b51053dc2dc3bbe9b2ba050fde42eeb6405fe092"
 
 [dependencies]
 arrayvec = "0.5"
+smallvec = "1"
 raw-window-handle = "0.3"
 
 [dev-dependencies]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,6 +5,7 @@ use future::GpuFutureCompletion;
 pub use future::GpuFuture;
 
 use arrayvec::ArrayVec;
+use smallvec::SmallVec;
 
 use std::ffi::CString;
 use std::ops::Range;
@@ -261,7 +262,6 @@ pub struct ComputePass<'a> {
 #[derive(Debug)]
 pub struct Queue {
     id: wgc::id::QueueId,
-    temp_command_buffers: Vec<wgc::id::CommandBufferId>,
 }
 
 /// A resource that can be bound to a pipeline.
@@ -569,7 +569,6 @@ impl Adapter {
         };
         let queue = Queue {
             id: wgn::wgpu_device_get_queue(device.id),
-            temp_command_buffers: Vec::new(),
         };
         (device, queue)
     }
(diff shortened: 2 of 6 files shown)


src/lib.rs
```diff
@@ -1382,14 +1380,13 @@ impl<'a> Drop for ComputePass<'a> {
 
 impl Queue {
     /// Submits a series of finished command buffers for execution.
-    pub fn submit(&mut self, command_buffers: &[CommandBuffer]) {
-        self.temp_command_buffers.clear();
-        self.temp_command_buffers
-            .extend(command_buffers.iter().map(|cb| cb.id));
+    pub fn submit(&self, command_buffers: &[CommandBuffer]) {
+        let temp_command_buffers: Vec<_> =
```
kvark: since we aren't caching these, let's at least have a `SmallVec<>` here?

src/lib.rs
```diff
@@ -1382,14 +1380,13 @@ impl<'a> Drop for ComputePass<'a> {
 
 impl Queue {
     /// Submits a series of finished command buffers for execution.
-    pub fn submit(&mut self, command_buffers: &[CommandBuffer]) {
-        self.temp_command_buffers.clear();
-        self.temp_command_buffers
-            .extend(command_buffers.iter().map(|cb| cb.id));
+    pub fn submit(&self, command_buffers: &[CommandBuffer]) {
```
kvark: upon some thinking, I think this is safe to do. We just need to make sure the expectations, especially with regards to threading, are well defined by wgpu-native and wgpu-core

LaylConway:
I noticed in `begin_render_pass`, `SmallVec` is used. Is that intentional? It seems to limit it to at most 4 color attachments.

src/lib.rs
LaylConway: It seems to me most functions that are currently `&mut seulf` could be safely changed to `&self` right now, because wgpu-core will acquire locks anyways.

kvark: `&mut self` also makes sure the access is synchronized, i.e. you don't want the clients to concurrently record to the same pass, for example. This is directly related to the non-send non-sync enforcement that was put in for the command encoders.
Even if we can, I don't think anything with regards to command recording should be changed to `&self`.

Cargo.toml
```diff
@@ -36,6 +36,7 @@ rev = "b51053dc2dc3bbe9b2ba050fde42eeb6405fe092"
 
 [dependencies]
 arrayvec = "0.5"
+smallvec = "1.1.0"
```
kvark: Do you need "1.1.0" specifically? If not, let's just keep it "1" or "1.0"

src/lib.rs
LaylConway: That makes sense yea

PR #152, opened 2020-01-04
Update to latest winit (0.20.0-alpha6)

grovesNL:
Updated to latest winit to fix examples:
- `RedrawRequested` is now used for rendering
- `EventsCleared` replaced by `MainEventsCleared` which requests redraw
- Removed Metal auto-capture and escape key handling from hello-triangle to simplify the example slightly (it doesn't use `framework`, so we should try to make it as simple as possible IMO)

Should we also remove the `feature = gl` parts of the examples to simplify them until GL is reenabled?


grovesNL:
Looks like I'm hitting an issue with the Travis cache size which causes the build to timeout (discussed in #128 a bit)

grovesNL:
Added a commit to (temporarily?) disable the cache

PR #166, opened 2020-01-24
Reduce Rust requirements for Pin usage

kvark:
That lowers Rust requirement from 1.39 to 1.33
Important, since Ubuntu packages are still on 1.37


kvark:
bors r=auto

PR #168, opened 2020-01-31
Enforce life times of render pass attachments

### LINKED ISSUE #167: Panic when TextureView drops before RenderPass that uses it
I had a panic, where I was dropping the depth attachment TextureView before the RenderPass was submitted.

I suppose this is a bug in the lifetime types somewhere

This was using trunk as of 30th Jan 2020, after the bind_group lifetime changes

kvark:
Fixes #167


kvark:
should be safe
bors r+

PR #169, opened 2020-02-02
Disable window bitmap redirection

kvark:
That sets `WS_EX_NOREDIRECTIONBITMAP`. According to @raphlinus, makes our resize better :)
There are still strange pauses I see on resizing with DX12 (with and without the change), to be followed-up.

### DIFF
--- a/examples/framework.rs
+++ b/examples/framework.rs
@@ -64,8 +64,14 @@ pub fn run<E: Example>(title: &str) {
 
     #[cfg(not(feature = "gl"))]
     let (window, size, surface) = {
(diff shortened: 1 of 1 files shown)


kvark:
I think it's working, based on the tests so far.
bors r+

TannerRogalsky:
@kvark This has broken all the examples that use `framework` for me.

Windows 10 Pro. NVIDIA GeForce GTX 970.

image

kvark:
Weird, I tested on Windows 10 Intel.
Lets back this out.

PR #179, opened 2020-02-19
Add vertex_attr_array macro

dhardy:
Specifying `VertexAttributeDescriptor` arrays is tedious, and for no good reason. This PR adds a macro to do the job.

All examples tested. Documentation is short but better than the standard for this crate I feel. :wink:

### DIFF
--- a/examples/boids/main.rs
+++ b/examples/boids/main.rs
@@ -9,6 +9,8 @@ mod framework;
 use std::fmt::Write;
 use zerocopy::{AsBytes};
 
+use wgpu::vertex_attr_array;
+
 
 // number of boid particles to simulate
 
@@ -123,32 +125,12 @@ impl framework::Example for Example {
                 wgpu::VertexBufferDescriptor {
                     stride: 4 * 4,
                     step_mode: wgpu::InputStepMode::Instance,
-                    attributes: &[
-                        // instance position
-                        wgpu::VertexAttributeDescriptor {
-                            offset: 0,
-                            format: wgpu::VertexFormat::Float2,
-                            shader_location: 0,
-                        },
-                        // instance velocity
-                        wgpu::VertexAttributeDescriptor {
-                            offset: 2 * 4,
-                            format: wgpu::VertexFormat::Float2,
-                            shader_location: 1,
-                        },
-                    ]
+                    attributes: &vertex_attr_array![0 => Float2, 1 => Float2],
                 },
                 wgpu::VertexBufferDescriptor {
                     stride: 2 * 4,
                     step_mode: wgpu::InputStepMode::Vertex,
-                    attributes: &[
-                        // vertex positions
-                        wgpu::VertexAttributeDescriptor {
-                            offset: 0,
-                            format: wgpu::VertexFormat::Float2,
-                            shader_location: 2,
-                        },
-                    ]
+                    attributes: &vertex_attr_array![2 => Float2],
                 },
             ],
             sample_count: 1,
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1,0 +1,77 @@
+//! Convenience macros
+
+/// Macro to produce an array of [`VertexAttributeDescriptor`]
+///
+/// Output has type: `[VertexAttributeDescriptor; _]`. Usage is as follows:
+/// ```
+/// # use wgpu::vertex_attr_array;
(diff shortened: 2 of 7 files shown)


examples/boids/main.rs
```diff
@@ -123,32 +125,12 @@ impl framework::Example for Example {
...
-                            shader_location: 0,
-                        },
-                        // instance velocity
-                        wgpu::VertexAttributeDescriptor {
-                            offset: 2 * 4,
-                            format: wgpu::VertexFormat::Float2,
-                            shader_location: 1,
-                        },
-                    ]
+                    attributes: &vertex_attr_array![Float2, Float2],
```
kvark: holy smokes, isn't this tidy!

src/macros.rs
```diff
@@ -0,0 +1,73 @@
+//! Convenience macros
+
+/// Macro to produce an array of [`VertexAttributeDescriptor`]
+///
+/// Output has type: `[VertexAttributeDescriptor; _]`. Usage is as follows:
+/// ```
+/// let attrs = vertex_attr_array![Float2, Float, Ushort4];
+///
+/// // In case the shader location does not start from 0, it may be specified:
```
kvark: So the limitation here is that all of the shader attributes have to be continuous in terms of shader locations. This is ok for our examples, and likely OK for other applications, but it introduces the restrictions we shouldn't have otherwise.

Would you be open to refactor the macro so that the locations are always explicitly provided for each attribute?

src/macros.rs
```diff
@@ -0,0 +1,73 @@
...
+                format: $crate::VertexFormat :: $item,
+                offset: $off,
+                shader_location: $loc,
+            },];
+            $off + vertex_attr_array!(size $item);
+            $loc + 1;
+            $($more,)*
+        )
+    };
+    (size Uchar2) => { 2 };
```
kvark: should this be a different macro?

src/macros.rs
dhardy: Sure can do. If you can suggest syntax, even better.

As I understand it, we are always counting from a specified start-point, so maybe something like `[2..; Float2, Float3]` would be clear enough?

kvark: we aren't counting anything, we are just providing locations specified in the shaders, they can be anything

src/macros.rs
dhardy: Possibly yes. Whether it should be visible to users is another matter — perhaps it is useful?

kvark: let's make it private for now :)

dhardy:
There may be other places macros would be useful — I haven't actively looked. Feel free to ask if you'd like some additional ones — ideally with a syntax suggestion.

src/macros.rs
dhardy: Aha, so a format like `[2, Float2; 3, Ushort4]` is preferable?

kvark: yes, it looks good to me

dhardy: I implemented a slightly different syntax, in part to be more "rustic" (Rust does not have map literals) and in part because `:` cannot follow an `expr` (which would restrict the index to being a literal).

HalfVoxel:
Are you now going to go with the macro that requires you to always specify the offsets manually?
Isn't it good to make the macro handle the (in my mind most common case) where all the items are consecutive and with no gaps? Specifying explicit offsets is very useful of course, but it seems like it would be nicer if the common case is as easy as possible?

kvark:
@HalfVoxel it's not the offsets that are specified manually (in fact, they are always implicit in this version of the macro), it's the shader locations.

HalfVoxel:
@kvark Ah. That makes sense.

gillesboisson:
hello, 
I'm playing around wgpu and trying to implement the example using wgpu as external crate.
But I can't access as its in the example project.
I know this is not really the place to ask for help but I couldn't find where to ask questions.
cheers

kvark:
It will work if you depend on the git version of wgpu.

gillesboisson:
I use the 0.4.0 crate

dhardy:
Like this:
```toml
wgpu = { git = "https://github.com/gfx-rs/wgpu-rs" }
```

gillesboisson:
You confirmed what I tried finally, it works fine now. cheers

Issue #101, opened 2019-10-25
Target the Web directly


kvark:
We need to make wgpu-rs do the actual Web calls and run in the browser supporting WebGPU API. The API itself is still evolving, but we should start exploring what our side would look like, and maybe we'll already get something running.

grovesNL:
Steps to get this working:

1. Ensure mozilla-central has up-to-date WebGPU WebIDL
   - This step can probably be omitted, but technically this is the upstream for many of the existing web-sys WebIDL files
2. Update the existing WebGPU WebIDL in web-sys so we can generate Rust bindings
   - There is actually an older version already in web-sys at https://github.com/rustwasm/wasm-bindgen/blob/master/crates/web-sys/webidls/enabled/WebGPU.webidl which is a great sign that this won't be too bad
3. Split calls in wgpu-rs to use `#[cfg(target_arch = "wasm32")]` to target web-sys, and `#[cfg(not(target_arch = "wasm32"))]` to target wgpu-native
   - We might want to consider splitting wgpu-rs into two backends (wgpu-native and web-sys), or use a trait like we do in glow, or something else
4. Now we can build for wasm32-unknown-unknown and use wasm-bindgen-cli or wasm-pack as usual
   - See https://github.com/rust-gamedev/wg/issues/51#issuecomment-527695166 – I don't know what the best practices are for this step, but using wasm-bindgen directly should be fine

kvark:
Thank you for describing the steps! They look very actionable, and make me believe we can get a prototype running (at least device creation) by the end of the year.

FYI, (1) was mostly updated to latest WebIDL by https://phabricator.services.mozilla.com/D46166, which took more of my blood and tears than I'd hope or expect.

kvark:
Gecko PR has landed, and now one can at least create a device on WebGPU (in Nightly!). I think the work can be started here, according to @grovesNL plan.

grovesNL:
I also already have some branches for web-sys changes and wgpu-rs changes for this – just need to update them based on the recent changes and then we can start making some progress here.

grovesNL:
Small update: https://github.com/rustwasm/wasm-bindgen/pull/1997 has been merged which allows us to use web-sys bindings for WebGPU while the WebIDL is still unstable

kvark:
Let's consider this fixed by #193 
Amazing work of @grovesNL and everyone involved, thank you all!

PR #193, opened 2020-03-10
Add web backend

grovesNL:
This branch has a bunch of WIP changes as I work through #101 

The branch is still very unstable but any feedback about the approach so far would be great.

I tried to use a backend trait initially (with associated types and functions) but it didn't work great here. For example, `async fn` can't be used on traits yet.

Instead, with the current approach we have to manually ensure that all type signatures match. This seems mostly ok because we wrap most types/functions in wgpu-rs anyway, but we do have to be careful wherever we directly expose types. For example, if an exposed type is `Copy` in one backend and not the other, we'd probably have to newtype them (probably relevant for all of the `XyzId` types).


examples/hello-compute/main.rs
```diff
@@ -1,19 +1,8 @@
...
 use zerocopy::AsBytes as _;
 
 async fn run() {
-    env_logger::init();
-
-    // For now this just panics if you didn't pass numbers. Could add proper error handling.
-    if std::env::args().len() == 1 {
-        panic!("You must pass a list of positive integers!")
-    }
-    let numbers: Vec<u32> = std::env::args()
```
kvark: we could move that into `main` and have `run` receiving this slice in an argument?

src/backend/web.rs
```diff
@@ -0,0 +1,464 @@
...
+        IndexFormat::Uint16 => f::Uint16,
+        IndexFormat::Uint32 => f::Uint32,
+    }
+}
+
+fn map_vertex_format(format: wgt::VertexFormat) -> web_sys::GpuVertexFormat {
+    use wgt::VertexFormat;
+    use web_sys::GpuVertexFormat as vf;
+    match format {
+        VertexFormat::Uchar2 => vf::Uchar2,
```
kvark: I wonder if these maps are truly necessary. Perhaps, we can always say that our enum values match the target WebIDL, and do a transmute?

grovesNL: These are strings in WebIDL, see https://docs.rs/web-sys/0.3.36/src/web_sys/features/gen_GpuVertexFormat.rs.html#4

examples/hello-compute/main.rs
grovesNL: Yeah we could easily add this back in. I just removed it for now because we need to pick some values at compile-time for wasm anyway

grovesNL:
I was able to run hello-compute in Firefox Nightly with the latest changes (top wasm, bottom native):

wasm-native

grovesNL:
And now hello-triangle is able to run on the web too:

wasm-triangle

(it's upside-down due to https://github.com/gfx-rs/wgpu/issues/519)

examples/hello-triangle/main.rs
```diff
@@ -60,7 +62,7 @@ async fn run() {
         }),
         primitive_topology: wgpu::PrimitiveTopology::TriangleList,
         color_states: &[wgpu::ColorStateDescriptor {
-            format: wgpu::TextureFormat::Bgra8UnormSrgb,
+            format: wgpu::TextureFormat::Bgra8Unorm,
```
mitchmindtree: Oh interesting. Is it now the case that `Bgra8Unorm` is assumed to be sRGB?

grovesNL: Nope both `Bgra8Unorm` and `Bgra8UnormSrgb` still exist – this is just a temporary hack for the hello-triangle example because Chrome Canary didn't seem to allow me to use `Bgra8UnormSrgb` for the swapchain :) I haven't looked into it yet but a swapchain with `Bgra8UnormSrgb` is supposed to work fine in browsers as far as I know.

Songtronix:
If you want to use async traits there is [async-trait](https://crates.io/crates/async-trait). Supports a wide range of usecases but has it's limitations.

grovesNL:
msaa-line is working now too:

image

grovesNL:
The next working example is cube:

image

grovesNL:
Skybox!
image

grovesNL:
All examples are available at http://joshgroves.com/gfx/wgpu/examples/?example=hello-triangle for now.

Each example is built in `debug` and much larger than the final wasm files will be (built in `release`/stripped of unused symbols). There are also still some issues with the winit integration, so the wasm file has to be fully loaded before sending any events to the canvas. `ControlFlow::Poll` isn't being used yet so interaction (e.g. by moving the cursor over the canvas) is needed to continue the event loop.

Both Chrome Canary and Firefox Nightly have issues running all examples at the moment, but eventually all examples should work fine 🤞

expenses:
> Both Chrome Canary and Firefox Nightly have issues running all examples at the moment, but eventually all examples should work fine crossed_fingers

No luck on Safari here either.

grovesNL:
@expenses I also forgot to mention that these examples are currently using SPIR-V, which won't work in Safari. We'll eventually need to support WGSL here, which will allow Safari to run them (once WebGPU support is ready in Safari in general)

leo60228:
http://joshgroves.com/gfx/wgpu/examples/?example=hello-triangle panics:
```
panicked at 'expected to get context from canvas', /Users/josh/.rustup/toolchains/nightly-x86_64-apple-darwin/lib/rustlib/src/rust/src/libstd/macros.rs:13:23
````
(`wgpu::backend::web::device_create_surface::h8363acb00c4a0b19@http://joshgroves.com/gfx/wgpu/examples/hello-triangle.js line 278 > WebAssembly.instantiate:wasm-function[195]:0x1fc56`)
I'm on Firefox Beta, not Firefox Nightly. Are necessary features explicitly locked to Nightly?

**Comment by leo60228** [comment]
I have the same issue on Nightly (77.0a1, 2020-04-13).

**Comment by Dispersia** [comment]
Did you go to about:config and enable the webgpu feature? It couldn't get a context from canvas, so it's in your browser, wouldn't be this project

**Comment by leo60228** [comment]
After enabling that I get this:
```
panicked at 'Unable to get adapter: JsValue(InvalidStateError: No matching adapter found!
)', /Users/josh/Projects/wgpu-rs/src/backend/web.rs:53:9
```

**Comment by leo60228** [comment]
The W3C demos have the same issue, hmmmmmmmm........

**Comment by grovesNL** [comment]
For Firefox Nightly you need to enable WebGPU and WebRender in `about:config`:

- set `dom.webgpu.enabled` to `true`
- set `gfx.webrender.all` to `true`

Then restart Nightly and try to run the examples (expect many crashes/issues rendering). Sometimes the canvas also has to be clicked on or resized in order to begin rendering – still need to look into that.

For Chrome Canary you can enable WebGPU under `chrome://flags`:

- set `Unsafe WebGPU` to `Enabled`

Expect crashes/issues rendering with Canary too :)

**Comment by kvark** [comment]
@grovesNL what is your status? What are the concerns/blockers right now?

**Comment by grovesNL** [comment]
@kvark it's mostly ready now if you'd like to start reviewing it. There are a few small `TODO`s left to cleanup before merging, and labels aren't passed to the web backend yet.

File: examples/framework.rs [hunk 4]
```diff
@@ -120,7 +76,7 @@ async fn run_async<E: Example>(title: &str) {
 
     let mut sc_desc = wgpu::SwapChainDescriptor {
         usage: wgpu::TextureUsage::OUTPUT_ATTACHMENT,
-        format: wgpu::TextureFormat::Bgra8UnormSrgb,
+        format: wgpu::TextureFormat::Bgra8Unorm,
```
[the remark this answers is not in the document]
grovesNL: Agreed, I mentioned this a bit in https://github.com/gfx-rs/wgpu-rs/pull/193#discussion_r392665836 – I wasn't sure what the plans are for supported swapchain formats and whether srgb would be supported initially on the web (e.g. https://github.com/WICG/canvas-color-space).

I'm not sure if there's a way to query whether srgb is available for the swapchain. We do have `getSwapChainPreferredFormat` which could give a weak signal of whether srgb is supported for the preferred format. I guess the only fall back path is to use `cfg` for now, although I was trying to avoid `cfg` in examples.

kvark:
Please proceed when ready. On my side, everything seems to be working. There are (presumably) `winit` related warnings, and that's it.

examples/framework.rs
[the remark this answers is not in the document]
grovesNL: I used `cfg` for now

grovesNL:
bors r=kvark

grovesNL:
@kvark I need to merge this manually due to https://forum.bors.tech/t/resource-not-accessible-by-integration/408/3 – I'm updating the CI actions but bors isn't allowed to merge those changes

Keavon:
🎉

grovesNL:
For anyone interested in continuing to follow the progress here, I'll keep updating the examples at http://joshgroves.com/gfx/wgpu/examples/.

Some of the recent fixes from @kvark should be landing in Firefox Nightly soon ([enable WebGPU using the instructions above](https://github.com/gfx-rs/wgpu-rs/pull/193#issuecomment-613182772)) which should get most of these examples working.

Keavon:
Should `Currently wgpu-rs works on native platforms, but WASM support is currently being added as well.` be removed from the project's readme?

grovesNL:
@Keavon yeah, we should be able to remove that and clarify that both native and wasm are supported now :+1: Would you like to submit a PR to update it?

Keavon:
I haven't been following this closely enough to know what would be most appropriate to write so I'd probably feel better not doing the PR myself.

grovesNL:
Maybe we could replace it with something like

> wgpu-rs can be run both natively (using wgpu-core and gfx-hal) and on the web (using the WebGPU API from wasm).

PR #197, opened 2020-03-13
Update for the latest spec changes

kvark:
Blocked on https://github.com/gfx-rs/wgpu/pull/518

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,15 +27,22 @@ vulkan = ["wgn/vulkan-portability"]
 package = "wgpu-native"
 version = "0.4"
 git = "https://github.com/gfx-rs/wgpu"
-rev = "429ca1d4460becd20d771f67024878948b60454b"
+rev = "39f17e50754aba6beeeabdd868ddfd700f9710c5"
 #path = "../wgpu/wgpu-native"
 
 [dependencies.wgc]
 package = "wgpu-core"
 version = "0.1"
 git = "https://github.com/gfx-rs/wgpu"
-rev = "429ca1d4460becd20d771f67024878948b60454b"
+rev = "39f17e50754aba6beeeabdd868ddfd700f9710c5"
 #path = "../wgpu/wgpu-core"
+
+[dependencies.wgt]
+package = "wgpu-types"
+version = "0.1"
+git = "https://github.com/gfx-rs/wgpu"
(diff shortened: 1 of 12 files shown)


kvark:
One of the tricky things here is how the size is provided to `set_vertex_buffer` and `set_index_buffer`. We basically have the following choices:
  1. require both the offset and the size be meaningful
  2. treat size=0 as the remaining part of the buffer (current version of this PR)
  3. be generic over [RangeBounds](https://doc.rust-lang.org/stable/std/ops/trait.RangeBounds.html)

I believe (3) is the nicest for the users, but the problem here is inconsistency with the buffer ranges we specify to `create_bind_group`. I think, with that in mind, (2) would work best for all the spots where we need the size.

grovesNL:
> I think, with that in mind, (2) would work best for all the spots where we need the size.

The special value of `0` for remainder could be a bit confusing if `0` is allowed in the WebGPU spec (`0` vs `undefined`).

Maybe we could have 2 accepting something like `Into<Option<u32>>` to make it easy to pass exact sizes, so `None` would only be needed for the remainder case? Or similarly we could have our own enum like `enum Size { Exact(u32), Remainder }` (optionally implementing `From<u32>` to make it easy to use exact sizes)?

kvark:
So it depends on whether the spec would allow 0 there. If not - then going with (2) seems ok.
If yes, getting our own `enum Size` looks to have most sense to me.

aloucks:
> treat size=0 as the remaining part of the buffer (current version of this PR)

Having `0` act as a magic number doesn't feel very good, imo.  `Into<Option<u32>>` or having `enum Size { Exact(u32), Remainder }` with convenient `From` implementations seems much nicer.

kvark:
@aloucks the problem with `Into<Xxx>` as a solution for ergonomics is that it requires the API to be generic. We are generally trying to have less generics in wgpu-rs surface. Also, while single-use entry points are fine, the stuff like `create_bind_group` takes arbitrary number of buffers, and they need individual ranges. Having a generic parameter there is not a solution.

That said, going for `Option<>` as a size could be a compromise: it doesn't require any new types, always in scope, and non-generic.

kvark:
Let's continue talking about ranges in https://github.com/gfx-rs/wgpu-rs/issues/199. This PR has a lot of goods that need to go regardless.
bors r=grovesNL

grovesNL:
bors retry

PR #200, opened 2020-03-15
Derive `Hash` and `PartialEq` for unique identifier wrapper types

mitchmindtree:
This is particularly useful downstream for distinguishing between
instances of these types.

I was unsure about `Device` as I noticed it has a `Temp` field that
looks like it might eventually store some non-PartialEq/Hash-friendly
fields.

Let me know if you'd like me to add a derive for `Device` or if there
are any issues with those I have updated.


kvark:
For `Device`, we'd need to implement the traits manually and exclude the `temp` thing.

PR #202, opened 2020-03-16
Remove `Hash` for ID types

grovesNL:
Reverts some of the `Hash` derives added in #200 because we don't have a good way to support `Hash` for JavaScript objects (for the web backend) at the moment. JavaScript objects don't have a unique identifier that we can use for `Hash`, so any solution here would probably mean adding our own map of JS object<->unique ID.


kvark:
bors retry

Issue #175, opened 2020-02-10
[meta] Async all the things


kvark:
All the asynchronous methods of the Web API should be async here as well:
  - [x] buffer mapping
  - [x] request adapter
  - [x] request device
  - [ ] get next swapchain texture

At the very least, we can return a future that is resolved instantly.

lachlansneff:
I did some digging and as far as I can tell, `SwapChain::get_next_texture` cannot be made an async function (aside from returning a future that instantly resolves). I think the function that ends up getting called that may block is [`gfx_hal::PresentationSurface::acquire_image`](https://docs.rs/gfx-hal/0.4.1/gfx_hal/window/trait.PresentationSurface.html#tymethod.acquire_image).

grovesNL:
Some of the async API changes are available in #193, but callbacks haven't been implemented for native though.

aloucks:
The call to [wgpu_device_poll](https://github.com/gfx-rs/wgpu-rs/blob/master/src/future.rs#L38)  in the `GpuFuture` also seems a little odd. I believe that it can block (or soft block). I don't think it's intended to poll resources directly, but rather attempt to resolve (again) after being woken up.

https://doc.rust-lang.org/beta/std/future/trait.Future.html#runtime-characteristics

> An implementation of poll should strive to return quickly, and should not block. Returning quickly prevents unnecessarily clogging up threads or event loops. If it is known ahead of time that a call to poll may end up taking awhile, the work should be offloaded to a thread pool (or something similar) to ensure that poll can return quickly.

In regards to `gfx_hal::PresentationSurface::acquire_image`, passing in `0` for the timeout should cause it to return immediately with `NotReady` if an image can't be acquired (at least in the case of vulkan). The `poll` result would be `Pending`.  

What I don't understand is how it's actually getting plumbed into the runtime. With networking, the socket is registered with epoll/kqueue/iocp. It's not clear (to me) how this would work with the GPU. I think there needs to be some sort of "IO driver" that's monitoring fences/semaphores/submissions/etc and waking up tasks when the resource it's waiting on has transitioned.

lachlansneff:
I agree with @aloucks. I traced `device_poll` all the way down and it looks like it blocks.

kvark:
Polling isn't supposed to block, by definition. What makes you think it does?

aloucks:
> Polling isn't supposed to block, by definition. What makes you think it does?

[...]

PR #204, opened 2020-03-16
Return futures from request adapter/request device

grovesNL:
Relevant to #175

Backporting some parts of #193 for async request adapter and async request device.


grovesNL:
bors retry

PR #208, opened 2020-03-19
Remove lifetime from SwapChainOutput

### LINKED ISSUE #207: Loosen lifetime requirements for SwapChainOutput.
As mentioned on matrix, the lifetime requirements for `SwapChainOutput` can be loosened to allow for more flexible usable (such as passing it between threads). I'm not sure how complicated this change is. If it just involves changing the stored reference to a value, I'd be happy to submit a pr.

lachlansneff:
The pr removes the lifetime from `SwapChainOutput`. It's not clear to me if the lifetime is necessary for soundness reasons, so let me know if so.

Resolves #207

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -397,9 +397,9 @@ pub struct RenderPassDescriptor<'a, 'b> {
 
 /// A swap chain image that can be rendered to.
 #[derive(Debug)]
-pub struct SwapChainOutput<'a> {
+pub struct SwapChainOutput {
     pub view: TextureView,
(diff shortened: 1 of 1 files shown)


kvark:
@grovesNL any concerns from your side?

lachlansneff:
> I think it's fine to land like this. The expectation here is that, if another swap chain image is requested while the first one is alive, the underlying `wgpu` would return an error, and we don't need to do anything specific here.

Perhaps we should note this in the documentation for `SwapChainOutput` or `SwapChain`?

kvark:
@lachlansneff would be great, yes!

PR #210, opened 2020-03-21
Update WGPU deps to rev 08e8d406c175579da5ef18c1abf4d6c00e2a9726

aloucks:
I also moved the `path` chunks in `Cargo.toml` to a single `[patch]` section so that it's one block to commment/uncomment for local development.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,26 +20,13 @@ use std::{
 
 pub use wgt::*;
 pub use wgc::{
-    Extent3d,
-    Origin3d,
-    command::{
-        CommandBufferDescriptor,
-    },
+    binding_model::TextureComponentType,
     device::{
         BIND_BUFFER_ALIGNMENT,
     },
     instance::{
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -20,27 +20,14 @@ use std::{
...
-    },
+    binding_model::TextureComponentType,
     device::{
         BIND_BUFFER_ALIGNMENT,
     },
     instance::{
         AdapterInfo,
         DeviceType,
     },
-    resource::{
```
kvark: how come we are no longer re-exporting these?

aloucks: They're included above in `pub use wgt::*;`

kvark:
Bors retry

kvark:
Bors r+

kvark:
bors retry

aloucks:
retry agian?

kvark:
bors retry

PR #211, opened 2020-03-21
Draft of BufferRange support

### LINKED ISSUE #199: Buffer ranges
Every place where a buffer is used, we need to provide a range. Currently, `crate_bind_group` requires the actual size to be passed, which is often inconvenient. The upstream spec allows to assume the whole size available, if nothing is provided. Also `set_index_buffer` and `set_vertex_buffer` now need the size (also optional).

There are many options here:
  1. require both the offset and the size be meaningful
  2. treat size=0 as the remaining part of the buffer. This may be a bit confusing if 0 is computed automatically and unintentionally.
  3. turn size into an `Option<BufferAddress>`. This is what gfx-rs does. It has slightly weak semantics, but hands down convenient.
  3a. introduce `enum Size { Exact(u32), Remainder }` - this is stronger but less convenient
  4. be generic over `RangeBounds`. This doesn't work well with `create_bind_group` since it has many buffers in it.

In addition, we should have a struct that encapsulates buffer + range to be used in all these entry points, e.g.
```rust
pub struct BufferRange<'a> {
  buffer: &'a Buffer,
  offset: BufferAddress,
  size: Option<BufferAddress>,
}
```

Somewhat relevant, `Option<BufferAddress>` takes 16 bytes, which is sad.

lachlansneff:
This pr contains quite a lot of stuff! Essentially the root of it is that I've added a new `BufferRange<Bounds>` type and a lot of code has been refactored to take advantage of it.

Here's what you can do:

```rust
// Copy from a bounded buffer range.
encoder.copy_buffer_to_buffer(source.range(0, 64), &destination);

// Copy from an unbounded buffer range.
encoder.copy_buffer_to_buffer(source.range(0, ToEnd), &destination);

let slice = some_buffer.range(0, 128).map_read().await?; // Or `map_write`.
// `some_buffer.range(0, ToEnd).map_read/map_write()` is statically disallowed!

// If you're not sure whether a buffer is going to be bounded or unbounded,
// you can use the `Unsure` bounds.
let br0: BufferRange<Unsure> = some_buffer.range(0, ToEnd).into();
let br1: BufferRange<Unsure> = some_buffer.range(0, 64).into();

// You can also convert back in a failable manner.
let br2: BufferRange<Unbounded> = br0.try_into()?;
let br3: BufferRange<Bounded> = br1.try_into()?;
```

The documentation isn't quite there yet for this, and there are probably edge-cases to work out, but I think it's quite clever :P.

----

`BufferRange` is interesting because it uses generics to represent both statically-known bounds and runtime-known bounds.

As such, the type parameter to `BufferRange` can be one of three types: `Unbounded`, `Bounded`, and `Unsure`, which all implement the `SizedBuffer` trait:

```rust
pub trait SizedBuffer: private::Sealed {
    type Storage: SizeStorage;
}
```

The `Storage` associated type of `SizedBuffer` must implement `SizeStorage`, which is implemented for `ToEnd` (a unit struct), `BufferAddress`, and `Option<BufferAddress>`:

```rust
pub trait SizeStorage: Copy + Clone + PartialEq + std::fmt::Debug + private::Sealed {
    fn to_option(self) -> Option<BufferAddress>;
}
```

A new `range(start, size)` method is added to `Buffer` and `BufferRange` through the `RangedBuffer` trait:

```rust
pub trait RangedBuffer<'a, End: SizeStorage, Bounds: SizedBuffer>: private::Sealed {
    /// Index a `Buffer` or `BufferRange` by a range.
    fn range(self, offset: BufferAddress, size: End) -> BufferRange<'a, Bounds>;
}
```

There are also two more traits, used to constrain types to be statically known bounds:

```rust
pub trait StaticSizeStorage: SizeStorage {} // Implemented for `ToEnd` and `BufferAddress`.
pub trait StaticSizedBuffer: SizedBuffer {} // Implemented automatically for the bounds that have `Storage`: `StaticSizeStorage`.
```

Resolves #199

### DIFF
--- a/examples/boids/main.rs
+++ b/examples/boids/main.rs
@@ -9,7 +9,11 @@ mod framework;
 use std::fmt::Write;
 use zerocopy::{AsBytes};
 
-use wgpu::vertex_attr_array;
+use wgpu::{
+    prelude::*,
+    ToEnd,
+    vertex_attr_array,
+};
 
 
 // number of boid particles to simulate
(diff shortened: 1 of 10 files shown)


lachlansneff:
@kvark Thanks for taking a look! I'll close the pr then :)

PR #212, opened 2020-03-22
Improve BufferCopyView documentation

rukai:
groves suggested that the alignment/multiple requirements are not implemented yet.
So I changed the documentation to reflect that.
Ideally we would change the examples to demonstrate how to work around this requirement, but I think its best to wait until we know for certain that this IS a requirement, currently the WebGPU spec does not have this as a requirement.

Additionally it seems like every example except for the skybox example failed to follow this part of the WebGPU spec.
"Note: rowsPerImage must be zero for copies with a copySize.depth of 1, and must be greater than zero otherwise."
So I fixed those examples and documented the requirement on the field.

### DIFF
--- a/examples/capture/main.rs
+++ b/examples/capture/main.rs
@@ -77,7 +77,7 @@ async fn run() {
                 buffer: &output_buffer,
                 offset: 0,
                 bytes_per_row: size_of::<u32>() as u32 * size,
-                rows_per_image: size,
+                rows_per_image: 0,
             },
             texture_extent,
         );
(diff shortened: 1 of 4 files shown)


kvark:
Bors retry

aloucks:
> Note: rowsPerImage must be zero for copies with a copySize.depth of 1, and must be greater than zero otherwise.

@kvark Is "_must_ be zero"  required here? I would think it would be "_may_ be zero".

https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkBufferImageCopy.html

> `bufferRowLength` and `bufferImageHeight` specify in texels a subregion of a larger two- or three-dimensional image in buffer memory, and control the addressing calculations. **If either of these values is zero, that aspect of the buffer memory is considered to be tightly packed according to the imageExtent**.

I would assume that the other APIs have similar requirements or the zero value wouldn't be an option. Having zero as a convenience seems fine, but allowing the explicit `rowsPerImage` seems reasonable too.

PR #213, opened 2020-03-23
Cleanup user-interface around async buffer mapping

lachlansneff:
This PR cleans up the interface for asynchronous buffer mapping a little. It turns `map_read` and `map_write` into `async` functions (instead of returning `impl Future<...>`). It also adds a standin error type and removes some confusing type aliases.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,7 +11,6 @@ use smallvec::SmallVec;
 
 use std::{
     ffi::CString,
-    future::Future,
     ops::Range,
     ptr,
     slice,
@@ -845,13 +844,14 @@ impl Drop for Device {
     }
 }
 
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct BufferAsyncErr;
+
 pub struct BufferReadMapping {
     data: *const u8,
     size: usize,
     buffer_id: wgc::id::BufferId,
 }
-//TODO: proper error type
-pub type BufferMapReadResult = Result<BufferReadMapping, ()>;
 
 impl BufferReadMapping
 {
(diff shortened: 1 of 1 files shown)


kvark:
@grovesNL could you have a look as well?

src/lib.rs
```diff
@@ -891,36 +889,23 @@ impl Drop for BufferWriteMapping {
...
 }
 
-pub struct BufferAsyncMapping<T> {
-    pub data: T,
-    buffer_id: wgc::id::BufferId,
-}
-//TODO: proper error type
-pub type BufferMapAsyncResult<T> = Result<BufferAsyncMapping<T>, ()>;
-
-impl<T> Drop for BufferAsyncMapping<T> {
```
grovesNL: Where do we do the buffer unmapping now if it doesn't happen during `Drop`?

src/lib.rs
```diff
@@ -996,7 +981,7 @@ impl Buffer {
             Box::into_raw(user_data) as *mut u8,
         );
 
-        future
+        future.await
```
grovesNL: Doesn't this wrap `GpuFuture` into another future? It seems better to avoid that if we can (this is why we were returning`GpuFuture` directly here)

src/lib.rs
lachlansneff: I believe that `BufferAsyncMapping` was left over from when `map_read` and `map_write` used callbacks, and is now completely unused. So this `Drop` impl is also unused.

src/lib.rs
lachlansneff: While this is technically true, it won't add any measurable impact here. And in a few weeks, futures are getting a few optimizations through generator resume arguments, so it's possible that this could get entirely optimized away.

src/lib.rs
grovesNL: Ok sounds good, thanks!

Issue #156, opened 2020-01-08
Case for using wgpu-core


kvark:
Currently, `wgpu-rs` uses the C API of `wgpu-native`, which then uses `wgpu-core` for the gory details. This issue raises the question of whether the middle step can be skipped. In order to answer the question, we need to consider all the possible targets for `wgpu-rs`:
  1. Native execution with `wgpu`. This is the only target supported *right now*. It could easily bypass `wgpu-native` and call into `wgpu-core`.
  2. Native execution with Dawn (see https://github.com/gfx-rs/wgpu/issues/47). We'd have to use the API calls according to [webgpu-headers](https://github.com/webgpu-native/webgpu-headers). This would technically work if `wgpu-native` implements the headers and becomes a drop-in replacement for Dawn. The target is incompatible with using `wgpu-core` directly. cc @Kangz
  3. Browser execution on the Web (see #101). The plan is to use `web-sys` directly and thus it doesn't depend on `wgpu`. cc @grovesNL 

With that in mind, the only blocker for using `wgpu-core` directly is running on Dawn. This is somewhat a minor goal for us, given that we are trying to make `wgpu` a batter (or comparable) solution.

My proposal would be to turn `wgpu-rs` into a multi-backend library (like `gfx-rs` is), enforcing the API with traits. The backends are:
  1. `wgpu-core` - for fastest native development, no FFI boundary.
  2. `webgpu-headers` - for generic standard-compatible target with `wgpu-native`, `dawn`, and possibly some `emscripten` drop-in libraries implementing the headers.
  3. `web-sys` - for web development

grovesNL:
I think I slightly prefer to go through wgpu-native (keeping the middle step), even if only to ensure that the C API works well through dog-fooding. We could also decide to skip wgpu-native later if it turns out to be problematic. We do pay slightly for the indirection through the FFI layer but we could try to measure the cost of this to understand whether it's actually meaningful in typical use cases.

In contrast, with gfx-hal we were focused on as little indirection as possible, which is why we didn't end up implementing the traits from ash or implement the Vulkan API directly. But maybe some amount of indirection isn't too bad in wgpu, especially because of how a lot of the API has been designed.

Kangz:
[...]

PR #216, opened 2020-03-24
Add PartialEq and Eq for more types

aloucks:
Depends on https://github.com/gfx-rs/wgpu/pull/535

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -74,18 +74,11 @@ pub use wgc::instance::{
     DeviceType,
 };
 
-//TODO: avoid heap allocating vectors during resource creation.
-#[derive(Default, Debug)]
-struct Temp {
-    //bind_group_descriptors: Vec<wgn::BindGroupDescriptor>,
-    //vertex_buffers: Vec<wgn::VertexBufferDescriptor>,
-}
-
 /// A handle to a physical graphics and/or compute device.
 ///
 /// An `Adapter` can be used to open a connection to the corresponding device on the host system,
 /// yielding a [`Device`] object.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq)]
 pub struct Adapter {
     id: wgc::id::AdapterId,
 }
@@ -103,10 +96,9 @@ pub struct RequestAdapterOptions<'a> {
 ///
 /// The `Device` is the responsible for the creation of most rendering and compute resources, as
 /// well as exposing [`Queue`] objects.
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq)]
 pub struct Device {
     id: wgc::id::DeviceId,
-    temp: Temp,
 }
 
 /// This is passed to `Device::poll` to control whether
@@ -118,14 +110,14 @@ pub enum Maintain {
 }
 
 /// A handle to a GPU-accessible buffer.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq)]
 pub struct Buffer {
     id: wgc::id::BufferId,
     device_id: wgc::id::DeviceId,
 }
 
 /// A handle to a texture on the GPU.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq)]
 pub struct Texture {
     id: wgc::id::TextureId,
     owned: bool,
@@ -135,7 +127,7 @@ pub struct Texture {
 ///
 /// A `TextureView` object describes a texture and associated metadata needed by a
 /// [`RenderPipeline`] or [`BindGroup`].
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq)]
 pub struct TextureView {
     id: wgc::id::TextureViewId,
     owned: bool,
@@ -146,7 +138,7 @@ pub struct TextureView {
 /// A `Sampler` object defines how a pipeline will sample from a [`TextureView`]. Samplers define
 /// image filters (including anisotropy) and address (wrapping) modes, among other things. See
 /// the documentation for [`SamplerDescriptor`] for more information.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq)]
 pub struct Sampler {
     id: wgc::id::SamplerId,
 }
@@ -155,7 +147,7 @@ pub struct Sampler {
 ///
 /// A `Surface` represents a platform-specific surface (e.g. a window) to which rendered images may
 /// be presented. A `Surface` may be created with [`Surface::create`].
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq)]
 pub struct Surface {
     id: wgc::id::SurfaceId,
 }
@@ -164,7 +156,7 @@ pub struct Surface {
 ///
 /// A `SwapChain` represents the image or series of images that will be presented to a [`Surface`].
 /// A `SwapChain` may be created with [`Device::create_swap_chain`].
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq)]
 pub struct SwapChain {
     id: wgc::id::SwapChainId,
 }
@@ -175,7 +167,7 @@ pub struct SwapChain {
 /// create a [`BindGroupDescriptor`] object, which in turn can be used to create a [`BindGroup`]
 /// object with [`Device::create_bind_group`]. A series of `BindGroupLayout`s can also be used to
 /// create a [`PipelineLayoutDescriptor`], which can be used to create a [`PipelineLayout`].
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq)]
 pub struct BindGroupLayout {
     id: wgc::id::BindGroupLayoutId,
 }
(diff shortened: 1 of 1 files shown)


grovesNL:
I think this will have the same issue as mentioned in https://github.com/gfx-rs/wgpu-rs/pull/202

aloucks:
@grovesNL I'm not sure I understand how adding Hash implementations to these handle types causes an issue with the web backend. I can understand that unique IDs for javascript objects could be a problem, but I don't see how they are related. Wouldn't the javascript object reference the handle, not the other way around?

aloucks:
I'd like to get a better understanding as to why adding `Hash` causes a problem. Admittingly, I know very little about wasm/bindgen interop with Rust, but many of the std structs and primitives implement `Hash` - I don't understand why having it here is different.

In any case, I think we could ultimately get around it with something like:

```rust
#[cfg_attr(not(target_arch = "wasm32"), derive(Hash))]
```

grovesNL:
Types like `Adapter` on the web would contain a reference to the JavaScript object (instead of an ID). For example, on the web backend we'd redefine `AdapterId` like https://github.com/gfx-rs/wgpu-rs/blob/1084bbdc430de5501fea75a9a5867b1d1642d075/src/backend/web.rs#L6 to reference [`GpuAdapter`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.GpuAdapter.html), but `GpuAdapter` isn't able to derive `Hash` because the `GpuAdapter` JS object is an opaque handle. We could add a layer of indirection to map ID<->JS object for the web backend but it would be nice to avoid this if possible.

aloucks:
@grovesNL Thank you, now I understand how it's an issue. I've pushed a new commit that will conditionally implement `Hash` for only the non-wasm32 targets

grovesNL:
Ok great, thanks! I'm still not sure whether it's best to hide it behind `target_arch`, hide it behind a feature, disallow it on both native+web, add extra indirection to the web backend (ID<->JS object map) to expose it on both native+web, or something else.

For example, it would be a bit unfortunate for portability if crates tend to rely on this, then later realize they can't easily target the web just by switching targets.

aloucks:
You could give every object a unique ID using [`AtomicU64::fetch_add`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU64.html#method.fetch_add) and implement `Hash` by hand.

aloucks:
Another option is to just use a non-atomic incrementing `u64` if it's strictly used for hash only (and not identity). An occasional dirty read won't be the end of the world.

Would you be comfortable to merge if I commented out the `Hash` derives to be sorted out later. This way we can at least get `Eq` and `PartialEq` implemented.

grovesNL:
Yeah that sounds good to me :+1: I think these are interesting ideas and worth exploring, but it might be safer to remove the `Hash` derives for now and figure this out later. It should still be possible for users to manually implement their own `Hash` on newtypes for now if it's really useful to them.

aloucks:
> It should still be possible for users to manually implement their own Hash on newtypes for now if it's really useful to them.

Are the `id` fields exposed? I don't think there's any value for a user to hash, which is ironically the same problem that you have in the web space.

grovesNL:
Right, because `id` isn't exposed it would basically mean using a solution like you mentioned above, or creating their own maps of handle<->object if they need handles anyway, etc.

aloucks:
@grovesNL I've removed the Hash derives.

grovesNL:
Thanks! It looks like CI has some build errors

aloucks:
> Thanks! It looks like CI has some build errors

Yeah, it depends the changes here: https://github.com/gfx-rs/wgpu/pull/535

grovesNL:
Oh whoops, sorry – I thought we already updated to a newer `rev`

kvark:
please rebase

kvark:
@aloucks would you want to snuck this in for 0.5 or follow-up?

aloucks:
@kvark Yes. I'll rebase and push this evening.

aloucks:
@kvark should be good now.

src/lib.rs
```diff
@@ -81,11 +81,25 @@ struct Temp {
     //vertex_buffers: Vec<wgn::VertexBufferDescriptor>,
 }
 
+impl PartialEq for Temp {
```
kvark, line 84: please remove

aloucks:
@kvark - That Temp struct still exists. If I remove the PartialEq impl it will break the PartialEq impl for Device. Do you want me to remove Temp and any references to it?

aloucks:
@kvark I've removed `Temp` and all references to it. As described above, I can't remove the `PartialEq` impl without removing the struct or it breaks `PartialEq` for device. The commits are squashed now too.

kvark:
@aloucks wow, `Temp` wasn't even introduced by your PR.
Sorry about bringing this up! I think I'm having an attention deficit, and this was my mistake.

Main question is: where is it useful to compare any ObjectId for equality?
I'm thinking, since you can't clone them, you have to have only one object anyway, so you'd only be comparing a reference to a reference, in which case you might as well do a pointer eq.

aloucks:
You can't clone them but you can have multiple references to them. The original PR included `Hash` implementations so you could make them keys in a map. I'm hoping to get that back eventually. baby steps..

kvark:
> You can't clone them but you can have multiple references to them.

Yes, thinking about this more, I feel stronger that these derives are not necessary. PartialEq only makes sense for things that you can have multiple instances of. If you have unique instances, and just different references, then you should be comparing the references.
```rust
struct RefKey<'a, T: 'a>(&'a T);
impl PartialEq for RefKey<'_, _> {
  fn eq(&self, other: &Self) -> bool {
    std::ptr::eq(self.0, other.0)
  }
}
```
Similarly can be done for `Arc`/`Rc`. In any case, this isn't needed in `wgpu-rs` itself.

aloucks:
Not having `PartialEq` and `Eq` also prevents comparing `Arc<Buffer>` with `Arc<Buffer>`.

aloucks:
>  I feel stronger that these derives are not necessary. PartialEq only makes sense for things that you can have multiple instances of. 

I strongly disagree with this. It's makes comparison with `Arc` extremely uneconomic for no real benefit. The lack of `Clone` does not mean it shouldn't implement `Eq`.

kvark:
> Not having PartialEq and Eq also prohibits comparing Arc<Buffer> with Arc<Buffer>.

Yes, and is it any different from `&Buffer`? In both cases, you can just wrap an object into a unit struct like `ArcKey` that has a generic implementation of `PartialEq`, and you are good to go. This is just 5 lines of code.

> Missing PartialEq and Eq also makes it impossible to drive those on structs that contain these handles.

If your struct contains `Buffer` by value, it means that structs is also unique, so you'd just implement PartialEq for the references of it (or pointers to it), not for the struct itself (or just use the `RefKey` one I posted). If you think about it, comparing by raw pointer/reference value is faster anyway, so in the end you'd want to have that regardless of whether or not we have `PartialEq`.

aloucks:
> In both cases, you can just wrap an object into a unit struct like ArcKey that has a generic implementation of PartialEq, and you are good to go. This is just 5 lines of code.

But how and why is this of any value? Now I need to add to also create a `Deref` impl for `ArcKey`, when `Arc` already has all this.  There is no downside to implementing `Eq`.

kvark:
Our object wrappers currently are mostly just carrying an ID, but this isn't necessarily the case. For example, `Device` and `Texture` have other data than the ID. When we get #156, we may see more difference: things like passes would now contain raw encoded data.

So in the end we are likely going to have 3 different backends, or implementations of these types (wgpu-core, wgpu-native, and web-sys). Putting any derive on them means more friction and less freedom (i.e. it will be less straightforward to extend the structs) to these implementations. The amount of code that would go into wgpu-rs itself (manual implementation of the traits for 5+ types, potentially growing) would then exceed the amount of code that an application would need to have (i.e. `RefKey` used for everything that it needs). Moreover, by taking the wrapper path, the implementation has more control over what it needs to compare, and the comparisons are faster (since there is no need to look at the contents of the structs at all).

So in the end, it seems to me that this task is not obviously better solved by wgpu-rs in practice. And in theory, `PartialEq` is not what you want to do for comparing references/smart pointers to types. Let's see if your library or application has any trouble with handling it on your side, let's hear if @mitchmindtree or @kyren share the same complaint, and adapt.

kyren:
Just my 2c, since you pinged me:

We currently have no need for Eq or PartialEq implementations for any of the opaque, unique handle types like `Buffer`.

I'd rather the API be consistent, and I definitely think `wgpu-rs` shouldn't pick a nonzero-cost strategy to *make* a type Eq or PartialEq, since the user should really make that choice.

If making a unique (non-Clone) handle type Eq or PartialEq is a compatibility or maintenance hazard then that seems like a good idea not to have those impls.

Now, if there is *definitely* no compatibility hazard or maintenance burden then I suppose it's not harmful to have those impls, but I can't foresee using them for anything (again, just talking about the unique handle-to-resources types like `Device`, `Buffer`, etc).

aloucks:
The `web_sys` objects also implement `PartialEq` and `Eq` for these types:

Example:
https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.GpuBuffer.html

kvark:
True. It's easy to see why this is the case though: JS objects on the web in general are refcounted, so web-sys has the semantics of `Arc` built-in. But we do not currently have that. There were talks about doing this though, in particular because of interest from Google in exposing this in Dawn portably (https://github.com/webgpu-native/webgpu-headers/pull/15).

kvark:
I think we reached the conclusion that so far the PartialEq/Eq are not needed. Please feel free to raise this topic again for a follow-up!

PR #219, opened 2020-03-26
Update Zerocopy to 0.3

cwfitzgerald:
In an effort to eliminate duplicate dependencies, especially on the proc-macro/syn stack, this updates zerocopy 0.2 (using syn 0.6) to zerocopy 0.3 (using syn 1). There appears to be no real api changes based on my skimming of the docs. Things compile on my windows machine, and hopefully it will pass on mac and linux.


cwfitzgerald:
Sounds great, the lower the compile time the better :)

PR #222, opened 2020-03-29
Update to latest wgpu

grovesNL:
- Avoid array indexing in hello-triangle
  - We need a fix from https://github.com/KhronosGroup/SPIRV-Cross/pull/1284 in spirv_cross to allow this to work again on certain macOS version+Intel driver version combinations, see https://github.com/KhronosGroup/SPIRV-Cross/issues/1210#issuecomment-582677237
- Update to latest wgpu commit
  - We re-export descriptors with `Option<String>` instead of `*c_char` for debug labels
  - All re-exports from `wgt` are explicitly named now so we can track them slightly better


src/lib.rs
```diff
@@ -821,8 +937,12 @@ impl Device {
 
     /// Creates a new buffer, maps it into host-visible memory, copies data from the given slice,
     /// and finally unmaps it, returning a [`Buffer`].
-    pub fn create_buffer_with_data(&self, data: &[u8], usage: BufferUsage) -> Buffer {
-        let mapped = self.create_buffer_mapped(data.len(), usage);
+    pub fn create_buffer_with_data(&self, data: &[u8], usage: BufferUsage, label: Option<String>) -> Buffer {
```
grovesNL: I included `label` as an argument here instead of leaving it blank, but it does make it a bit less convenient to use

src/lib.rs
```diff
@@ -1497,3 +1627,18 @@ impl SwapChain {
         }
     }
 }
+
+struct OwnedLabel(Option<CString>);
```
grovesNL: Note that I haven't actually tested this yet but I think the pointer lifetime is ok. Also interested in ways to improve this

src/lib.rs
```diff
@@ -276,9 +321,14 @@ pub struct BindGroupLayoutEntry {
...
 }
 
+/// A description of a bind group layout.
 #[derive(Clone, Debug)]
 pub struct BindGroupLayoutDescriptor<'a> {
     pub bindings: &'a [BindGroupLayoutEntry],
+
+    /// An optional label to apply to the bind group layout.
+    /// This can be useful for debugging and performance analysis.
+    pub label: Option<String>,
```
aloucks: Why are the labels using owned `String` instead of `&str`? The descriptors already have lifetimes. If we're going to make these owned, we might as well consider relaxing the others (e.g. `&[T]` to `Vec<T>`).

src/lib.rs
aloucks: Should we have `label` on the convenience methods? The descriptor could just be be used instead.

src/lib.rs
aloucks: I have a fully tested inlinable label mechanism here that we didn't end up using in wgpu-core.

https://github.com/aloucks/wgpu/commit/ca02eba2a7d9106b3e0ccaadb467072f32e025a2

src/lib.rs
aloucks: I also think `create_buffer_mapped` will prefer CPU visible memory instead of device local - which should primarily be used for staging buffers. Not having the label is sort of a hint that this method isn't for your read-only long-lived data.

grovesNL: Yeah we could omit it here but labels can also still be useful for staging buffers

src/lib.rs
grovesNL: Some of these descriptors don't have lifetimes currently (`BufferDescriptor`, `CommandEncoderDescriptor`, `TextureDescriptor`).

We could make some `Option<String>` and others `Option<&str>` (a slight downside is that they'd be inconsistent), or we could make all `Option<&str>`  (potentially large downside that `BufferDescriptor` etc. can't be stored as easily)

src/lib.rs
kvark: I think it's fine to have the argument on `create_buffer_mapped` but let's not put it on convenience methods (note to @aloucks  - in case of `create_buffer_with_data` one can't use a descriptor as a workaround, they need to call `create_buffer_mapped`).

In the nearest future hopefully the upstream API around mapping will be better, i.e. you'd just pass `mapped: true` in the descriptor, and then you'd obtain the mapping pointer through a different method.

> I also think create_buffer_mapped will prefer CPU visible memory instead of device local - which should primarily be used for staging buffers.

Filed https://github.com/gfx-rs/wgpu/issues/542 to make this better.

src/lib.rs
kvark: I'd say let's consistently have `Option<&str>` everywhere

src/lib.rs
kvark: That's a useful mechanism indeed. I think it's fine to own `OwnedLabel` here for it being simple enough, and then follow-up.

src/lib.rs
grovesNL: Ok :+1: will do

kvark:
There is a problem with linking on CI - https://travis-ci.org/github/gfx-rs/wgpu-rs/jobs/668551855#L326

grovesNL:
It's the same issue we workaround in gfx https://github.com/gfx-rs/gfx/blob/master/.travis.yml#L62 when using gcc.

We can either default to msvc on Windows (probably more common) or use the workaround from gfx (the workaround is used there because we still want to be able to run the gfx Makefile).

kvark:
@grovesNL thank you!
CI is green now. Next time please just proceed on your own, no need to wait for me :)

PR #223, opened 2020-03-29
Move multithreaded_compute.rs into hello-compute tests

rukai:
Also setup hello-compute tests to run during `cargo test`, as per the solution posted in https://github.com/rust-lang/cargo/issues/6831


rukai:
The old test was broken due to not being run on ci.
So I copy pasted the compute example into the old test.
But then I decided that was silly let's avoid this code duplication by moving it in with the other tests.

aloucks:
CI doesn't run the tests, presumably because travis does not have GPU-enabled runners.

https://github.com/gfx-rs/wgpu-rs/blob/d91b78bdfba1dfd991ddc0fa605b88336dfaf688/.travis.yml#L14

kvark:
Bors retry

PR #234, opened 2020-04-03
Implement Send and Sync for BufferReadMapping and BufferWriteMapping

Healthire:


aloucks:
`Sync` is not safe for `BufferWriteMapping`

https://github.com/gfx-rs/wgpu-rs/blob/e198d638fdebe0943fc64ecea95841e8bc20d7d4/src/lib.rs#L1055-L1062

Making it sync enables two mutable references to the same data.

That method should be called `as_mut_slice` also to follow rust naming conventions.

kvark:
No, it is still safe. Sync is true when it's safe to hold multiple `&T` and access them. You can't call this method by `&T` (since it's `&mut self`), so this method doesn't prevent `Sync`.

PR #245, opened 2020-04-07
Check `cargo doc` in CI

hecrj:
This PR makes CI run `cargo doc` using the latest nightly.

It is currently panicking here:

https://github.com/rust-lang/rust/blob/6dee5f1126dfd5c9314ee5ae9d9eb010e35ef257/src/librustdoc/passes/mod.rs#L411

I thought it was an incorrect link causing it, but the issue seems to persist even after fixing the link.

__Update__: Issue opened upstream by @grovesNL (https://github.com/rust-lang/rust/issues/70874).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,15 @@ os:
   - osx
   - windows
 
+jobs:
+  include:
+    - os: linux
+      rust: nightly
+      script: cargo doc --lib --no-deps
+  allow_failures:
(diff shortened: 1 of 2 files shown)


kvark:
Thank you!
if this fails for now, we should have allow-fail on it
bors r+

hecrj:
I have added the new job to the `allow-failures` section and rebased.

PR #246, opened 2020-04-07
Replace current gpu future implementation with one that doesn't allocate.

lachlansneff:
Through the magic of Pin, it's possible to remove every allocation from the `GpuFuture` implementation. Since a future is guaranteed to be pinned to a location after it's been polled, we can pass pointers to "stack" allocated values safely! There is one caveat: that we have to be careful about canceling a mapping operation if we drop the future before it resolves because the backend would then have a dangling pointer to a value that no longer exists. Thankfully, this is easily solved.

The contract with the user changes very little, with the only notable difference being that the buffer mapping will only be initiated once the future is first polled. This is necessary for soundness. The previous behavior here was only useful for examples, and would never be used in an actual application, so I don't count this as a regression.


src/backend/mod.rs
```diff
@@ -1 +1 @@
-pub(crate) mod native_gpu_future;
+pub(crate) mod native_gpu_future;
```
kvark, line 1: missing newline?

src/lib.rs
```diff
@@ -1067,9 +1067,29 @@ impl Drop for BufferWriteMapping {
     }
 }
 
+impl Buffer {
+    extern "C" fn buffer_map_read_callback(
```
kvark, line 1071: should be `unsafe` function

lachlansneff:
@kvark Thanks for reviewing! If the internal costs of mapping a buffer are significantly more than the cost of an allocation, then the additional complexity here might not be worth it. I'll try to benchmark when I have time.

kvark:
It seems to me that the biggest unresolved question is whether or not WebGPU will have `writeToBuffer` way of uploading data. If yes, we don't need to care too much about the mapping allocation, since it will be rarely used. If no, we'll need to make it fast.

aloucks:
> Mapping in wgpu is kind of a heavy thing to do: it waits for GPU, it may potentially allocate things internally, it's not as light as native mapping.

@kvark  I thought the idea of async mapping was to prevent it from being a heavy-weight operation that stalls the GPU. I guess my understanding of how async mapping was supposed to work is as follows.

1. Request to map
2. If not mapped and not in by GPU use resolve immediately
3. If not mapped but in use by GPU, en-queue to be mapped
4. If mapped .. (error or en-queue?)

I don't think it would prevent you from having to triple buffer frequently updated data, but it does prevent you from having to think about it too much. If the buffer(s) you're mapping are never in use (because you've triple buffered), I would think the mapping request would resolve immediately and be effectively free.

kvark:
Your description is correct. How do you know, however, the number of buffers to queue? In general the approach would be to request mapping of a buffer once you use it, and then re-add it to your "free pool" once it's mapped. It's the only way to guarantee that the mapping is ready today. So going the idiomatic way, you'd still be relying on the "heavy" path of mapping.

src/backend/mod.rs
lachlansneff: Oops, yes.

src/lib.rs
lachlansneff: Yep, good catch!

aloucks:
> How do you know, however, the number of buffers to queue?

Are you asking about the user perspective? Wouldn't they use a ring buffer of `wgpu::Buffer`s with the count >= to the number of swapchain images for data that's updated very frequently?

For example, suppose you have some entities that are constantly moving and need their model transforms updated on every frame. Ideally the async write mapping would always resolve immediately:

```rust
let ring_buffer: &mut RingBuffer<wgpu::Buffer> = &mut self.buffers;
let next_buffer = ring_buffer.next();
let mapped_buffer = next_buffer.map_async_write().await;
self.write_model_matrices(mapped_buffer.data);
```

And for items that are infrequently updated, you don't care as much:

```rust
let mapped_buffer_future = buffer.map_async_write();

...

tokio::select! {
  mapped_buffer = &mut mapped_buffer_future => {
    // our buffer mapping resolved, let's do something with it
    mapped_buffer.data.copy_from_slice(...);
  frame = swap_chain.get_next_texture() => {
    // otherwise, render the frame as usual
  }
}

```

kvark:
Yes, except that you don't control how many images are in the swapchain: https://docs.rs/wgpu-types/0.5.0/wgpu_types/struct.SwapChainDescriptor.html , at least today.
So you could pessimistically assume a higher bound, or do it the idiomatic way.

aloucks:
Can you show an example of what you're describing as the idiomatic way? The "free pool" is essentially the ring buffer in my example. 

> Yes, except that you don't control how many images are in the swapchain

Is this a gap in the spec? I can understand not allowing the user to _control_ the amount, but maybe they should be able to _query_ the count?

kvark:
Here is a piece of WebGPU code that roughly implements it idiomatically - https://hackmd.io/@jgilbert/StockBeltUploader

> Is this a gap in the spec? I can understand not allowing the user to control the amount, but maybe they should be able to query the count?

It's not a gap in the upstream spec, it's an implementation detail. When you are a multi-process browser, telling the swapchain image count is still not any guarantee at any time that something is not used. There is an extra latency for IPC communication involved.

Whether or not to expose this for native-only in webgpu-headers can be discussed.

aloucks:
@kvark Thanks for digging up the example! 

I think the major difference from what I was describing is that it's growing it's list of buffers on the fly and moving them in and out of the `mappedBufferList`. Then it's remapping them immediately after submission. I think that in this case, yes, they will follow the "heavy" path. 

However, I think the example could be re-organized so that the re-mapping is deferred until it loops back around and tries to use the buffer in question. It would then need to check if the promise has been resolved or not. If not, it creates a new buffer. Once the buffer list length naturally exceeds the number of swapchain images, the promises should be resolved by the time it loops around then the "heavy" path gets phased out.

kvark:
> re-mapping is deferred until it loops back around and tries to use the buffer in question

In browsers, requesting for mapping isn't necessarily instantaneous even if a resource is not currently used by the GPU, since the content process is separated from the GPU. So if the example was rewritten to only request a map when needed, it would simply not work as you want it to be.

For this to function, there need to be changes in how WebGPU is specified, along the lines of @jdashg proposals, e.g. https://github.com/gpuweb/gpuweb/pull/506

kvark:
Maybe I misunderstood what you are describing. I do agree that potentially the code could be refactored in a way that expects the buffers *mostly* be available when asked for, in which case it's not the heavy path any more.

kvark:
@aloucks Let me ask it in a different way: do you understand the code of this PR? Would you consider sharing the burden of maintaining this code with @lachlansneff if it's merged?

aloucks:
@kvark Sure.

I don't think we should include the `map_[read|write]_blocking` functions though. If you want to block on the future you could use `futures::block_on`. 

This does, however, press the issue with the fact that there is no GPU reactor. Tokio and async-std provide an executor _and_ a reactor for socket based IO, but there is no **reactor** for GPU futures. wgpu will have to provide one. I'm not sure that this can be done in wgpu-rs alone though without support from wgpu-core/native. There needs to be an event notification system of some kind. 

There's a similarly related issue in the dawn tracker:
https://bugs.chromium.org/p/dawn/issues/detail?id=119&q=&can=2

A naive reactor could simply be calling `Device:poll` (is this going to be renamed?) in a loop. I think @lachlansneff had something similar in a prior PR.

lachlansneff:
There is something special about `map_read_blocking` in that it makes sure that the device polls at the right time so you don't need `device.poll` to be called elsewhere.

aloucks:
> There is something special about `map_read_blocking` in that it makes sure that the device polls at the right time so you don't need `device.poll` to be called elsewhere.

I don't think that's a good idea, but I do think that the rest of the PR is a step in the right direction.

lachlansneff:
I'll remove it, but it does make the examples more complicated until we get a real gpu future reactor.

aloucks:
Maybe wait for more feedback from @kvark . I don't want to cause you to do extra work or churn the code more.

kvark:
@lachlansneff don't listen to the bot please, it's experimental :)

lachlansneff:
@kvark these seem like fine suggestions haha.

lachlansneff:
Closing this until the buffer write/read story is filled out more.

kvark:
Thanks! I'm looking forward to the revival of this.

PR #248, opened 2020-04-07
Added sync feature for blocking operations

zesterer:


src/lib.rs
```diff
@@ -1118,8 +1139,14 @@ impl Buffer {
         future
     }
 
+    /// Synchronously map the buffer for reading.
+    #[cfg(feature = "sync")]
+    pub fn map_read_sync(&self, start: BufferAddress, size: BufferAddress) -> Result<BufferReadMapping, BufferAsyncErr> {
+        pollster::block_on(self.map_read(start, size))
```
kvark, line 1145: I'm not confident about this being a good idea. Mapping a buffer is async not only because it depends on client-server separation in a browser, but also because the user may be currently used by GPU. In this case, we don't want the users to block on it. I'm thinking maybe we have only the `map_read_sync` but keep the write version only async, to avoid users shooting themselves in da foots when uploading data.

kvark (changes requested):
Thank you!
Main question is - did you test this?
I wonder if this actually works, since blocking the thread means there is nothing else that will be happening other than polling on futures. Something needs to call `device.poll()` in order for the futures to advance.

src/lib.rs
zesterer: That makes sense. I'll make this change in a little while.

lachlansneff:
Yeah, this won't work for buffer mapping. `Adapter::request` and `adapter.request_device` work because they're just faking that they're asynchronous in the current implementation.

kvark:
> Unfortunately, this won't work for buffer mapping. Adapter::request and adapter.request_device work because they're just faking that they're asynchronous in the current implementation.

And that means it will not work when apps are built for the Web, so this new functionality would need to be a part of the "native" feature.

zesterer:
@kvark Would it be acceptable to have these functions call `Device::poll` prior to polling their own futures?

kvark:
They'd need to somehow do the request first, then poll(), and only then block on an executor. This doesn't work with `async` functions, it requires us to return a `Future` like we do for mapping.

lachlansneff:
Well, you could have a modified gpu future implementation call device.poll internally...

zesterer:
@kvark Is a spurious call to `Device::poll` prior to polling the future a problem?

grovesNL (commented):
This is an interesting idea, but as I mentioned on Matrix I'm a bit concerned with exposing an optional `sync` feature. Crates will tend towards this by default because it's easier to use, but by requiring these functions to block internally we don't have a good way of implementing this on the web backend (e.g. they would need to be hidden behind `#[cfg(target_arch = "wasm32")]` which may cause portability issues).

IMO we should prefer to let users choose if they want all async operations to block by allowing deciding which executor they run the futures from, which we already get by returning futures from the async API (e.g. the user calls `pollster::block_on`/`futures::executor::block_on`/`async_std::task::block_on`/`tokio::runtime::Runtime::block_on` wherever they prefer).

I think it would also be really helpful if we provide a great way to integrate device polling with some kind of event loop when running natively, which is one of the bigger issues with the futures we return here at the moment.

kvark:
Project moved to `wgpu`

PR #250, opened 2020-04-09
Migrate CI to GitHub Actions

### LINKED ISSUE #225: Move CI to Github Actions
How about use CI with Github Actions ?

Gordon-F:
fix #225


kvark:
Where is the GH actions status? or is it not supposed to be here? That would mean, we don't have any kind of CI on this commit.

Gordon-F:
@kvark I think we can't see CI status until we merge this PR. Only in fork repo https://github.com/Gordon-F/wgpu-rs/actions/runs/74829862

@grovesNL thx. I'll try to fix bors.

Gordon-F:
@kvark @grovesNL I think all done. Can we try merge this PR without bors?

PR #251, opened 2020-04-11
cargo fmt the code and tweak format rules

chao-mu:
Ran cargo fmt after removing rules per @kvark, two of which were because they needed nightly to run.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -33,36 +33,96 @@ macro_rules! vertex_attr_array {
 // For internal usage
 #[macro_export]
 macro_rules! vertex_format_size {
-    (Uchar2) => { 2 };
-    (Uchar4) => { 4 };
-    (Char2) => { 2 };
-    (Char4) => { 4 };
-    (Uchar2Norm) => { 2 };
-    (Uchar4Norm) => { 4 };
-    (Char2Norm) => { 2 };
-    (Char4Norm) => { 4 };
-    (Ushort2) => { 4 };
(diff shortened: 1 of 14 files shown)


src/macros.rs
```diff
@@ -33,36 +33,96 @@ macro_rules! vertex_attr_array {
 // For internal usage
 #[macro_export]
 macro_rules! vertex_format_size {
-    (Uchar2) => { 2 };
```
kvark, line 36: can we maybe change these to avoid `{` and `}` entirely, so that rustfmt doesn't put the value on a separate line?

chao-mu: How might you suggest things be changed to this effect? Also, I looked into rust-fmt settings and unfortunately while there exists something for this (format_macro_bodies=false) it is not stable yet.

kvark: I was thinking we could edit this to be `(Uchar2) => 2,` and so forth before/after formatting

chao-mu: That doesn't seem to be valid syntax, or I am doing something wrong. When I go to build, I get this error:
```
error: no rules expected the token `,`
  --> src/macros.rs:36:18
   |
36 |     (Uchar2) => 2,
   |                  ^ no rules expected this token in macro call

```
Also, doing 
```
(Uchar2) => (2);
```
Auto expands to the multiline {} when I save it with my editor which runs rust-fmt automatically (so presumably also with cargo fmt).

PR #256, opened 2020-04-14
Update wgpu for texture descriptor changes

kvark:
Updates for https://github.com/gfx-rs/wgpu/pull/577
Detaches us from crates-io dependency (again).

### DIFF
--- a/examples/skybox/main.rs
+++ b/examples/skybox/main.rs
@@ -180,15 +180,12 @@ impl framework::Example for Skybox {
             })
             .collect::<Vec<_>>();
 
-        let texture_extent = wgpu::Extent3d {
-            width: image_width,
-            height: image_height,
-            depth: 1,
-        };
-
         let texture = device.create_texture(&wgpu::TextureDescriptor {
-            size: texture_extent,
-            array_layer_count: 6,
+            size: wgpu::Extent3d {
+                width: image_width,
+                height: image_height,
+                depth: 6,
+            },
             mip_level_count: 1,
             sample_count: 1,
             dimension: wgpu::TextureDimension::D2,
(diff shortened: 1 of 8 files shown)


examples/skybox/main.rs
```diff
@@ -180,15 +180,12 @@ impl framework::Example for Skybox {
...
-        };
-
         let texture = device.create_texture(&wgpu::TextureDescriptor {
-            size: texture_extent,
-            array_layer_count: 6,
+            size: wgpu::Extent3d {
+                width: image_width,
+                height: image_height,
+                depth: 6,
+            },
```
mitchmindtree, line 188: Ahh interesting, so `depth` now takes the place of `array_layer_count`? Is there somewhere I could read more about this decision?

One way that we started using layers within nannou is for efficient animation of short image sequences in a manner that allows for random access, reverse playback, different playback rates, etc. The HAP video codec allows for doing this in a more efficient and general manner for files from disk, but is also vastly more complicated to setup for users compared to a simple image sequence. Video decoding is something I might bring up in another issue in the future :)

You can find an example of this texture layer usage at the [wgpu_image_sequence.rs example](https://github.com/nannou-org/nannou/blob/master/examples/wgpu/wgpu_image_sequence/wgpu_image_sequence.rs#L158-L166) at the nannou repo. How would you recommend working around this with the new API? Perhaps using depth instead as you are here, but rather than taking a `TextureView` of the layer we want, we use a uniform buffer to specify what depth we currently want to sample from?

kvark: The internal logic is the same, it's only the API change here, since we figured that there is a confusion between depth slices and array layers. Discussion was in https://github.com/gpuweb/gpuweb/pull/613

You should be able to continue doing what you are doing :) Just need to update the TextureDescriptor.

PR #257, opened 2020-04-14
Improve error message when trying to create an empty buffer

HalfVoxel:


HalfVoxel:
Ah, that's even better :)
In my code I already implemented a workaround which, in case I try to allocate an empty buffer, allocates a 4 byte buffer instead, but pretends it is 0 bytes large.

If I understand the vulkan spec correctly, the spec explicitly prohibits allocating buffers smaller than 4 bytes.

kvark:
Yes, Vulkan doesn't like that. It also doesn't like many other things:)
We think there is a use case worth supporting, and having support for these buffers is easy.

HalfVoxel:
True.
Yes, I too think this is a very useful use case. It removes a lot of special cases in application code.

PR #259, opened 2020-04-17
Remove zerocopy and replace with bytemuck.

### LINKED ISSUE #146: Alternative to ZeroCopy
Unfortunately, ZeroCopy take a really long time to build. It's a part of Fuchsia code base, and communicating with developers or filing issues seems... problematic. Perhaps, we could make a simpler version of it for ourselves, or find an existing alternative.

StarArawn:
fixes #146 

I've removed `zerocopy` from the examples and replaced it with `bytemuck`. I ran all of the examples/tests and everything ran great in vulkan on my windows box.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,7 @@ log = "0.4"
 png = "0.15"
 winit = "0.22"
 rand = "0.7.2"
-zerocopy = "0.3"
+bytemuck = "1"
 futures = "0.3"
 
 [[example]]
--- a/examples/boids/main.rs
+++ b/examples/boids/main.rs
@@ -7,7 +7,6 @@ extern crate rand;
 mod framework;
 
 use std::fmt::Write;
-use zerocopy::AsBytes;
 
 use wgpu::vertex_attr_array;
 
@@ -165,7 +164,7 @@ impl framework::Example for Example {
 
         let vertex_buffer_data = [-0.01f32, -0.02, 0.01, -0.02, 0.00, 0.02];
         let vertices_buffer = device.create_buffer_with_data(
-            vertex_buffer_data.as_bytes(),
+            bytemuck::bytes_of(&vertex_buffer_data),
             wgpu::BufferUsage::VERTEX | wgpu::BufferUsage::COPY_DST,
         );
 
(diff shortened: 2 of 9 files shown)


Cargo.toml
```diff
@@ -55,7 +55,7 @@ log = "0.4"
 png = "0.15"
 winit = "0.22"
 rand = "0.7.2"
-zerocopy = "0.3"
+bytemuck = "1.2.0"
```
kvark, line 58: does it have to be "1.2.0" specifically, or could be just "1"?

examples/cube/main.rs
```diff
@@ -123,10 +126,10 @@ impl framework::Example for Example {
         let (vertex_data, index_data) = create_vertices();
 
         let vertex_buf =
-            device.create_buffer_with_data(vertex_data.as_bytes(), wgpu::BufferUsage::VERTEX);
+            device.create_buffer_with_data(bytemuck::try_cast_slice(&vertex_data).unwrap(), wgpu::BufferUsage::VERTEX);
```
kvark, line 129: why not just `cast_slice` if we are unwrapping anyway?

examples/boids/main.rs
```diff
@@ -204,7 +203,7 @@ impl framework::Example for Example {
         let mut particle_bind_groups = Vec::<wgpu::BindGroup>::new();
         for _i in 0..2 {
             particle_buffers.push(device.create_buffer_with_data(
-                initial_particle_data.as_bytes(),
+                bytemuck::try_cast_slice(&initial_particle_data).unwrap(),
```
kvark, line 206: similarly in all cases: if we are just unwrapping, no need to `try_`

Cargo.toml
kvark: this is unresolved?

StarArawn:
@kvark 
Sorry the file didn't save and I thought it did! 😓

Cargo.toml
```diff
@@ -55,7 +55,7 @@ log = "0.4"
 png = "0.15"
 winit = "0.22"
 rand = "0.7.2"
-zerocopy = "0.3"
+bytemuck = "1.0.0"
```
kvark, line 58: just "1" is good :)

kvark:
Whoa, github actions are so fast!

rukai:
Unfortunate that the examples now need unsafe...

kvark:
@rukai true. Is this something you wanted to avoid? Technically, there could be a crate that implements `Pod` and `Zeroable` by reflecting the struct contents and ensuring this is safe. The benefit of going with this PR is that this crate would not need to be forced upon the users, and they don't have to pay for the compile times (which zerocopy was really bad at).

StarArawn:
@rukai @kvark 
There is an issue open for deriving instead of implementing, and you'd need a separate crate for it to work:
https://github.com/Lokathor/bytemuck/issues/4

PR #262, opened 2020-04-19
Reverse srgb in hello-triangle

grovesNL:
Reverse srgb support in hello-triangle (these were backwards by mistake)


grovesNL:
@kvark I removed the later commit, but the first commit is still needed

PR #271, opened 2020-04-22
Improve docs for BindingType

HalfVoxel:
I think everything should be correct.

The readonly flags seem obvious at face value, but I couldn't find anything in the specifications about them (other than some unrelated validation), so I didn't add any documentation for those fields.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -242,28 +242,96 @@ pub struct Binding<'a> {
     pub resource: BindingResource<'a>,
 }
 
-/// Specific type of a binding..
+/// Specific type of a binding.
+/// WebGPU spec: https://gpuweb.github.io/gpuweb/#dictdef-gpubindgrouplayoutentry
 #[derive(Copy, Clone, Debug, Hash, Eq, PartialEq)]
 pub enum BindingType {
+    /// A buffer for uniform values.
+    ///
+    /// Example GLSL syntax:
+    /// ```
+    /// layout(std140, binding = 0)
+    /// uniform Globals {
+    ///     vec2 aUniform;
+    ///     vec2 anotherUniform;
+    /// };
+    /// ```
     UniformBuffer {
+        /// Indicates that the binding has a dynamic offset.
+        /// One offset must be passed to [RenderPass::set_bind_group] for each dynamic binding in increasing order of binding number.
         dynamic: bool,
     },
+    /// A storage buffer.
+    ///
+    /// Example GLSL syntax:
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -242,27 +242,65 @@ pub struct Binding<'a> {
...
 }
 
-/// Specific type of a binding..
+/// Specific type of a binding.
+/// WebGPU spec: https://gpuweb.github.io/gpuweb/#dictdef-gpubindgrouplayoutentry
 #[derive(Copy, Clone, Debug, Hash, Eq, PartialEq)]
 pub enum BindingType {
+    /// A buffer for uniform values.
+    ///
+    /// Example GLSL syntax:
```
kvark, line 251: this will need to be changed to WGSL soon :)

src/lib.rs
```diff
@@ -242,27 +242,65 @@ pub struct Binding<'a> {
...
     },
+    /// A sampler that can be used to sample a texture.
+    ///
+    /// Example GLSL syntax:
+    /// ```
+    /// layout(binding = 0)
+    /// uniform sampler s;
+    /// ```
     Sampler {
+        /// Use as a comparison sampler.
```
kvark, line 278: might need to reword that to say that it *must* be used as a comparison sampler

src/lib.rs
```diff
@@ -242,27 +242,65 @@ pub struct Binding<'a> {
...
     UniformBuffer {
+        /// Indicates that the binding has a dynamic offset.
+        /// One offset must be passed to [RenderPass::set_bind_group] for each dynamic binding in increasing order of binding number.
         dynamic: bool,
     },
     StorageBuffer {
+        /// Indicates that the binding has a dynamic offset.
+        /// One offset must be passed to [RenderPass::set_bind_group] for each dynamic binding in increasing order of binding number.
         dynamic: bool,
         readonly: bool,
```
kvark, line 268: The storage buffer can only be read, and has to be annotated correspondingly (with "readonly" qualifier) in GLSL/WGSL. Here is when an example would be useful.

src/lib.rs
HalfVoxel: True. But having some examples is better than none :p

src/lib.rs
HalfVoxel: Fixed

src/lib.rs
HalfVoxel: Added example

PR #272, opened 2020-04-22
Force the wasm32 types to match Send / Sync with the native types

kyren:
This is potentially a controversial change so this is just a draft to discuss it.

Currently, when building wgpu on wasm32, none of the interface types are Send or Sync, whereas on any other platform they are generally Send and Sync and this can cause a lot of pain, especially in the presence of async.

This PR is the nuclear option: lie and forcibly declare nearly all interface types to be Send or Send + Sync.

I think I understand the situation with proposed wasm32 threading a bit better now, and as of right now at least with my current understanding, it seems to me unlikely that javascript values held by rust could *ever* be sent transparently to another worker.  The mechanism that rust apparently will use to implement transparent threading will be to use `SharedArrayBuffer` to share memory between multiple workers.  Javascript provides a way for certain (serializable) objects to be shared with with other workers through `worker.postMessage()`, so Javascript has *some* story for sending javascript values to other threads, but only plain rust values that exist solely in the `SharedArrayBuffer` memory are send-able under this scheme.  I've looked briefly through some of the wasm proposals like the interface types proposal, but I don't have a good sense if there's any proposed mechanism at all for interface types and threads to interact at all.

If that's true that's really unfortunate because it seems like any thread safe API shared by native and web is doomed to mismatch.  This is even more unfortunate because many of the webgpu types in the draft standard are marked as 'Serializable', meaning that they *are* actually safe to send to another worker via `postMessage`, but there does not appear to be a planned way for rust to do this transparently.

I definitely could be misunderstanding the situation though and even if I am understanding it, the situation may have since changed.  I also don't know the timing here, I don't know what the eventual timing of wasm32 threading support will be vs webgpu support and interface types support.  *Right now* this PR is almost certainly sound, but there's definitely a possible future where it becomes unsound, and even maybe a possible future where it becomes unsound but then could later be made sound again.

### DIFF
--- a/src/backend/web.rs
+++ b/src/backend/web.rs
@@ -6,27 +6,66 @@ use crate::{
 };
 
 use futures::FutureExt;
-use std::{marker::PhantomData, ops::Range};
+use std::{
+    future::Future,
+    marker::PhantomData,
+    ops::Range,
+    pin::Pin,
+    task::{self, Poll},
+};
 use wasm_bindgen::prelude::*;
 
-pub type Context = web_sys::Gpu;
+// We need to make a wrapper for some of the handle types returned by the web backend to make them
+// implement `Send` and `Sync` to match native.
+//
+// SAFETY: All webgpu handle types in wasm32 are internally a `JsValue`, and `JsValue` is neither
+// Send nor Sync.  Currently, wasm32 has no threading support so implementing `Send` or `Sync` for a
+// type is (for now) harmless.  Eventually wasm32 will support threading, and depending on how this
+// is integrated (or not integrated) with values like those in webgpu, this may become unsound.
+
+#[derive(Debug, Clone)]
+pub(crate) struct Sendable<T>(T);
+unsafe impl<T> Send for Sendable<T> {}
+unsafe impl<T> Sync for Sendable<T> {}
+
+pub(crate) type Context = Sendable<web_sys::Gpu>;
+
 pub(crate) struct ComputePass(web_sys::GpuComputePassEncoder);
 pub(crate) struct RenderPass(web_sys::GpuRenderPassEncoder);
 
+// We need to assert that any future we return is Send to match the native API.
+//
+// This is safe on wasm32 *for now*, but similarly to the unsafe Send impls for the handle type
+// wrappers, the full story for threading on wasm32 is still unfolding.
+
+pub(crate) struct MakeSendFuture<F>(F);
+
+impl<F: Future> Future for MakeSendFuture<F> {
+    type Output = F::Output;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut task::Context) -> Poll<Self::Output> {
+        // This is safe because we have no Drop implementation to violate the Pin requirements and
+        // do not provide any means of moving the inner future.
+        unsafe { self.map_unchecked_mut(|s| &mut s.0) }.poll(cx)
+    }
+}
+
+unsafe impl<F> Send for MakeSendFuture<F> {}
+
 impl crate::ComputePassInner<Context> for ComputePass {
-    fn set_pipeline(&mut self, pipeline: &web_sys::GpuComputePipeline) {
-        self.0.set_pipeline(pipeline);
+    fn set_pipeline(&mut self, pipeline: &Sendable<web_sys::GpuComputePipeline>) {
+        self.0.set_pipeline(&pipeline.0);
     }
     fn set_bind_group(
         &mut self,
         index: u32,
-        bind_group: &web_sys::GpuBindGroup,
+        bind_group: &Sendable<web_sys::GpuBindGroup>,
         offsets: &[wgt::DynamicOffset],
     ) {
         self.0
             .set_bind_group_with_u32_array_and_f64_and_dynamic_offsets_data_length(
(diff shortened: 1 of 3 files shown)


kyren:
Okay, the situation may be more hopeful (and maybe more murky) than I first thought.

I'll try to summarize the discussion in #wgpu:matrix.org as best I can:

It may be reasonable to have a matching Send / Sync API even without explicit extra wasm support.

It may be that the eventual WebGPU standard explicitly supports workers via integer handles that can be stored in a `SharedArrayBuffer`.  Also it seems fairly straightforward to turn any !Send javascript Promise into a Send future by means of something like a oneshot channel, so async APIs don't seem to be an issue.

And even in the total absence of ways to directly share handles across wasm threads, it may be possible even to transparently send backing objects via `postMessage` somehow.

So in that case, maybe this is actaully a good idea?  I wouldn't want to paint wgpu-rs into a corner and force it into a sub-standard or hacky solution when there is the eventual appearance of rust wasm threading, but it at least doesn't seem totally hopeless?

kyren:
Okay, I rebased this on current master, made more web backend handle types into Send + Sync, added bounds to the `Context` trait to ensure that things match on native and web, and also incorporated #294 because it's very closely related.

I'm marking this as ready for review, but I'm still up for discussing a lot of the specifics here, it's quite possible I made the `Context` bounds too strict for example.  I'm going ahead and marking it as ready for review though and I'll be around quickly to fix anything, because it's quite an invasive change and rebasing it is pretty hard, so if at all possible it would be nice for it not to sit for too long.

src/lib.rs
```diff
@@ -84,37 +84,38 @@ trait RenderPassInner<Ctx: Context> {
...
+    type BindGroupLayoutId: Send + Sync;
+    type BindGroupId: Send + Sync;
+    type TextureViewId: Send + Sync;
+    type SamplerId: Send + Sync;
+    type BufferId: Send + Sync;
+    type TextureId: Send + Sync;
+    type PipelineLayoutId: Send + Sync;
+    type RenderPipelineId: Send + Sync;
+    type ComputePipelineId: Send + Sync;
+    type CommandEncoderId: Send + Sync;
```
kvark, line 100: we don't have send/sync on command encoders, so I don't think this line is needed

src/backend/web.rs
```diff
@@ -6,27 +6,143 @@ use crate::{
...
+// We need to make wrappers for the handle types returned by the web backend to make them implement
+// `Send` and `Sync` to match native.
+//
+// SAFETY: All webgpu handle types in wasm32 are internally a `JsValue`, and `JsValue` is neither
+// Send nor Sync.  Currently, wasm32 has no threading support so implementing `Send` or `Sync` for a
+// type is (for now) harmless.  Eventually wasm32 will support threading, and depending on how this
+// is integrated (or not integrated) with values like those in webgpu, this may become unsound.
+
+#[derive(Debug, Clone)]
+pub(crate) struct Context(web_sys::Gpu);
```
kvark, line 21: could we have it more generically, i.e.
```rust
#[derive(Debug, Clone)]
struct Sendable<T>(T);
unsafe impl Send for Context {}
unsafe impl Sync for Context {}
```

src/backend/web.rs
```diff
@@ -6,106 +6,45 @@ use crate::{
...
-
-#[derive(Debug, Clone, PartialEq)]
-pub(crate) struct SwapChainId(web_sys::GpuSwapChain);
-unsafe impl Send for SwapChainId {}
-unsafe impl Sync for SwapChainId {}
+pub(crate) struct Sendable<T>(T);
+unsafe impl<T> Send for Sendable<T> {}
+unsafe impl<T> Sync for Sendable<T> {}
+
+pub(crate) type Context = Sendable<web_sys::Gpu>;
```
kvark, line 31: Why/where do we need these typedefs?

kyren: We don't, but you end up needing them in a lot of places, I can get rid of the typedefs (other than Context which is required) if you'd like.

kyren:
> Also, we'll need this squashed please :)

Sure thing I'll squash as soon as you're done reviewing since it makes reviewing harder.

src/backend/web.rs
kyren: And certainly all of the type aliases other than `Context` don't need to be `pub(crate)`, that was a mistake.

kvark: In the `impl Context` I think you can just use `Self::Buffer` instead of `Buffer` typedef, and same for everything else.
In the implementations for the passes we could just write plain `Sendable<websys::Buffer>`.
Would you be open for this? I think the less types internally defined, the better.

kyren: Yeah that's completely fine I'll do that right now.

kyren:
Okay squashed and fixed some bad formatting.

lachlansneff:
Apologies for #294 not working off the bat.

kyren:
> Apologies for #294 not working off the bat.

Oh no it definitely did on native, I'm sorry I had to merge it into this, it was kind of awkward to separate them.

lachlansneff:
No worries! This isn't a competition, it's collaboration.

PR #278, opened 2020-04-24
Add screenshots of examples

chinedufn:
This PR is geared towards people that are browsing the examples on GitHub looking to get a better sense of what the examples are showing.

Not much going on just a bunch of READMEs and screenshots.


.gitignore
```diff
@@ -9,11 +9,9 @@ Cargo.lock
 # These are backup files generated by rustfmt
 **/*.rs.bk
 
-# Other
-.DS_Store
-
-# VSCode project
+# IDE/Editor configuration files
```
grovesNL, line 12: Could we keep `.DS_Store` here please?

examples/hello-compute/README.md
```diff
@@ -0,0 +1,30 @@
...
+- If n is even, n = n/2
+- If n is odd, n = 3n+1
+
+that it will take to finish and reach the number `1`.
+
+## To Run
+
+```
+# Pass in any 4 numbers as arguments
+RUST_LOG=info cargo run --example hello-compute 1 2 3 4
```
grovesNL, line 15: We could limit the log info to `hello_compute` only by using something like `RUST_LOG=hello_compute cargo run --example hello-compute 1 2 3 4`, I think this would make the output clearer

examples/hello-triangle/README.md
```diff
@@ -0,0 +1,13 @@
+# hello-triangle
+
+This example renders a triangle to a native window.
```
grovesNL, line 3: This could probably be "window" instead of "native window" because it also runs on the web now :)

.gitignore
chinedufn: Ahhh accident sorry. I’ll add it back in the morning

examples/hello-compute/README.md
chinedufn: Didn’t even know you could do that thanks - will do

examples/hello-triangle/README.md
chinedufn: Will do! The running on the web part was what got me starting to take a look :)

chinedufn:
All feedback addressed!

grovesNL:
Thanks! Could we also remove the `.DS_Store`s which were committed?

chinedufn:
Sorry about that - removed!

grovesNL:
bors retry

PR #281, opened 2020-04-26
The Context trait

### LINKED ISSUE #156: Case for using wgpu-core

### LINKED ISSUE #267: Consume command buffers on `submit`
This would reflect the semantics of the API. You can't re-use them anyway.

kvark:
The main motivation here is to avoid blocking the wgpu-core updates by `wgpu-native`. Instead, `wgpu-native` becomes a branch, and the dependency of `wgpu-rs` -> `wgpu-native` starts adhering to the contract/API of the standard webgpu-native headers.

The biggest change is the introduction of the Context trait. I recall us discussing 2 downsides to having this trait:
  1. inconvenient for the users to include. This is a non-issue here, since it's private.
  2. more code to maintain. This is less of an issue if we aim to have 3 backends.

What this gives in return is a well established contract with the backends. Unlike gfx-rs, the backend code is right here, a part of the crate, so the contract is only for internal use.

Fixes #156 : the "direct" implementation of it goes straight to wgpu-core. What this gives us is less overhead for command recording, since there is no longer an extra indirection on every command, and no heap allocation at the end of a render pass.

The downside of this PR is one extra `Arc` (with addref) per object.

This commit also has small improvements:
- consuming command buffers on submit (Fixes #267)
- Instance type
- proper call to destructors
- fallible `request_device`


src/lib.rs
```diff
@@ -1137,29 +1554,36 @@ impl<'a> ComputePass<'a> {
...
+                .encoder_end_compute_pass(&self.parent.id, &mut self.id);
         }
     }
 }
 
 impl Queue {
     /// Submits a series of finished command buffers for execution.
-    pub fn submit(&self, command_buffers: &[CommandBuffer]) {
-        backend::queue_submit(&self.id, command_buffers);
+    pub fn submit<I: IntoIterator<Item = CommandBuffer>>(&self, command_buffers: I) {
```
lachlansneff, line 1574: Couldn't this be inefficient for webgpu.h, since that takes a pointer to a buffer of command buffers? You'd have to allocate a Vec or similar to use that. Whereas, a slice could just be passed efficiently.

src/lib.rs
```diff
@@ -583,126 +893,190 @@ impl CreateBufferMapped<'_> {
...
 
+    /// Creates a surface from a raw window handle.
+    pub fn create_surface<W: raw_window_handle::HasRawWindowHandle>(&self, window: &W) -> Surface {
+        Surface {
+            id: self.context.instance_create_surface(window),
+        }
+    }
+
+    #[cfg(any(target_os = "ios", target_os = "macos"))]
+    pub fn create_surface_from_core_animation_layer(
```
lachlansneff, line 935: This seems like it should be unsafe (since the layer pointer can forged) and also in an platform extension trait?

src/lib.rs
```diff
@@ -583,126 +893,190 @@ impl CreateBufferMapped<'_> {
...
         }
     }
-}
 
-impl Adapter {
     /// Retrieves all available [`Adapter`]s that match the given backends.
     #[cfg(not(target_arch = "wasm32"))]
-    pub fn enumerate(backends: BackendBit) -> Vec<Self> {
-        wgn::wgpu_enumerate_adapters(backends)
+    pub fn enumerate_adapters(&self, backends: wgt::BackendBit) -> Vec<Adapter> {
```
lachlansneff, line 916: Thoughts on returning an iterator here instead of collecting?

src/lib.rs
```diff
@@ -21,24 +22,288 @@ pub use wgt::{
...
+    fn draw_indexed(&mut self, indices: Range<u32>, base_vertex: i32, instances: Range<u32>);
+    fn draw_indirect(&mut self, indirect_buffer: &Ctx::BufferId, indirect_offset: BufferAddress);
+    fn draw_indexed_indirect(
+        &mut self,
+        indirect_buffer: &Ctx::BufferId,
+        indirect_offset: BufferAddress,
+    );
+}
+
+trait Context: Sized {
```
lachlansneff, line 86: I think the traits should be in a different file so it's easier to read this one.

src/lib.rs
kvark: Nah, this is more flexible than a slice. You can always go from a slice to an iterator for free, but not the other way around. I.e. `wgpu-native` would do something like: `command_buffer_slice.iter().cloned()` for passing it here.

src/lib.rs
kvark: Yeah, it's the same as for `create_surface`. I'll make those unsafe. It doesn't matter much, since this is internal API.

src/lib.rs
kvark: Sounds good!

src/lib.rs
kvark: It could be, yes. I find it convenient here, because it uses all the types in the proximity, and everything uses it. We can follow-up with moving it if we see fit, after the initial work is done.

kvark:
@lachlansneff thank you for the review!

> I'm a little wary of all the Arc::clone calls everywhere, especially when compiling to wasm, since the browser backend will already be doing that internally.

It's a fair thing to be concerned about. My understanding is that this is an extremely light overhead, since it only happens at object creation time, which is dominated by other factors.
Overall, I think we can have it, and there will not by any observable effect.

kvark:
@grovesNL I think this is done. Please take a final (?) look.

src/lib.rs
```diff
@@ -583,126 +887,196 @@ impl CreateBufferMapped<'_> {
...
+            .enumerate_adapters(wgc::instance::AdapterInputs::Mask(backends, || PhantomData))
+            .into_iter()
+            .map(move |id| crate::Adapter {
+                id,
+                context: Arc::clone(&context),
+            })
+    }
+
+    /// Creates a surface from a raw window handle.
+    pub unsafe fn create_surface<W: raw_window_handle::HasRawWindowHandle>(
```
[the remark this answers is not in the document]
kvark: I'd very much like that, but not sure if this is achievable via `RawWindowHandle`: after all, it's just a bunch of raw pointers. I.e. a `winit`-based API could be safe, but once we lower to the raw handles, safety zone is gone.

src/backend/web.rs
```diff
@@ -561,653 +499,801 @@ fn map_color(color: wgt::Color) -> web_sys::GpuColorDict {
...
-    crate::Buffer {
-        id: device.create_buffer(&mapped_desc),
-        detail: (),
+
+    fn instance_request_adapter(
+        &self,
+        options: &crate::RequestAdapterOptions<'_>,
+        _backends: wgt::BackendBit,
+    ) -> Self::RequestAdapterFuture {
+        //assert!(backends.contains(wgt::BackendBit::BROWSER_WEBGPU));
```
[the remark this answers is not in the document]
kvark: It's annoying/hard to get this check back in (but possible). I've added a detailed TODO here.

kvark:
@grovesNL thank you for the quick review!
bors r=grovesNL

Issue #284, opened 2020-04-27
How to enable serde feature on wgpu-types through wgpu-rs ?


gzp-crey:
wgpu-types has a feature to allow (de)serialization for types, descriptors and enums. 
How can I enable it through wgpu-rs ?

kvark:
@gzp-crey can't you add a dependency to `wgpu-types` to your project and enable the `serde` feature this way?

gzp-crey:
I'm using the git version right now (to play with the latest wasm support ):
`wgpu = {git = "https://github.com/gfx-rs/wgpu-rs.git", branch = "master"}`

I'm not sure how to add the wgpu-types without some extra check during updates as type is using git revision and I have no idea what to provide for package version in this case. I'd use wgpu-types directly if someone shows and example how to do so :)

Also according to https://doc.rust-lang.org/cargo/reference/features.html#usage-in-packages I think it'd make sense to add the feature but I'm far for being an expert on the topic :).

kvark:
I hear you. So if you have this in your `Cargo.toml`
```toml
wgpu-types = {git = "https://github.com/gfx-rs/wgpu", rev = "5c172dd4756aa152b4f3350e624d7b1b5d24ddda", features = "serde"}
```
This should get you the same result. It may be slightly annoying (I agree!) that whenever you do `cargo update -p wgpu`, you'll need to check what revision of wgpu-types is being used by `wgpu` and adjust your `Cargo.toml` accordingly.

I wonder if it would work to just leave the git path without a revision here, such that Cargo picks the revision it already has for `wgpu` crate... If that works, it would be magically wonderful!

gzp-crey:
Well I've tried it but the no magic happened :( 

in Cargo.toml
```
wgpu-types = {git = "https://github.com/gfx-rs/wgpu", features = ["serde"]}
wgpu = {git = "https://github.com/gfx-rs/wgpu-rs.git", branch = "master"}
```

compile result:
```
 Compiling gfx-backend-vulkan v0.5.6
   Compiling wgpu-types v0.5.0 (https://github.com/gfx-rs/wgpu#d4532350)
   Compiling wgpu-types v0.5.0 (https://github.com/gfx-rs/wgpu?rev=5c172dd4756aa152b4f3350e624d7b1b5d24ddda#5c172dd4)
[...]

PR #285, opened 2020-04-27
expose wgt serde feature

gzp-crey:
#284 
Add serde feature (wgt/serde)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ exclude = ["etc/**/*", "examples/**/*", "tests/**/*", "Cargo.lock", "target/**/*
 default = []
 # Make Vulkan backend available on platforms where it is by default not, e.g. macOS
 vulkan = ["wgc/gfx-backend-vulkan"]
+serde = ["wgt/serde"]
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies.wgc]
 package = "wgpu-core"


Cargo.toml
```diff
@@ -22,6 +22,7 @@ exclude = ["etc/**/*", "examples/**/*", "tests/**/*", "Cargo.lock", "target/**/*
 default = []
 # Make Vulkan backend available on platforms where it is by default not, e.g. macOS
 vulkan = ["wgc/gfx-backend-vulkan"]
+serde = ["wgt/serde"]
```
kvark, line 25: Are there any types you'd like to serialize that are in `wgpu-rs` as opposed to `wgpu-types`? Having `serde` feature would conflict with depending on servo in this case.

kvark: also, this would probably need to include "wgc/serde" as well

grovesNL: @kvark assuming you meant `serde` instead of `servo`, I think we can do have both `serde` as a feature and a dependency using the same strategy we use in wgpu-types (https://github.com/rust-lang/api-guidelines/issues/180#issuecomment-597418796)

kvark: Fortunately, I don't think this is going to be a problem, see https://github.com/gfx-rs/wgpu/pull/619

gzp-crey: > 
> 
> Are there any types you'd like to serialize that are in `wgpu-rs` as opposed to `wgpu-types`? Having `serde` feature would conflict with depending on servo in this case.

In my usecase I have a material description file that is very close to the pipeline. Like required primitive_topology, some settings for the color_stage, vertex attribute that can be turned into 
RenderPipeline. Without the serde feature I'd have to duplicate all these types and write some wrappers and/or From implementations.

To be more specific:
geometry read from file (ex FBX, obj etc) has a vertex format desc, like: Position(Float3, offset), Normal(Float3, offset), etc.
Material(pipeline) has shader descriptions, binding info, uniforms, etc.

In the "engine" there is a map from (material + vertex_format) -> RenderPipeline. During the population of this map the pipeline requirements and vertex formats are matched and thus the same geometry can be used for multiple pipeline. And the connection (binding) b/n the pipeline and geometry is mostly automatic:
```
let m = materials::<VertexPos2Normal3>::create("Z-pass").
let command_buffer = m.render(encoder, geometry_with_pos2_norm3);
```

HeroesGrave: I've wanted to (de)serialize `BindGroupLayoutEntry`.
Although I wonder why it isn't part of wgpu-types in the first place.

kvark: we may ask @grovesNL . In general, only those types go into wgpu-types that can be re-used directly for the Web target.

grovesNL: wgpu-rs [defines its own `BindGroupLayoutEntry`](https://github.com/gfx-rs/wgpu-rs/blob/449924bb428fbed88b9b3c94e08dee56ed77b06d/src/lib.rs#L647), so we shouldn't need to share the `BindGroupLayoutEntry` from wgpu-core

PR #291, opened 2020-04-29
Update for wgpu-core

kvark:
Depends on https://github.com/gfx-rs/wgpu/pull/619
Closes #285

### DIFF
--- a/examples/hello-compute/main.rs
+++ b/examples/hello-compute/main.rs
@@ -3,7 +3,7 @@ use std::{convert::TryInto, str::FromStr};
 async fn run() {
     let numbers = if std::env::args().len() <= 1 {
         let default = vec![1, 2, 3, 4];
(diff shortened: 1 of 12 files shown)


examples/hello-compute/main.rs
```diff
@@ -12,8 +12,10 @@ async fn run() {
             .collect()
     };
 
-    // To see the output, run `RUST_LOG=info cargo run --example hello-compute`.
-    log::info!("Times: {:?}", execute_gpu(numbers).await);
+    let times = execute_gpu(numbers).await;
+    println!("Times: {:?}", times);
```
grovesNL, line 16: IMO it's slightly distracting to special-case wasm with `cfg` here  – I'd prefer to force the log level in `main` or pass a logging function to `run` if we want to ensure it's displayed by default

kvark: ok, sure. I just got confused by the lack of output when running it

PR #295, opened 2020-05-02
Remove wasm-bindgen patches

rukai:
A new release occurred with the required fixes, so we no longer need to patch it.


grovesNL:
bors r-
Actually could we set the patch version to the latest for these?

rukai:
Good idea!
Done.

grovesNL:
Thanks!
bors r+

PR #302, opened 2020-05-08
Add BufferRange struct for #199

paulkernfeld:
- Use in set_index_buffer
- Use in set_vertex_buffer
- Use in BindingResource

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,7 +6,13 @@ mod backend;
 mod macros;
 
 use futures::FutureExt as _;
-use std::{future::Future, marker::PhantomData, ops::Range, sync::Arc, thread};
+use std::{
+    future::Future,
+    marker::PhantomData,
+    ops::{Bound, Range, RangeBounds},
+    sync::Arc,
+    thread,
+};
 
 #[cfg(not(target_arch = "wasm32"))]
 pub use wgc::instance::{AdapterInfo, DeviceType};
@@ -339,6 +345,21 @@ pub struct Buffer {
     //detail: <C as Context>::BufferDetail,
 }
 
+/// A description of what portion of a buffer to use
+pub struct BufferSlice<'a> {
+    buffer: &'a Buffer,
+    offset: BufferAddress,
+    size: Option<BufferAddress>,
+}
+
+impl<'a> BufferSlice<'a> {
+    /// This fn can be used for calling lower-level APIs where `0` denotes that the slice should
+    /// extend to the end of the buffer.
+    fn size_or_0(&self) -> BufferAddress {
+        self.size.unwrap_or(0)
+    }
+}
+
 /// A handle to a texture on the GPU.
 pub struct Texture {
     context: Arc<C>,
@@ -535,10 +556,7 @@ pub struct Queue {
 
 /// A resource that can be bound to a pipeline.
 pub enum BindingResource<'a> {
-    Buffer {
-        buffer: &'a Buffer,
-        range: Range<BufferAddress>,
-    },
+    Buffer(BufferSlice<'a>),
     Sampler(&'a Sampler),
     TextureView(&'a TextureView),
 }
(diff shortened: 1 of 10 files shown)


paulkernfeld:
I haven't yet done an audit of all the places where I'm doing `unwrap_or(0)` to make sure that that is the correct thing to do.

paulkernfeld:
All right, I pushed this change. It really makes the usage more terse!

A couple thoughts as I was writing this:

- Using `impl RangeBounds` opens up the door for users to pass in some crazy inputs like inclusive upper bounds. For now I just panicked in this case.
- None of the examples actually use a meaningful slice, they all do `.slice(..)`. If we want to optimize for this use case, we could probably make a trait like `BufferView` that is implemented by both `Buffer` and `BufferSlice`. Then, users could just do like `pass.set_vertex_buffer(0, &entity.vertex_buf);`.
- I'm getting a bit nervous that I might be doing some math wrong here. What do you think about adding some tests for slicing? I'm not sure what the long-term testing plan is.

kvark:
> Using impl RangeBounds opens up the door for users to pass in some crazy inputs like inclusive upper bounds. For now I just panicked in this case.

Why don't we just support it? I.e. inclusive range from A to B just means the size is "B - A + 1", there shouldn't be any magic or blockers in this.

> If we want to optimize for this use case, we could probably make a trait like BufferView

Nah, let's not optimize for this specifically. `slice(..)` is pretty concise already :)

> What do you think about adding some tests for slicing? I'm not sure what the long-term testing plan is.

Supposedly, it's just a single pure function that gets `impl RangeBounds` and total size, and returns an exact range. My feeling is that the amount of code needed to test it would be much more than the code in function itself (which is what, 2 lines?). Feel free to add the tests if you feel they are useful here.

src/lib.rs
```diff
@@ -339,6 +339,27 @@ pub struct Buffer {
     //detail: <C as Context>::BufferDetail,
 }
 
+/// A description of what portion of a buffer to use
+pub struct BufferSlice<'a> {
+    buffer: &'a Buffer,
+    start_bound: Option<BufferAddress>,
```
kvark, line 345: why don't we call them `offset: BufferAddress` and `size: Option<BufferAddress>`? No need to have `start_bound` as optional.

src/lib.rs
```diff
@@ -339,6 +339,21 @@ pub struct Buffer {
...
+pub struct BufferSlice<'a> {
+    buffer: &'a Buffer,
+    offset: BufferAddress,
+    size: Option<BufferAddress>,
+}
+
+impl<'a> BufferSlice<'a> {
+    /// This fn can be used for calling lower-level APIs where `0` denotes that the slice should
+    /// extend to the end of the buffer.
+    pub fn size_or_0(&self) -> BufferAddress {
```
kvark, line 352: should probably be private, it doesn't make sense for the wgpu-rs client

src/lib.rs
```diff
@@ -1167,6 +1179,26 @@ impl Drop for BufferWriteMapping {
 }
 
 impl Buffer {
+    /// Use only a portion of this Buffer for a given operation. Choosing a range with 0 size will
+    /// return a slice that extends to the end of the buffer.
+    pub fn slice<S: std::ops::RangeBounds<BufferAddress>>(&self, bounds: S) -> BufferSlice {
+        let offset = match bounds.start_bound() {
+            std::ops::Bound::Included(&bound) => bound,
```
kvark, line 1186: can we have `use std::ops::{Bound, RangeBounds}` somewhere at the start of the file?

src/lib.rs
```diff
@@ -1167,6 +1179,26 @@ impl Drop for BufferWriteMapping {
...
+            std::ops::Bound::Excluded(&bound) => bound + 1,
+            std::ops::Bound::Unbounded => 0,
+        };
+        let end_bound = match bounds.end_bound() {
```
kvark, line 1190: nit: could do `let size = ` here right away instead of matching the option later on

paulkernfeld:
Great, will do!

PR #306, opened 2020-05-11
Improve example friendliness: Don't busy-redraw in examples

khoek:
Looking around for solid Rust graphics libraries, I was a bit shocked when the `wgpu-rs` examples made my whole X desktop environment in Ubuntu laggy---maximizing/minimizing other windows or moving any of them had very noticeable latency when an example was running.

This almost turned me off, but after playing around I found that this was just because of the
```rust
match event {
            event::Event::MainEventsCleared => window.request_redraw(),
            ...
```
in all of the examples. Trying to add the least code possible I've replaced a `ControlFlow::Poll` with `ControlFlow::WaitUntil(...)` in the event loops and capped the redraws-per-second below 50, which completely solves this problem for me (plus the window resize response on the examples themselves are much improved, etc.).

I was just going for an unintrusive fix---this is by no means a perfect solution, but I think in the worst case it won't be any worse that what was there originally. Plus, I think it will make people like me who try to start by copying an example more likely to stick around in the short term.

### DIFF
--- a/examples/framework.rs
+++ b/examples/framework.rs
@@ -1,3 +1,4 @@
+use std::time;
 use winit::{
     event::{self, WindowEvent},
     event_loop::{ControlFlow, EventLoop},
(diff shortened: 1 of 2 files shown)


grovesNL:
Thanks! I'm not sure about using a timer here -- we should be able to rely on vsync if we'd prefer this behavior. Maybe we could prefer `Fifo` instead of `Mailbox` for the examples (assuming `Fifo` fixes the environment lag for you)?

khoek:
@grovesNL Ok, that sounds less hacky. But unfortunately changing `Mailbox` -> `Fifo` in `framework.rs` and running (e.g.) `boids` doesn't seem to change the situation for me. Do you know why?

khoek:
The "situation" being 1) resizing the window being extremely slow (~1fps), 2) `gnome-shell` cpu usage constantly at ~35% (of one core), and 3) this behaviour where all of the other windows lag (I would guess ~10 fps while dragging the browser I'm typing this on around). All of these things go away with my commit, if that helps tracking-down.

Also I guess there is 4) the 50% of times the window pauses for a full second or two doing nothing after a `Control-C` in the terminal running it, and then closes.

kvark:
I actually like this quite a bit! Thank you for coming in with a patch, @khoek !
@grovesNL relying on the back pressure from the swapchain could work too, but it would also block the whole event loop for that duration. This means any inputs received between the request of a new frame and the actual start of recording will not be taken into account...

I suppose another way we could approach this is by introducing something like `Swapchain::get_next_texture_if_ready`, which would try to acquire it without waiting, and return `None` if waiting is needed. But even then, busy looping in the event loop until that texture becomes available is not the best. There must be some kind of a timeout set between the loops.

grovesNL:
@kvark alright, this seems ok for now then. I'm mostly interested in away to handle frame pacing consistently between native and web (e.g. it would be nice to avoid `cfg` blocks here).

For the web we want to use `ControlFlow::Poll` (at the moment this is the only way to use `requestAnimationFrame` in winit's web backend AFAIK), so longer term it would be nice to find a way to make this work well on native too.

PR #307, opened 2020-05-11
Update wgpu and use write_buffer in the examples

kvark:
Depends on https://github.com/gfx-rs/wgpu/pull/653
IMO, makes the examples quite a bit nicer!

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -3,7 +3,7 @@ name: Publish
 on:
   push:
     branches:
-      - master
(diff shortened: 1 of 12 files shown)


kvark:
bors r=nobody

kvark:
@grovesNL hmm, this will break the examples on the Web, since we don't have writeBuffer implemented yet in Firefox. I guess we can wait for it a week or two, although it's unfortunate.

kvark:
I like the idea. Let's see if this works.
bors r=grovesNL

PR #323, opened 2020-05-25
Rename SwapChain::get_next_texture to SwapChain::get_next_frame, and return errors to the user.

AlphaModder:
Depends on https://github.com/gfx-rs/wgpu/pull/668.

This creates a new type `SwapChainResult` just for the method. 
It also alters the signature of `Context::swap_chain_get_next_texture` to return 
```rust
(Option<Self::TextureViewId>, wgt::SwapChainStatus, Self::SwapChainOutputDetail)
```
which is a little ugly but I couldn't think of anything better.

### DIFF
--- a/examples/framework.rs
+++ b/examples/framework.rs
@@ -44,7 +44,7 @@ pub trait Example: 'static + Sized {
     fn update(&mut self, event: WindowEvent);
     fn render(
         &mut self,
-        frame: &wgpu::SwapChainOutput,
+        frame: &wgpu::SwapChainTexture,
         device: &wgpu::Device,
         queue: &wgpu::Queue,
     ) -> wgpu::CommandBuffer;
@@ -143,7 +143,6 @@ async fn run_async<E: Example>(event_loop: EventLoop<()>, window: Window) {
                 log::info!("Resizing to {:?}", size);
                 sc_desc.width = size.width;
                 sc_desc.height = size.height;
-                swap_chain = device.create_swap_chain(&surface, &sc_desc);
                 example.resize(&sc_desc, &device, &queue);
             }
             event::Event::WindowEvent { event, .. } => match event {
@@ -164,10 +163,17 @@ async fn run_async<E: Example>(event_loop: EventLoop<()>, window: Window) {
                 }
             },
             event::Event::RedrawRequested(_) => {
-                let frame = swap_chain
-                    .get_next_texture()
-                    .expect("Timeout when acquiring next swap chain texture");
-                let command_buf = example.render(&frame, &device, &queue);
+                let frame = match swap_chain.get_next_frame() {
+                    Ok(frame) => frame,
+                    Err(_) => {
+                        swap_chain = device.create_swap_chain(&surface, &sc_desc);
+                        swap_chain
+                            .get_next_frame()
+                            .expect("Failed to acquire next swap chain texture!")
+                    }
+                };
+
+                let command_buf = example.render(&frame.output, &device, &queue);
                 queue.submit(Some(command_buf));
             }
             _ => {}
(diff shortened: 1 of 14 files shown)


examples/framework.rs
```diff
@@ -172,7 +172,8 @@ async fn run_async<E: Example>(event_loop: EventLoop<()>, window: Window) {
             event::Event::RedrawRequested(_) => {
                 let frame = swap_chain
                     .get_next_texture()
-                    .expect("Timeout when acquiring next swap chain texture");
+                    .output()
+                    .expect("Failed to acquire next swap chain texture");
```
kvark, line 176: we should re-create the swapchain here

examples/hello-triangle/main.rs
```diff
@@ -111,7 +111,8 @@ async fn run(event_loop: EventLoop<()>, window: Window, swapchain_format: wgpu::
             Event::RedrawRequested(_) => {
                 let frame = swap_chain
                     .get_next_texture()
-                    .expect("Timeout when acquiring next swap chain texture");
+                    .output()
+                    .expect("Failed to acquire next swap chain texture");
```
kvark, line 115: while we can leave this without recreation, hello-triangle needs to be simple

src/lib.rs
```diff
@@ -832,6 +832,39 @@ pub struct SwapChainOutput {
     detail: <C as Context>::SwapChainOutputDetail,
 }
 
+/// The result of a `SwapChain::get_next_texture` operation.
+pub enum SwapChainResult {
```
kvark, line 836: We should probably rely on Rust's `Result` a bit more here:
```rust
struct SwapChainFrame {
  output: SwapChainOutput,
  suboptimal: bool,
};
enum SwapChainError {
...
}
...
Result<SwapChainFrame, SwapChainError>
```

AlphaModder: Sounds good.

AlphaModder: It might be nice to try and have some consistency across the project about what the output of a swapchain is actually called though. Right now you obtain it by asking for a "texture" but it's a "frame" that contains an "output" (which the examples then store in a variable called `frame`).

examples/framework.rs
AlphaModder: Will do.

src/lib.rs
kvark: Totally agree, the swapchain terminology is a bit of a mess

kvark:
@AlphaModder please rebase!

src/lib.rs
AlphaModder: What do you think about either of these:
- `get_next_frame` returning `SwapChainFrame` containing `SwapChainTexture` containing `TextureView`
- `get_next_texture` returning `SwapChainOutput` containing `SwapChainTexture` containing `TextureView`

kvark: get_next_frame sounds better!

src/lib.rs
```diff
@@ -835,11 +820,26 @@ pub type TextureViewDescriptor<'a> = wgt::TextureViewDescriptor<Option<&'a str>>
...
-pub struct SwapChainOutput {
+pub struct SwapChainTexture {
     pub view: TextureView,
     detail: <C as Context>::SwapChainOutputDetail,
 }
 
+/// The result of a successful call to `SwapChain::get_next_frame`.
+pub struct SwapChainFrame {
+    pub output: SwapChainTexture,
+    pub suboptimal: bool,
```
Wumpf, line 831: I know I'm too late to the party, but can we please copy this documentation text from here about what "suboptimal" means in there?
https://github.com/gfx-rs/gfx/blob/bbfdebe6135a9161500594443be8aae1f802a8cc/src/hal/src/window.rs#L456
Saw this now and got confused & curious so I followed the trail to find the doc, now it makes sense. But being relatively new to gfx-rs/ that was long trip ;)

(though I'd argue the name is a bit off, nothing that can't be solved with docs though)

kvark: that would make sense! We need more documentation in general

Wumpf: Ah that's encouraging :) - gonna do a pr once I have at least one more doc thing to make it worthwhile.
Overall I think the in-library docs are getting decent already. Manual like docs & samples is ofc much harder.

PR #335, opened 2020-05-31
Enable X11 in Vulkan

kvark:
Closes #334 (as an alternative)
It was disabled by default by https://github.com/gfx-rs/wgpu/pull/678
This PR brings the old behavior back. We'll still need to consider it if makes sense to switch it to be optional.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,8 +44,8 @@ smallvec = "1"
 raw-window-handle = "0.3"
 parking_lot = "0.10"
 
-//Note: we may consider switching this to "dev-dependencies" if users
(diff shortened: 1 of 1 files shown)


rukai:
I'm happy with this over my PR.
If we ever make x11 optional as per the comment, then I believe it should be default feature.
Any developer who does not personally use x11 will be surprised to find their project does not work there. I think most Linux users are unfortunately still on x11(I know I am)

kvark:
The problem with default features is that they can't be platform dependent... so we'd have to go hoops and introduce an intermediate dependency, just to make sure that we only hook up to X11 on Unix and not Darwin.

PR #338, opened 2020-06-01
Add Extension/Limit Interface

cwfitzgerald:
Follow up to https://github.com/gfx-rs/wgpu/pull/690.

This forwards the extension/limit access interface into wgpu-rs.

It appears that the web-sys doesn't actually implement any of the methods we need, so I had them return the defaults. `Device::limits` exists, but it returns `Object` not `GpuLimits`, so doesn't appear usable.

### DIFF
--- a/examples/skybox/main.rs
+++ b/examples/skybox/main.rs
@@ -130,6 +130,7 @@ impl framework::Example for Skybox {
             lod_min_clamp: 0.0,
             lod_max_clamp: 100.0,
             compare: wgpu::CompareFunction::Undefined,
+            anisotropy_clamp: 1,
         });
 
         let paths: [&'static [u8]; 6] = [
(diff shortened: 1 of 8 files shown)


examples/skybox/main.rs
```diff
@@ -130,6 +130,7 @@ impl framework::Example for Skybox {
             lod_min_clamp: 0.0,
             lod_max_clamp: 100.0,
             compare: wgpu::CompareFunction::Undefined,
+            anisotropy_clamp: 1,
```
kvark, line 133: I think we should change most (if not all) sampler descriptors to end with `.. Default::default()`. That would be more future proof, match the web better, and avoid exposing stuff that is not relevant

cwfitzgerald: I agree. Should I change this in this PR? Or should that be separate?

kvark: let's have this merged, the PR just continues the convention of the old code, it's not required to change that

PR #344, opened 2020-06-02
Use the new map-async

kvark:
Depends on https://github.com/gfx-rs/wgpu/pull/675
It changes the API of mapping and removes `create_buffer_mapped`.
The new functionality is not implemented in Gecko yet, so this breaks the web target, for now.


kvark:
> +210 −419

hell yeah!
bors r+

PR #346, opened 2020-06-03
Add Makefile for shader compilation

HanKruiger:
I couldn't find clear instructions on how to compile the shaders in the examples, and had to do some digging. So I added instructions to the README to hopefully save other people's time.

Sidenote: When I compile the shaders as they are now (with my instructions), I end up with different `.spv` shaders with (seemingly) equivalent behaviour. I expect this is due to a different version/environment of `glslang` (see below) but it could be a mistake on my part.

```
> glslangValidator --version        
Glslang Version: 8.13.3743
ESSL Version: OpenGL ES GLSL 3.20 glslang Khronos. 13.3743
GLSL Version: 4.60 glslang Khronos. 13.3743
SPIR-V Version 0x00010500, Revision 3
GLSL.std.450 Version 100, Revision 1
Khronos Tool ID 8
SPIR-V Generator Version 8
GL_KHR_vulkan_glsl version 100
ARB_GL_gl_spirv version 100
```

Cheers!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,20 @@
+# This Makefile generates SPIR-V shaders from GLSL shaders in the examples.
+
+shader_compiler = glslangValidator
+
+# All input shaders.
+glsls = $(wildcard examples/*/*.vert examples/*/*.frag examples/*/*.comp)
+
+# All SPIR-V targets.
+spirvs = $(addsuffix .spv,$(glsls))
+
+.PHONY: default
+default: $(spirvs)
+
+# Rule for making a SPIR-V target.
+$(spirvs): %.spv: %
+	$(shader_compiler) -V $< -o $@
+
+.PHONY: clean
+clean:
+	rm -f $(spirvs)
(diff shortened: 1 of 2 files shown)


HanKruiger:
Happy to help!

I added the `Makefile`. You can now simply run `make` to compile all the shaders, and it doesn't recompile when not necessary.

Out of interest: Are you aware of a timeline for when WGSL will be supported?

Makefile
```diff
@@ -0,0 +1,18 @@
+# This Makefile generates SPIR-V shaders from GLSL shaders in the examples.
+
```
khoek, line 2: Perhaps a `.PHONY`?

HanKruiger: I considered this, but it is very unlikely that a file named `default` or `clean` will be created, and (in my opinion) not worth the extra complexity.

kvark: It's just easier to have all phony targets in `.PHONY` without thinking of whether they can accidentally match the files

kvark:
Would you mind squashing (and maybe adding the phony target) before we merge?

Makefile
HanKruiger: Sure thing, I added them!

PR #348, opened 2020-06-04
Add aribitrary texture size handling to the capture example

rukai:
* User can specify width + height via arguments.
* Defaults to (100, 200) when no arguments. Odd values chosen to make it more likely to pick up issues when testing.
* Converts from bgra as I believe that is the most commonly supported format for swapchains?


examples/capture/main.rs
```diff
@@ -108,14 +127,35 @@ async fn run() {
...
-        let data = output_buffer.get_mapped_range(0, wgt::BufferSize::WHOLE);
-        let mut png_encoder = png::Encoder::new(File::create("red.png").unwrap(), size, size);
+        let padded_buffer = output_buffer.get_mapped_range(0, wgt::BufferSize::WHOLE);
+        // There are two transforms occuring here:
+        // 1.   map bgra -> rgba
+        // 2.   remove padding from buffer
+        let mut real_buffer = vec!(0; width * height * bytes_per_pixel);
+        for x in 0..width as usize {
+            for y in 0..height as usize {
+                let padded_offset = x + y * padded_width as usize;
```
kvark, line 137: let's make this in bytes, and compute as `x * 4 + y * padded_bytes_per_row` instead

examples/capture/main.rs
```diff
@@ -108,14 +127,35 @@ async fn run() {
     }
 
     if let Ok(()) = buffer_future.await {
-        let data = output_buffer.get_mapped_range(0, wgt::BufferSize::WHOLE);
-        let mut png_encoder = png::Encoder::new(File::create("red.png").unwrap(), size, size);
+        let padded_buffer = output_buffer.get_mapped_range(0, wgt::BufferSize::WHOLE);
+        // There are two transforms occuring here:
+        // 1.   map bgra -> rgba
+        // 2.   remove padding from buffer
```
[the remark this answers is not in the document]
rukai: When generating a 1000x1000 image, using Rgba + stream_writer was ~5ms faster then my old implementation.
Assuming the user is rendering in a headless mode i.e. not copying from the swapchain.
Then theres nothing stopping them from configuring their pipelines etc. to use rgba instead of bgra when rendering to an image.
So using rgba + stream_writer seems like a good idea!

rukai:
Thanks for feedback, should be all addressed now.

examples/capture/main.rs
```diff
@@ -27,20 +29,37 @@ async fn run() {
...
+            println!("*   2 arguments - uses specified width and height values");
+            return;
+        }
+    };
+
+    // It is a webgpu requirement that BufferCopyView.layout.bytes_per_row % 256 == 0
+    // So we calculate an appropriate padded_bytes_per_row that fits that requirement
+    let bytes_per_pixel = size_of::<u32>();
+    let real_bytes_per_row = width * bytes_per_pixel;
+    let padded_bytes_per_row_padding = (256 - (real_bytes_per_row) % 256) % 256;
```
kvark, line 49: Let's use the actual constant of `wgpu::COPY_BYTES_PER_ROW_ALIGNMENT`.

cwfitzgerald, line 49: The set of parentheses on `real_bytes_per_row` don't seem to be affecting order of operations in any way. Additionally is there any way to simplify/explain this expression as it's very unclear what is going on at first glance, largely in part to the double mod. I thought of `((real_bytes_per_row + 255) / 256) * 256` but that's not terribly clearer either.

examples/capture/main.rs
```diff
@@ -27,20 +29,37 @@ async fn run() {
...
+            println!("*   0 arguments - uses default width and height of (100, 200)");
+            println!("*   2 arguments - uses specified width and height values");
+            return;
+        }
+    };
+
+    // It is a webgpu requirement that BufferCopyView.layout.bytes_per_row % 256 == 0
+    // So we calculate an appropriate padded_bytes_per_row that fits that requirement
+    let bytes_per_pixel = size_of::<u32>();
+    let real_bytes_per_row = width * bytes_per_pixel;
```
cwfitzgerald, line 48: `unpadded_bytes_per_row` is probably a better name for this as `real` is one of those annoying words that could mean either the padded or the unpadded size.

examples/capture/main.rs
```diff
@@ -108,15 +127,24 @@ async fn run() {
...
             .write_header()
             .unwrap()
-            .write_image_data(data)
-            .unwrap();
+            .into_stream_writer_with_size(real_bytes_per_row);
+
+        // from the padded_buffer we write just the real bytes into the image
+        for y in 0..height as usize {
+            let start = y * padded_bytes_per_row;
+            let end = start + real_bytes_per_row;
```
cwfitzgerald, line 143: Not entirely necessary, but I would add a comment explaining the use of the unpadded bytes here.

examples/capture/main.rs
rukai: Oh nice!
I reexported it from wgpu-types.

examples/capture/main.rs
```diff
@@ -108,15 +128,24 @@ async fn run() {
...
-        png_encoder
+        let mut png_writer = png_encoder
             .write_header()
             .unwrap()
-            .write_image_data(data)
-            .unwrap();
+            .into_stream_writer_with_size(real_bytes_per_row);
+
+        // from the padded_buffer we write just the real bytes into the image
+        for y in 0..height as usize {
```
kvark, line 142: oh, also, we can go fancy mode:
```rust
for chunk in padded_buffer.chunks(padded_bytes_per_row) {
  png_writer.write(&chunk[..real_bytes_per_row]).unwrap();
}
```

examples/capture/main.rs
rukai: yeah ... I agree it is weird looking. I had to lookup how to do it efficiently after kvark told me `align - real_bytes_per_row % align` was inefficient when `real_bytes_per_row == 0`.

Would including a link to https://en.wikipedia.org/wiki/Data_structure_alignment#Computing_padding be sufficient?

examples/capture/main.rs
rukai: Not sure what you mean here. Does your comment still apply after the refactor?

examples/capture/main.rs
rukai: good idea

examples/capture/main.rs
rukai: Fancy mode engaged

examples/capture/main.rs
cwfitzgerald: Seems clear enough to me post refractor.

examples/capture/main.rs
cwfitzgerald: I would just add something like

```rust
// Round size up to the next multiple of COPY_BYTES_PER_ROW_ALIGNMENT
```

to make things clear even if they can't understand the math.

rukai: Thoughts?

PR #350, opened 2020-06-05
Implement Extensions Interface

cwfitzgerald:
This implements https://github.com/gfx-rs/wgpu/pull/703 in wgpu-rs. Notable changes include the removal of the anisotropic field from the examples, in favor of the now mandatory `..Default::default()` syntax.

PR #354, opened 2020-06-07
Update to wgpu with shader validation

kvark:
Depends on https://github.com/gfx-rs/wgpu/pull/705

### DIFF
--- a/examples/mipmap/main.rs
+++ b/examples/mipmap/main.rs
@@ -154,9 +154,6 @@ impl Example {
             mag_filter: wgpu::FilterMode::Linear,
             min_filter: wgpu::FilterMode::Nearest,
             mipmap_filter: wgpu::FilterMode::Nearest,
-            lod_min_clamp: 0.0,
-            lod_max_clamp: 100.0,
-            compare: None,
             ..Default::default()
         });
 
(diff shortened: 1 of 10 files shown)


examples/mipmap/main.rs
```diff
@@ -154,10 +154,7 @@ impl Example {
             mag_filter: wgpu::FilterMode::Linear,
             min_filter: wgpu::FilterMode::Nearest,
             mipmap_filter: wgpu::FilterMode::Nearest,
-            lod_min_clamp: 0.0,
-            lod_max_clamp: 100.0,
-            compare: None,
-            anisotropy_clamp: None,
+            .. wgpu::SamplerDescriptor::default()
```
cwfitzgerald, line 157: This might cause conflicts with #350 as that does this same thing, but less aggressively.

cwfitzgerald:
Now that #350 merged, this should be good to go

src/lib.rs
```diff
@@ -926,9 +926,12 @@ impl Instance {
     pub fn enumerate_adapters(&self, backends: wgt::BackendBit) -> impl Iterator<Item = Adapter> {
```
cwfitzgerald, line 926: This should take unsafe extensions as an argument, so the user can choose.

kvark:
Ready now

PR #356, opened 2020-06-09
Add generic ranges to buffer mapping api and make it safe.

lachlansneff:
[Rendered](https://charted.space/notes/wgpu-rs/wgpu/struct.Buffer.html)

The safety issues with the current api (being able to unmap while still holding a slice to mapped data) are fixed by having `get_mapped_range` and `get_mapped_range_mut` return `BufferView` and `BufferViewMut`, which notify the buffer that those ranges are no longer being used when they're dropped. `Buffer.unmap` asserts that the list of mapped ranges is empty, therefore it is safe.


examples/capture/main.rs
```diff
@@ -117,7 +117,7 @@ async fn run() {
     queue.submit(Some(command_buffer));
 
     // Note that we're not calling `.await` here.
-    let buffer_future = output_buffer.map_async(wgpu::MapMode::Read, 0, wgt::BufferSize::WHOLE);
+    let buffer_future = output_buffer.map_async(wgpu::MapMode::Read, ..);
```
kvark, line 120: didn't you want to have it on the `BufferSlice`? what made you to reconsider?

src/lib.rs
```diff
@@ -1190,20 +1209,80 @@ pub enum MapMode {
...
+        Bound::Included(&bound) => BufferSize(bound + 1 - offset),
+        Bound::Excluded(&bound) => BufferSize(bound - offset),
+        Bound::Unbounded => BufferSize::WHOLE,
+    };
+
+    (offset, size)
+}
+
+pub struct BufferView<'a> {
+    buffer: &'a Buffer,
```
kvark, line 1228: let's just include the existing type here and in `BufferViewMut`:
```rust
slice: BufferSlice<'a>,
```

src/lib.rs
```diff
@@ -1190,20 +1209,80 @@ pub enum MapMode {
...
+    fn deref(&self) -> &[u8] {
+        self.data
+    }
+}
+
+impl std::ops::Deref for BufferViewMut<'_> {
+    type Target = [u8];
+
+    fn deref(&self) -> &[u8] {
+        self.data
```
kvark, line 1253: Can we also have a flag in `BufferViewMut`, something like `readable: bool`?

This would allow us to add the following check on `deref()`:
```rust
assert!(self.readable, "Attempting to read a write-only mapping for buffer {:?}", self.slice.buffer.id);
```
Reading a write-only mapping may have unintended consequences: the implementation may have a write-combined memory type on native, reading from which is horribly slow. We don't want these surprises.

src/lib.rs
```diff
@@ -1190,20 +1209,80 @@ pub enum MapMode {
...
+
+impl std::ops::DerefMut for BufferViewMut<'_> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.data
+    }
+}
+
+impl Drop for BufferView<'_> {
+    fn drop(&mut self) {
+        self.buffer
```
kvark, line 1265: let's put this inside `if !thread::panicking() {}` check

examples/capture/main.rs
lachlansneff: Oh, yeah, we can do that. Some of my previous attempts didn't work on `BufferSlice`, so I forgot to put it there this time. Should `get_mapped_range` and `get_mapped_range_mut` be on `BufferSlice` as well? Also, should `BufferSlice` have a `slice` method?

lachlansneff: That being said, I'm a little worried that moving the methods over to `BufferSlice` could obscure their usage a little.

src/lib.rs
lachlansneff: Good idea. Didn't realize there was write-only memory.

src/lib.rs
lachlansneff: Okay, but what does this do?

examples/capture/main.rs
kvark: Since we can do this, and since the relative frame of reference for both `get_mapped_range*` and `map_async` is the same (beginning of the buffer), it makes sense to me that we'd have all of them on `BufferSlice`.

> Also, should BufferSlice have a slice method?

Hah, we could have that, yes, if you feel motivated to do this :)

kvark: > That being said, I'm a little worried that moving the methods over to BufferSlice could obscure their usage a little.

I share the concern... and I think it's fine, given that buffer mapping should be considered an advanced feature of the API. Most users are ok without mapping at all.

src/lib.rs
kvark: Strictly speaking, users can still read from `&mut []` that they get from `DerefMut`, but that one is much harder to prevent, so it's ok.

src/lib.rs
kvark: I suppose it's not actually needed here, so nevermind :)

To clarify, if a drop() method does something non-trivial, like calling into `wgpu`, we don't want that to happen if the drop is calling during stack unwinding of another panic. That would be double panic, and it's annoying to look at :)

src/lib.rs
lachlansneff: There may be some way of having a write-only slice with a new-type.

lachlansneff:
Not ready to merge yet. Still needs the `slice` function on `BufferSlice`.

src/lib.rs
```diff
@@ -1190,27 +1213,123 @@ pub enum MapMode {
...
 
+    /// Flushes any pending write operations and unmaps the buffer from host memory.
+    pub fn unmap(&self) {
+        self.map_context.lock().reset();
+        Context::buffer_unmap(&*self.context, &self.id);
+    }
+}
+
+impl BufferSlice<'_> {
+    pub fn slice<S: RangeBounds<BufferAddress>>(&self, bounds: S) -> BufferSlice {
```
kvark, line 1309: should probably be `-> Self` to avoid ambiguity about what lifetime this has

src/lib.rs
```diff
@@ -1190,27 +1213,123 @@ pub enum MapMode {
...
+        let (new_offset, new_size) = range_to_offset_size(bounds);
+
+        let new_size = if new_size == BufferSize::WHOLE {
+            self.size
```
kvark, line 1313: this doesn't look right, it should be `self.offset + self.size - new_offset`

src/lib.rs
```diff
@@ -1190,27 +1213,123 @@ pub enum MapMode {
...
+        } else {
+            assert!(new_size.0 <= self.size.0, "cannot slice with a larger size");
```
kvark, line 1315: we shouldn't compare size of a slice directly, but rather the end points

src/lib.rs
```diff
@@ -1190,27 +1213,123 @@ pub enum MapMode {
...
+            new_size
+        };
+
+        assert!(
+            new_offset <= new_size.0,
+            "offset cannot be larger than size"
```
kvark, line 1321: this seems wrong?

PR #360, opened 2020-06-11
Implement TEXTURE_BINDING_ARRAY extension

cwfitzgerald:
This extends https://github.com/gfx-rs/wgpu/pull/711 into wgpu-rs.

Notable changes: 
- Added an example showing off both this extension and the future descriptor indexing extension.
- Changed the framework so there is a static function showing what extensions you need. This is provided by the trait,, which defaults to no extensions, so existing examples and new ones don't have to care about extensions.


examples/texture-arrays/main.rs
```diff
@@ -0,0 +1,323 @@
...
+    GREEN,
+}
+
+fn create_texture_data(color: Color) -> Vec<u8> {
+    let mut data: Vec<u8> = match color {
+        Color::RED => vec![255, 0, 0, 255],
+        Color::GREEN => vec![0, 255, 0, 255],
+    };
+
+    // Pad data out to require bytes per row alignment
```
kvark, line 63: This isn't needed at all for `write_texture` - the benefit of a built-in function. To clarify, since `write_data` does the copy into the staging area, it takes responsibility on aligning the rows properly for you.

examples/texture-arrays/main.rs
```diff
@@ -0,0 +1,323 @@
...
+            }],
+            depth_stencil_attachment: None,
+        });
+        rpass.set_pipeline(&self.pipeline);
+        rpass.set_bind_group(0, &self.bind_group, &[]);
+        rpass.set_vertex_buffer(0, self.vertex_buffer.slice(..));
+        rpass.set_index_buffer(self.index_buffer.slice(..));
+        rpass.draw_indexed(0..12, 0, 0..1);
+
+        drop(rpass);
```
kvark, line 315: I guess Rust can do this automatically for us, does it?

src/backend/direct.rs
```diff
@@ -291,64 +292,9 @@ impl crate::Context for Context {
         device: &Self::DeviceId,
         desc: &BindGroupLayoutDescriptor,
     ) -> Self::BindGroupLayoutId {
-        use wgc::binding_model as bm;
-
-        let temp_layouts = desc
```
kvark, line 296: nicely removed! 🎉

src/backend/direct.rs
```diff
@@ -361,13 +307,14 @@ impl crate::Context for Context {
...
 
+        let texture_view_arena: Arena<wgc::id::TextureViewId> = Arena::new();
         let bindings = desc
             .bindings
             .iter()
             .map(|binding| bm::BindGroupEntry {
                 binding: binding.binding,
-                resource: match &binding.resource {
-                    BindingResource::Buffer(buffer_slice) => {
+                resource: match binding.resource {
```
kvark, line 316: nice! I was just noticing this myself, happy you are fixing this :)

examples/texture-arrays/main.rs
cwfitzgerald: What should the bytes_per_row be set to for write_texture, the actual value (in this case 4)?

examples/framework.rs
```diff
@@ -60,23 +63,27 @@ async fn run_async<E: Example>(event_loop: EventLoop<()>, window: Window) {
...
         .unwrap();
 
+    let adapter_extensions = adapter.extensions();
+
     let trace_dir = std::env::var("WGPU_TRACE");
     let (device, queue) = adapter
         .request_device(
             &wgpu::DeviceDescriptor {
-                extensions: wgpu::Extensions::empty(),
+                extensions: adapter_extensions & needed_extensions,
```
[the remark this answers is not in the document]
cwfitzgerald: The reason I chose to do this is for examples with optional extensions. We don't want the framework to error out, we want the example to choose what it should do with the extensions that it got.

examples/texture-arrays/main.rs
cwfitzgerald: It could if I put it in a block, but I much prefer the style where places where we need a drop are explicitly annotated to show "this object definitively ends here". Up to choice, but I like this style.

examples/framework.rs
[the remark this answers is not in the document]
kvark: ah I see, ok we can leave this `&` like this

examples/texture-arrays/main.rs
kvark: yes, the actual value of 4

examples/texture-arrays/main.rs
kvark: no, I mean, even if you don't put it into a block and just remove the `drop`, will the compiler figure out that it has to drop the pass (since otherwise the `encoder` is borrowed and can't issue the `finish()` call)?

cwfitzgerald: Nope (just tried it). Lifetime data never influences where drops happen, so if something impls drop, it's going to have a nice `&mut self` at the end of the function, which ofc wouldn't work here.

cwfitzgerald:
Should be good from here :)

examples/texture-arrays/main.rs
```diff
@@ -0,0 +1,318 @@
...
+
+#[derive(Copy, Clone)]
+enum Color {
+    RED,
+    GREEN,
+}
+
+fn create_texture_data(color: Color) -> Vec<u8> {
+    match color {
+        Color::RED => vec![255, 0, 0, 255],
```
kvark, line 59: nit: could just return the fixed-size `[u8; 4]` without a `Vec`

cwfitzgerald:
Picked!
bors r=kvark

PR #366, opened 2020-06-11
Fix buffer mapping examples

### LINKED ISSUE #364: Panic on running capture and hello-compute examples
Hi, 

I'm wondering whether this might be something on my end. Running the mentioned examples  on b1f3fba results in a `panic` on my system. 

I'm currently running Wayland on Arch Linux 5.6.15 with `sway` as a window manager.

```shell
$ cargo run --example capture
$ cargo run --example hello-compute 1 2 3 4
```
Both result in the following `panic` message:
```shell
thread 'main' panicked at 'You cannot unmap a buffer that still has accessible mapped views', src/lib.rs:379:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'unable to remove range from map context', src/lib.rs:411:21
```

Commenting out `output_buffer.unmap()` in the [`capture example`](https://github.com/gfx-rs/wgpu-rs/blob/b1f3fbaab0ae18fb97f9126d6994605731a35311/examples/capture/main.rs#L156) allows it to not crash.

### LINKED ISSUE #365: WrongType when sending instance vertex attribute data to the vertex shader
When sending Mat4 data as instance vertex attributes to the vertex shader, I have to de-compose the Mat4 colums in the ```VertexBufferDescriptor``` into it's four Vec4:

```rust
        wgpu::VertexBufferDescriptor {
            stride: 16 * 4,
            step_mode: wgpu::InputStepMode::Instance,
            // 4: mat4 MVP matrix
            attributes: &vertex_attr_array![4 => Float4, 5 => Float4, 6 => Float4, 7 => Float4],
        }
```

I normaly would assume that I could then use the Mat4 data inside the vertex shader like this (GLSL):
```glsl 
layout(location=4) in vec4 a_mvp;
```

This will currently (b1f3fba) result in the following error:
```rust
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Stage { flag: VERTEX, error: Input { location: 4, error: WrongType } }', C:\Users\Username\.cargo\git\checkouts\wgpu-rs-40ea39809c03c5d8\7b866b1\src\backend\direct.rs:460:9
```

As a workaround I can use the following shader definitions and composing the Mat4 myself:

```glsl
layout(location=4) in vec4 a_mvp_c0;
layout(location=5) in vec4 a_mvp_c1;
layout(location=6) in vec4 a_mvp_c2;
layout(location=7) in vec4 a_mvp_c3;

void main() {
    mat4 mvp = mat4(a_mvp_c0, a_mvp_c1, a_mvp_c2, a_mvp_c3);
}
```

Is this the intended behavior or might this be a bug?

Edit: Tested under Windows 10 with the Vulkan backend.
Edit2: And DX12.

lachlansneff:
Fixes #364

### DIFF
--- a/examples/capture/main.rs
+++ b/examples/capture/main.rs
@@ -153,6 +153,10 @@ async fn run() {
         }
         png_writer.finish().unwrap();
 
+        // With the current interface, we have to make sure all mapped views are
+        // dropped before we unmap the buffer.
+        drop(padded_buffer);
+
         output_buffer.unmap();
     }
 }
--- a/examples/hello-compute/main.rs
+++ b/examples/hello-compute/main.rs
@@ -123,7 +123,12 @@ async fn execute_gpu(numbers: Vec<u32>) -> Vec<u32> {
             .chunks_exact(4)
             .map(|b| u32::from_ne_bytes(b.try_into().unwrap()))
             .collect();
+
+        // With the current interface, we have to make sure all mapped views are
+        // dropped before we unmap the buffer.
+        drop(data);
         staging_buffer.unmap();
+
         result
     } else {
         panic!("failed to run compute on gpu!")


lachlansneff:
> I would probably prefer to see the drop included to show you have to drop the mapping before unmapping it, maybe with a comment on it.

That's reasonable. I think the ultimate fix for this is to improve the panic messages. I'll make a pr later that does so.

cwfitzgerald:
Thanks!
bors r+

PR #369, opened 2020-06-12
Implement SAMPLED_TEXTURE_ARRAY_NON_UNIFORM_INDEXING

cwfitzgerald:
This implements https://github.com/gfx-rs/wgpu/pull/715 in wgpu-rs. I haven't changed the example, as I want to actually think up a better example to use and didn't want that to block this. It will change in the future however.


examples/texture-arrays/shader.frag
```diff
@@ -8,11 +10,5 @@ layout(set = 0, binding = 0) uniform texture2D u_Textures[2];
...
 
 void main() {
-    if (v_Index == 0) {
-        o_Color = vec4(texture(sampler2D(u_Textures[0], u_Sampler), v_TexCoord).rgb, 1.0);
-    } else if (v_Index == 1) {
-        o_Color = vec4(texture(sampler2D(u_Textures[1], u_Sampler), v_TexCoord).rgb, 1.0);
-    } else {
-        o_Color = vec4(0.0, 0.0, 1.0, 1.0);
-    }
+    o_Color = vec4(texture(sampler2D(u_Textures[v_Index], u_Sampler), v_TexCoord).rgb, 1.0);
```
kvark, line 13: How do you feel about making this example to support both uniform and non-uniform indexing?
That could be a good demonstration of how to support optional extensions. We'd just have 2 different shaders here, one picked at initialization.

cwfitzgerald: Sure

examples/texture-arrays/constant.frag
```diff
@@ -1,7 +1,7 @@
 #version 450
 
 layout(location = 0) in vec2 v_TexCoord;
-layout(location = 1) flat in int v_Index;
+layout(location = 1) flat in int v_Index;  // dynamically non-uniform
```
kvark, line 4: why is this dynamically non-uniform?

examples/texture-arrays/non-uniform.frag
```diff
@@ -0,0 +1,14 @@
+#version 450
+
+#extension GL_EXT_nonuniform_qualifier : require
+
+layout(location = 0) in vec2 v_TexCoord;
+layout(location = 1) flat in int v_Index;  // dynamically non-uniform
```
kvark, line 6: This is actually uniform across the wave, since it's flat-interpolated. Is it supposed to be non-uniform in a difference sense?

examples/texture-arrays/constant.frag
cwfitzgerald: My intuition was that "dynamically uniform" meant that it was the same across the entire drawcall which is confirmed by a careful reading of the glsl spec ([pg. 24](https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.40.pdf)) which says (emphasis mine):

> A fragment-shader expression is dynamically uniform if **all fragments** [note - not all fragments in the primative] evaluating it get the same resulting value.

By this rule, something flat shaded over the primative isn't dynamically uniform.

examples/texture-arrays/non-uniform.frag
cwfitzgerald: If you consider how gpus run fragment shaders, it may not be even coherent in the wave. They take 2x2 blocks of pixels, put N of them together into a wave. These blocks may not be from the same primative, so this may be different across the wave. This 2x2 action is also why gpus are bad at rendering small polygons, they always get a 2x2 section, even if they are only 1x1.

examples/texture-arrays/constant.frag
kvark: I'll need to look at this deeper. The same section you link to also mentions this:

> This is similarly defined for other shader stages, based on the per-instance data they process.

If VS shader is dynamically uniform on per-instance data, that means it's not uniform across the whole draw call like you suggested... It's more suggestive that "dynamically uniform" means uniform across the wave, and each primitive is scheduled in separate waves, and each VS instance is scheduled across different waves. That would match the spec fragment.

examples/texture-arrays/non-uniform.frag
kvark: My understanding is that 2x2 blocks are *always* from the same primitive. If some of these pixels lie outside, they become "ghost" threads, but regardless - `flat` is there precisely to inform the compiler that this is constant across the wave.

The link you provided is very familiar to me, although I'll make sure to rehash that part in my mind.

examples/texture-arrays/constant.frag
cwfitzgerald: As a more concrete piece of proof, glslc won't let you index a descriptor array with this value without GL_EXT_nonuniform qualifier and spirv descriptor indexing extensions on.

examples/texture-arrays/non-uniform.frag
cwfitzgerald: The pixels from within the 2x2 blocks are always from the same primitive, being masked out and such as needed, but when 8 of those are combined into a wave, each of those 8 blocks could be from a different primitive. If that wasn't allowed, having a 1x1 triangle would waste an entire wave of processing power, not just a 2x2 block.

examples/texture-arrays/uniform.frag
```diff
@@ -0,0 +1,15 @@
+#version 450
+
+layout(location = 0) in vec2 v_TexCoord;
+// layout(location = 1) flat in int v_Index;  // dynamically non-uniform
+layout(location = 0) out vec4 o_Color;
+
+layout(set = 0, binding = 0) uniform texture2D u_Textures[2];
+layout(set = 0, binding = 1) uniform sampler u_Sampler;
+layout(set = 1, binding = 0) uniform Uniforms {
+    int u_Index;  // dynamically uniform
```
kvark, line 10: wait, is this not "unsized" as well?

cwfitzgerald: Sized vs unsized is about if `uniform texture2D u_Textures[2];` has a compile time array size or not.

cwfitzgerald:
Alright, I think this is finally ready for review on this one

PR #370, opened 2020-06-13
API update for RODS2

kvark:
Reflects upstream changes in https://github.com/gfx-rs/wgpu/pull/716
I'm not too happy about the API just yet, has another idea that I'm going to discuss with the group on Monday, hopefully. But we need to get going and not block other wgpu-rs changes, so this should land in some way.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,20 +28,20 @@ vulkan = ["wgc/gfx-backend-vulkan"]
 package = "wgpu-core"
 version = "0.5"
 git = "https://github.com/gfx-rs/wgpu"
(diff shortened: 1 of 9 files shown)


examples/framework.rs
```diff
@@ -182,6 +182,7 @@ async fn run_async<E: Example>(event_loop: EventLoop<()>, window: Window) {
 
                 let command_buf = example.render(&frame.output, &device, &queue);
                 queue.submit(Some(command_buf));
+                //*control_flow = ControlFlow::Exit;
```
cwfitzgerald, line 185: Commented?

kvark: indeed!

kvark:
bors r=cwfitzgerald

PR #373, opened 2020-06-14
add test case for empty buffer

rukai:
Depends on https://github.com/gfx-rs/wgpu/pull/720


rukai:
(push spam was because I carelessly updated the commit which broke things due to https://github.com/gfx-rs/wgpu/commit/5a7fb4c1e57408c308ba058c90e28c8598c86bbb)
Can we land this as is?

@cwfitzgerald can you make a separate PR for wgpu-rs to use the latest wgpu commit?

PR #374, opened 2020-06-14
Add debug markers and debug markers example

krupitskas:
Depends on https://github.com/gfx-rs/wgpu/pull/719

### DIFF
--- a/examples/cube/main.rs
+++ b/examples/cube/main.rs
@@ -345,10 +345,13 @@ impl framework::Example for Example {
                 }],
                 depth_stencil_attachment: None,
             });
+            rpass.push_debug_group("Prepare data for draw.");
             rpass.set_pipeline(&self.pipeline);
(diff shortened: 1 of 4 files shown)


kvark:
Thank you!
Let's not add a dedicated example here. It's trivial to just add markers to an existing example, instead.

krupitskas:
I afraid that this CI can't be completed successfully because changed from wgpu crate isn't in master yet.

kvark:
Yep, I edited the description accordingly.

kvark:
@krupitskas could you update this please?

krupitskas:
> 
> 
> @krupitskas could you update this please?

What exactly? About build status I think you can trigger it manually? Or you mean codebase?

kvark:
Actually, you are right, you don't need to change the dependency pointer, because it was done by other recent PRs. We just need to merge this :)
bors r+

Issue #329, opened 2020-05-28
Example for water rendering


kvark:
It would be wonderful to have an example that renders water, to feature the following use cases:
  - separation of rendering into opaque and transparent passes
  - depth testing and writing
  - blending
  - read-only depth/stencil attachments - requires https://github.com/gfx-rs/wgpu/issues/680
  - procedural geometry (possibly, with dynamic updates?)
  - fancy shaders with reflection and refraction

OptimisticPeach:
I'm rewriting a project I was working on with wgpu. It had water, however it was low-poly water:
![Screenshot_20200423-194538|690x388](https://aws1.discourse-cdn.com/business5/uploads/rust_lang/optimized/2X/9/9e22cefbc2585b7d69947b2fb0e8606980091640_2_690x388.png)

A few notes though:
- I didn't do refraction, and I didn't perfect reflection (I used the "flat" mesh's reflection, and just made it a vertex attribute, so it would look like the water's movement would distort the reflection).
- By blending do you mean alpha blending, thereby making the water transparent? I originally didn't make the water transparent, and instead rendered the below-water objects to a separate framebuffer and just used that as a texture.
- I didn't use a depth attachment.
- The procedural geometry was done on the GPU. Since it's low-poly, it would be more data to update on the CPU-side of things (because you have to duplicate all your vertices), so I instead optimized the data delivery method, and update the vertices in the vertex shader. 

If I have a working implementation before someone else has posted, would it be okay for me to upload this as an example?

OptimisticPeach:
I'm working on porting water from an old project of mine to wgpu, and would love to use this to gain experience. For reference, here is a screenshot of the water:
[...]

PR #375, opened 2020-06-15
Add water example. 

OptimisticPeach:
Solves #329, water example requested by @kvark.

I tuned it to my personal preference of visuals, however it might be different for you. 

Note: I used https://github.com/ashima/webgl-noise for 3D open simplex noise.

I've commented to explain what most of the things in the rust side of the example mean. However, I'm not 100% sure I did the best job at giving a brief overview, and wouldn't mind someone making sure my terminology/definitions are correct. 

Thanks!
Patrik


examples/water/main.rs
```diff
@@ -0,0 +1,797 @@
...
+
+        // Will have a render from under the water.
+        let reflection_texture = device.create_texture(&wgpu::TextureDescriptor {
+            label: None,
+            size: texture_extent,
+            mip_level_count: 1,
+            sample_count: 1,
+            dimension: wgpu::TextureDimension::D2,
+            format: sc_desc.format,
+            usage: wgpu::TextureUsage::SAMPLED | wgpu::TextureUsage::COPY_DST | wgpu::TextureUsage::OUTPUT_ATTACHMENT,
```
kvark, line 248: why does this and the other one need `COPY_DST` usage?

examples/water/main.rs
```diff
@@ -0,0 +1,797 @@
...
+                    operation: wgpu::BlendOperation::Add,
+                },
+                alpha_blend: wgpu::BlendDescriptor {
+                    src_factor: wgpu::BlendFactor::One,
+                    dst_factor: wgpu::BlendFactor::One,
+                    operation: wgpu::BlendOperation::Max,
+                },
+                write_mask: wgpu::ColorWrite::ALL,
+            }],
+            // Describes how us writing to the depth/stencil buffer
```
kvark, line 437: thank you for commenting all the interesting bits!

examples/water/main.rs
```diff
@@ -0,0 +1,797 @@
...
+            vertex_state: wgpu::VertexStateDescriptor {
+                // We don't actually use indices, since it's unnecessary
+                // because we duplicate all the data anyway. This is
+                // necessary to achieve the low-poly effect.
+                index_format: wgpu::IndexFormat::Uint16,
+                vertex_buffers: &[
+                    wgpu::VertexBufferDescriptor {
+                        stride: water_vertex_size as wgpu::BufferAddress,
+                        step_mode: wgpu::InputStepMode::Vertex,
+                        attributes: &[
```
kvark, line 467: did you consider using the `vertex_attr_array` macro like some of the other examples?

examples/water/main.rs
```diff
@@ -0,0 +1,797 @@
...
+        sc_desc: &wgpu::SwapChainDescriptor,
+        device: &wgpu::Device,
+        queue: &wgpu::Queue,
+    ) {
+        if sc_desc.width == 0 && sc_desc.height == 0 {
+            // Stop rendering altogether.
+            self.active = std::usize::MAX;
+            return;
+        } else {
+            // Skip a frame, since it crashes otherwise.
```
kvark, line 596: what is this crash about?

examples/water/main.rs
```diff
@@ -0,0 +1,797 @@
...
+        self.draw_depth_buffer = draw_depth_buffer.create_default_view();
+        self.reflect_view = reflection_texture.create_default_view();
+
+        let (
+            normal_view,
+            flipped_view,
+            projection,
+        ) = Self::generate_matrices(sc_desc.width as f32 / sc_desc.height as f32);
+
+        let normal_view: &[f32; 16] = normal_view.as_ref();
```
kvark, line 637: let's put all of this initialization code in a static method that we can re-use on both initialization and resize

examples/water/main.rs
```diff
@@ -0,0 +1,797 @@
...
+        let (water_sin, water_cos) = ((self.current_frame as f32) / 400.0).sin_cos();
+        queue.write_buffer(&self.water_uniform_buf, std::mem::size_of::<[f32; 16]>() as wgpu::BufferAddress * 2, bytemuck::cast_slice(&[water_sin, water_cos]));
+
+        // The encoder provides a way to turn our instructions here, into
+        // a command buffer the GPU can understand.
+        let mut encoder =
+            device.create_command_encoder(&wgpu::CommandEncoderDescriptor { label: None });
+
+        // Only render valid frames. See resize method.
+        if self.active < self.current_frame {
```
kvark, line 681: it doesn't explain what the crash is
also, consider doing an early out here: all the encoding code becomes indented otherwise

examples/water/water_shader.frag
```diff
@@ -0,0 +1,41 @@
...
+    float z_n = 2.0 * depth - 1.0;
+    float z_e = 2.0 * zNear * zFar / (zFar + zNear - z_n * (zFar - zNear));
+    return z_e;
+}
+
+void main() {
+    vec3 reflection_colour = texture(sampler2D(reflection, colour_sampler), f_In.f_WaterScreenPos.xy).xyz;
+
+    float my_depth = to_linear_depth(gl_FragCoord.z);
+    float terrain_depth = to_linear_depth(texture(sampler2D(terrain_depth_tex, colour_sampler), gl_FragCoord.xy / vec2(1024.0, 768.0)).r);
```
kvark, line 29: so this 1024x768 will become wrong as soon as we resize
can we pass it as a uniform constant?

examples/water/main.rs
OptimisticPeach: The `sc_desc` passed in the frame we undo minimization has the correct size (1024 x 768), however the next call to render has frames whose sizes are 0 x 0:
```rust
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Extent { width: 1024, height: 768, depth: 1 }`,
 right: `Extent { width: 0, height: 0, depth: 1 }`: Extent state must match extent from view'
```
Skipping a frame seemed like a valid way of circumventing this.

kvark: ok, I see. I suppose it's related to the resizing bugs of examples we have open :sweat_smile:

kvark:
Also, we still need to change the shaders to not require this:
>     VALIDATION [UNASSIGNED-CoreValidation-Shader-FeatureNotEnabled (-1940616195)] : Validation Error: [ UNASSIGNED-CoreValidation-Shader-FeatureNotEnabled ] Object 0: handle = 0x556668263ea0, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x8c548bfd | Shader requires VkPhysicalDeviceFeatures::shaderClipDistance but is not enabled on the device
    object info: (type: DEVICE, hndl: 93898322362016)

OptimisticPeach:
@kvark, I've removed the usage of user-defined clipping planes, however I believe only certain, older mobile hardware doesn't support it, so it might be more logical to enable it by default.

kvark:
>  I've removed the usage of user-defined clipping planes, however I believe only certain, older mobile hardware doesn't support it, so it might be more logical to enable it by default.

It's not just a matter of us requesting this feature. It's also a matter of WGSL (the official WebGPU shading language) to figure out if it even has this feature. So it's safer to avoid it for now, or otherwise we'll not be able to compile this example to the Web.

OptimisticPeach:
It seems I made a mistake when squashing... why does it say that I coauthored all those other commits?

cwfitzgerald:
I want to take a brief look at this before we merge. Not sure what caused the git issues

kvark:
@OptimisticPeach try `git reflog` to locate the old revision, then squash properly again.

examples/water/main.rs
```diff
@@ -0,0 +1,770 @@
...
+            bind_group_layouts: &[&water_bind_group_layout],
+        });
+
+        let terrain_pipeline_layout = device.create_pipeline_layout(&wgpu::PipelineLayoutDescriptor {
+            bind_group_layouts: &[&terrain_bind_group_layout],
+        });
+
+        let water_uniform_buf = device
+            .create_buffer_with_data(
+                bytemuck::cast_slice(&[0.0f32; 40]),
```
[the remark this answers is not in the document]
OptimisticPeach: Please check this next time I commit. Not sure if `mapped_at_creation` is supposed to be `false` in this case.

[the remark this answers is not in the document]
cwfitzgerald: If you're using write_buffer, yes, this is supposed to be false.

cwfitzgerald:
bors r=kvark,cwfitzgerald

PR #377, opened 2020-06-16
Update with minBufferBindingSize

kvark:
Depends on https://github.com/gfx-rs/wgpu/pull/726
Also reverts #373 : buffer bindings now have to include at least one element of an unsized struct portion, so they can't be zero-sized.


kvark:
bors r=cwfitzgerald

kvark:
Updated the water example and fixed it to work on Metal.
bors r=cwfitzgerald

kvark:
web backend fixes
bors r=cwfitzgerald

cwfitzgerald:
Third times the charm :)

PR #380, opened 2020-06-16
Documentation Pass

### LINKED ISSUE #378: Currently Outstanding Documentation Issues
### `Surface::create`

This still suggests `Surface::create`, but it's removed by https://github.com/gfx-rs/wgpu-rs/commit/ba953547346ec230726c03e35ee0e6f977742f95, IIUC.

https://github.com/gfx-rs/wgpu-rs/blob/7b887f12a35c26a21f56f6540b1b771177dabd6a/src/lib.rs#L471-L472

### The descriptions about `Instance` and `Adapter`

They use the same phrases, so either of them is not the corresponding one?

https://github.com/gfx-rs/wgpu-rs/blob/092d5bb6fdf765ff4c80344a4f3de1b49a0879ec/src/lib.rs#L314-L329

cwfitzgerald:
The wgpu-rs component of https://github.com/gfx-rs/wgpu/pull/728.

Closes #378.

Notable changes not mentioned in wgpu PR:
- Reexported the wgt-local types whenever we have a type definition. These are reexported as `ThingBase`. This is unfortunately the only way to actually get their documentation to show up, aliases have no way of "inheriting" the documentation.
- Added wgpu logo to docs!
- Hid vertex_format_size macro.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,7 @@
 //! A cross-platform graphics and compute library based on WebGPU.
 
+#![doc(html_logo_url = "https://raw.githubusercontent.com/gfx-rs/wgpu-rs/master/logo.png")]
+
 mod backend;
 pub mod util;
 #[macro_use]
@@ -29,7 +31,7 @@ pub use wgt::{
     StencilOperation, StencilStateFaceDescriptor, StoreOp, SwapChainDescriptor, SwapChainStatus,
     TextureAspect, TextureComponentType, TextureDataLayout, TextureDimension, TextureFormat,
     TextureUsage, TextureViewDimension, UnsafeExtensions, VertexAttributeDescriptor, VertexFormat,
-    BIND_BUFFER_ALIGNMENT, COPY_BYTES_PER_ROW_ALIGNMENT,
+    BIND_BUFFER_ALIGNMENT, COPY_BUFFER_ALIGNMENT, COPY_BYTES_PER_ROW_ALIGNMENT,
 };
 
 use backend::Context as C;
@@ -340,18 +342,21 @@ trait Context: Sized {
     );
 }
 
+/// Instance of wgpu. First thing you create when using wgpu.
+///
 /// An instance sets up the context for all other wgpu objects.
 ///
-/// An `Adapter` can be used to open a connection to the corresponding device on the host system,
-/// yielding a [`Device`] object.
+/// Instances and adapters do not have to be kept alive, only devices.
 pub struct Instance {
     context: Arc<C>,
 }
 
-/// A handle to a physical graphics and/or compute device.
+/// Handle to a physical graphics and/or compute device.
 ///
 /// An `Adapter` can be used to open a connection to the corresponding device on the host system,
 /// yielding a [`Device`] object.
+///
+/// Instances and adapters do not have to be kept alive, only devices.
 pub struct Adapter {
     context: Arc<C>,
     id: <C as Context>::AdapterId,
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -600,7 +617,10 @@ impl Drop for RenderPipeline {
     }
 }
 
-/// A handle to a compute pipeline.
+/// Handle to a compute pipeline.
+///
+/// A `ComputePipeline` object represents a compute pipeline and it's single shader stage.
```
lachlansneff, line 622: *its is the possessive form of it.

src/lib.rs
```diff
@@ -1467,6 +1587,12 @@ impl CommandEncoder {
     }
 
     /// Copy data from a buffer to a texture.
+    ///
+    /// # Panics
+    ///
+    /// - Copy would overrun buffer.
+    /// - Copy would overrun texture.
+    /// - source.layout.rows_per_image isn't divisible by 256.
```
Aeledfyr, line 1595: Wouldn't this be `source.layout.bytes_per_row` rather than `source.layout.rows_per_image`?

cwfitzgerald: Yup! Messed this up at least once before...

src/lib.rs
```diff
@@ -1671,7 +1813,10 @@ impl<'a> Drop for RenderPass<'a> {
 }
 
 impl<'a> ComputePass<'a> {
-    /// Sets the active bind group for a given bind group index.
+    /// Sets the active bind group for a given bind group index. The bind group layout
```
kvark, line 1816: we should change that a bit. The active pipeline doesn't matter at *this* step, it only matters at draw() time.

src/macros.rs
```diff
@@ -32,6 +32,7 @@ macro_rules! vertex_attr_array {
 
 // For internal usage
 #[macro_export]
+#[doc(hidden)]
```
kvark, line 35: do we have to hide that? could be useful to people

src/lib.rs
```diff
@@ -359,24 +364,28 @@ pub struct Adapter {
...
+///
+/// A device may be requested from an adapter with [`Adapter::request_device`].
 pub struct Device {
     context: Arc<C>,
     id: <C as Context>::DeviceId,
 }
 
-/// This is passed to `Device::poll` to control whether
-/// it should block or not.
+/// Passed to `Device::poll` to control if it should block or not.
```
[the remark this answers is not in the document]
cwfitzgerald: Which part of this isn't web compatible?

src/lib.rs
```diff
@@ -816,12 +873,18 @@ pub struct ComputePipelineDescriptor<'a> {
     pub compute_stage: ProgrammableStageDescriptor<'a>,
 }
 
+// The underlying types are also exported so that documentation shows up for them
+
+pub use wgt::RenderPassColorAttachmentDescriptorBase;
```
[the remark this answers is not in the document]
cwfitzgerald: Because they are special and are rename things for the purpose of aliasing, they almost feel like they should be local to the alias.

src/lib.rs
```diff
@@ -830,37 +893,45 @@ pub struct RenderPassDescriptor<'a, 'b> {
     pub depth_stencil_attachment: Option<RenderPassDepthStencilAttachmentDescriptor<'a>>,
 }
 
-/// A description of a buffer.
-pub type BufferDescriptor<'a> = wgt::BufferDescriptor<Option<&'a str>>;
+// The underlying types are also exported so that documentation shows up for them
+
+pub use wgt::BufferDescriptor as BufferDescriptorBase;
```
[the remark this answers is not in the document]
cwfitzgerald: The problem is that if you build just wgpu's documentation, you don't get a link to the wgt type as those docs weren't ever built. Additionally, even when all docs are built (such as on docs.rs), I feel like we shouldn't be linking to wgpu-types' doc page because that would break doc links for these types, and people will likely get confused as to where all their types went.

src/lib.rs
```diff
@@ -1345,6 +1456,8 @@ impl BufferSlice<'_> {
         )
     }
 
+    /// Synchronously and immediately map a buffer for reading. If the buffer is not immediately mappable
+    /// through [`BufferDescriptor::mapped_at_creation`] or [`BufferSlice::map_async`], will panic.
     pub fn get_mapped_range(&self) -> BufferView {
```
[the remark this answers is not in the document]
cwfitzgerald: agree

src/macros.rs
cwfitzgerald: Nah, I just thought it wasn't intended to be used due to the comment. If we want to keep it, I'll give it some simple docs.

src/lib.rs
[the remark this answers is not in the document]
kvark: I thought the whole `poll` part is not compatible. On the Web we ask the event loop to proceed, and we can't busy wait for it, be cause we are executed within the event loop. All we can do is to return from the event (i.e. render event) and let it do its thing.

cwfitzgerald:
Thank you!
bors r=kvark

PR #382, opened 2020-06-17
Remove interface blocks from the water shaders

kvark:
Fixes https://github.com/gfx-rs/gfx/issues/3274


kvark:
Confirmed to fix DX12 🎉 
bors r+

PR #384, opened 2020-06-17
Add include_spirv macro

kvark:
Depends on https://github.com/gfx-rs/wgpu/pull/729
Makes out shader loading in the examples a bit tidier.

### DIFF
--- a/examples/texture-arrays/main.rs
+++ b/examples/texture-arrays/main.rs
@@ -97,29 +97,26 @@ impl framework::Example for Example {
         queue: &wgpu::Queue,
     ) -> (Self, Option<wgpu::CommandBuffer>) {
         let mut uniform_workaround = false;
-        let vs_bytes: &[u8] = include_bytes!("shader.vert.spv");
-        let fs_bytes: &[u8] = match device.capabilities() {
+        let vs_module = device.create_shader_module(wgpu::include_spirv!("shader.vert.spv"));
+        let fs_bytes: Vec<u8> = match device.capabilities() {
             c if c.contains(wgpu::Capabilities::UNSIZED_BINDING_ARRAY) => {
-                include_bytes!("unsized-non-uniform.frag.spv")
+                include_bytes!("unsized-non-uniform.frag.spv").to_vec()
             }
             c if c.contains(wgpu::Capabilities::SAMPLED_TEXTURE_ARRAY_NON_UNIFORM_INDEXING) => {
-                include_bytes!("non-uniform.frag.spv")
+                include_bytes!("non-uniform.frag.spv").to_vec()
             }
             c if c.contains(wgpu::Capabilities::SAMPLED_TEXTURE_ARRAY_DYNAMIC_INDEXING) => {
                 uniform_workaround = true;
-                include_bytes!("uniform.frag.spv")
+                include_bytes!("uniform.frag.spv").to_vec()
             }
             c if c.contains(wgpu::Capabilities::SAMPLED_TEXTURE_BINDING_ARRAY) => {
-                include_bytes!("constant.frag.spv")
+                include_bytes!("constant.frag.spv").to_vec()
             }
             _ => {
                 panic!("Graphics adapter does not support any of the capabilities needed for this example");
             }
         };
-        let vs_module =
-            device.create_shader_module(&wgpu::read_spirv(std::io::Cursor::new(vs_bytes)).unwrap());
-        let fs_module =
-            device.create_shader_module(&wgpu::read_spirv(std::io::Cursor::new(fs_bytes)).unwrap());
+        let fs_module = device.create_shader_module(wgpu::util::make_spirv(&fs_bytes));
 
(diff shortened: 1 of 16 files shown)


src/util/mod.rs
```diff
@@ -1,41 +1,15 @@
...
-            io::ErrorKind::InvalidData,
-            "input missing SPIR-V magic number",
-        ));
-    }
-    Ok(result)
+
+    let (pre, words, post) = unsafe { data.align_to::<u32>() };
+    assert_eq!(pre, &[], "data offset is not aligned to words!");
+    assert_eq!(post, &[], "data size is not aligned to words!");
+    assert_eq!(words[0], MAGIC_NUMBER, "wrong magic word {:x}!", words[0]);
```
rukai, line 13: Can we extend this to: "wrong magic word {:x}. Make sure you are using a SPIRV file."

src/util/mod.rs
```diff
@@ -1,41 +1,15 @@
-use std::{io, slice};
+#[repr(C)] // guarantee 'bytes' comes after '_align'
+pub struct WorldAligned<Bytes: ?Sized> {
```
rukai, line 2: This should be WordAligned

examples/texture-arrays/main.rs
```diff
@@ -97,29 +97,26 @@ impl framework::Example for Example {
...
             }
             _ => {
                 panic!("Graphics adapter does not support any of the capabilities needed for this example");
             }
         };
-        let vs_module =
-            device.create_shader_module(&wgpu::read_spirv(std::io::Cursor::new(vs_bytes)).unwrap());
-        let fs_module =
-            device.create_shader_module(&wgpu::read_spirv(std::io::Cursor::new(fs_bytes)).unwrap());
+        let fs_module = device.create_shader_module(wgpu::util::make_spirv(&fs_bytes));
```
rukai, line 119: There is no guarantee that fs_bytes is aligned!

cwfitzgerald: It's going to be aligned to the alignment of the global memory allocator, which on all platforms we care about is > 4. Wouldn't hurt to add an assert here though.

src/lib.rs
```diff
@@ -571,6 +570,11 @@ impl Drop for ShaderModule {
     }
 }
 
+pub enum ShaderModuleSource<'a> {
```
cwfitzgerald, line 573: Could we get some documentation on this?

src/util/mod.rs
```diff
@@ -1,41 +1,15 @@
...
+    pub _align: [u32; 0],
```
cwfitzgerald, line 3: Is there any reason this struct can't have an explicit alignment attribute?

src/util/mod.rs
cwfitzgerald: "wrong magic word {:x}. Make sure you are using a **binary** SPIRV file."

examples/texture-arrays/main.rs
rukai: There already is an assert.

So heap memory is automatically aligned to 4.
But stack memory has no such automatic alignment?
I haven't heard of this before.

cwfitzgerald: Heap memory on linux is aligned to 16, windows 8, don't remember mac but it's probably one of them. Depends on the properties of the allocator, but it's often helpful inside the allocator to have a minimum size that you can allocate.

src/util/mod.rs
kvark: ooops

src/util/mod.rs
kvark: good idea! I'll try that

kvark:
Thank you both for reviewing! This is updated now, PTAL

src/util/mod.rs
```diff
@@ -1,41 +1,16 @@
-use std::{io, slice};
+#[repr(align(16))]
+pub struct WordAligned<Bytes: ?Sized>(pub Bytes);
```
cwfitzgerald, line 2: Minor nit: This should either be `align(4)` (which spirv requires) or be called something other than `WordAligned` which implies alignment to 2/4/8s (at least on x86 with a 16/32/64bit word)

src/util/mod.rs
```diff
@@ -1,41 +1,16 @@
...
-    unsafe {
-        // Writing all bytes through a pointer with less strict alignment when our type has no
-        // invalid bitpatterns is safe.
-        x.read_exact(slice::from_raw_parts_mut(
-            result.as_mut_ptr() as *mut u8,
-            words * 4,
-        ))?;
-        result.set_len(words);
-    }
+pub fn make_spirv<'a>(data: &'a [u8]) -> super::ShaderModuleSource<'a> {
```
rukai, line 4: Considering this is public, we should have documentation specifying data must be 4 aligned.

kvark:
Thank you! Last notes are now addressed.
bors r=cwfitzgerald,rukai

PR #385, opened 2020-06-17
Move `backends: BackendBit` parameter to `Instance::new()`

### LINKED ISSUE #337: Create Surface/Instance only for targeted backend
When calling `Surface::create` in the 0.5 release, or `Instance::new` in master, one Instance is created per backend that's supported. On Windows this means three Instances are created: one for Vulkan, DX12, and DX11. Perhaps my Vulkan drivers are particularly bad, but out of the 711ms it takes to create those three Instances on my computer, 706ms are taken up by the Vulkan Instance. If I chose to use a non-Vulkan Adapter/Device, it seems silly to have to pay the price for all that Vulkan instantiation.

Kimundi:
Fixes #337
Blocked on https://github.com/gfx-rs/wgpu/pull/730.

### DIFF
--- a/src/backend/direct.rs
+++ b/src/backend/direct.rs
@@ -329,8 +329,8 @@ impl crate::Context for Context {
         Ready<Result<(Self::DeviceId, Self::QueueId), crate::RequestDeviceError>>;
     type MapAsyncFuture = native_gpu_future::GpuFuture<Result<(), crate::BufferAsyncError>>;
 
-    fn init() -> Self {
-        wgc::hub::Global::new("wgpu", wgc::hub::IdentityManagerFactory)
+    fn init(backends: wgt::BackendBit) -> Self {
+        wgc::hub::Global::new("wgpu", wgc::hub::IdentityManagerFactory, backends)
     }
 
     fn instance_create_surface(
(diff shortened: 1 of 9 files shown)


src/backend/direct.rs
```diff
@@ -344,15 +344,14 @@ impl crate::Context for Context {
...
-        backends: wgt::BackendBit,
     ) -> Self::RequestAdapterFuture {
         let id = self.pick_adapter(
             &wgc::instance::RequestAdapterOptions {
                 power_preference: options.power_preference,
                 compatible_surface: options.compatible_surface.map(|surface| surface.id),
             },
             unsafe_extensions,
-            wgc::instance::AdapterInputs::Mask(backends, |_| PhantomData),
+            wgc::instance::AdapterInputs::Mask(wgt::BackendBit::all(), |_| PhantomData),
```
kvark, line 354: we could get the mask from the instance, could we?

Kimundi: Well technically we can only create a mask for the backends that sucessfully instantiated (unless we explicitly store backends in the `Instance` struct). And the way `pick_adapter` works means it would make no difference in any way, so...

kvark: yep, hence we are all merged :)

Issue #387, opened 2020-06-18
Merge `capture` example into another one as a screenshot feature


kvark:
I'm trying to get us less code to maintain, while still demonstrating all the features. A screenshot feature is useful, it could be added on top of any of the framework-based examples.

rukai:
I think it has value as is because it demonstrates headless rendering. But I understand your point about making it harder to maintain.

kvark:
ooh, you are right, it has headless stuff, which other examples aren't bothering with...

PR #390, opened 2020-06-18
Example feature matrix

kvark:
This PR introduces a feature matrix to show what feature can be found in which example. I expect we'll be linking to it a lot :)

It also renames the `describe` example so that we can distinguish easily between ones that are framework-based and standalone ones (starting with "hello").

Concerns:
- the star may be a bit too bleak... I haven't found a better one that would be short to type (choosing a long one means editing is uncomfortable).
- sometimes hard to figure out what deserves a feature in the table and what not... still a win to have at least something here
- we need to do something with `capture`, see #387, but I haven't figured out what exactly

also r? @rukai


cwfitzgerald:
That's interesting, on my machine (windows/firefox) it's quite prominent:

examples/README.md
```diff
@@ -0,0 +1,40 @@
...
+  - `hello-compute` for pure computing
+
+Notably, `capture` example show rendering without a surface/window. It reads back the contents and saves them to a file.
+
+All framework-based examples render to the window.
+
+## Feature matrix
+| Feature                | boids  | cube   | mipmap | msaa-line | shadow | skybox | texture-arrays | water  |
+| ---------------------- | ------ | ------ | ------ | --------- | ------ | ------ | -------------- | ------ |
+| vertex attribute       | :star: | :star: | :star: | :star:    | :star: |        | :star:         | :star: |
```
cwfitzgerald, line 15: Even though the features themselves are generally not pluralized, I would prefer to see these pluralized as it reads much better in docs.

examples/README.md
```diff
@@ -0,0 +1,40 @@
...
+| lines and points       |        |        |        | :star:    |        |        |                |        |
+| sampled color texture  | :star: | :star: | :star: |           |        | :star: | :star:         | :star: |
+| storage texture        | :star: |        |        |           |        |        |                |        |
+| binding array          |        |        |        |           |        |        | :star:         |        |
+| comparison sampler     |        |        |        |           | :star: |        |                |        |
+| subresource view       |        |        | :star: |           | :star: |        |                |        |
+| cubemap                |        |        |        |           |        | :star: |                |        |
+| multisampling          |        |        |        | :star:    |        |        |                |        |
+| off-screen rendering   |        |        |        |           | :star: |        |                | :star: |
+| stencil test           |        |        |        |           |        |        |                |        |
```
cwfitzgerald, line 26: I'd prefer: `stencil testing`

examples/README.md
```diff
@@ -0,0 +1,40 @@
...
+| depth test             |        |        |        |           | :star: |        |                | :star: |
+| read-only depth        |        |        |        |           |        |        |                | :star: |
```
cwfitzgerald, line 28: depth/stencil? or is it too large?

examples/README.md
```diff
@@ -0,0 +1,40 @@
...
+| blending               |        |        |        |           |        |        |                | :star: |
+| render bundle          |        |        |        | :star:    |        |        |                | :star: |
+| compute pass           | :star: |        |        |           |        |        |                |        |
+| optional extension     |        |        |        |           |        |        | :star:         |        |
```
cwfitzgerald, line 32: do we actually want to list the optional things it supports? We could have something like
`| - binding indexing |` to show it's a sub item of the optional extensions.

examples/README.md
kvark: yeah, I unconsciously pluralized a few things anyway :)

examples/README.md
```diff
@@ -0,0 +1,41 @@
...
+| cubemaps               |        |        |        |           |        | :star: |                |        |
+| multisampling          |        |        |        | :star:    |        |        |                |        |
+| off-screen rendering   |        |        |        |           | :star: |        |                | :star: |
+| stencil testing        |        |        |        |           |        |        |                |        |
+| depth testing          |        |        |        |           | :star: |        |                | :star: |
+| read-only depth        |        |        |        |           |        |        |                | :star: |
+| blending               |        |        |        |           |        |        |                | :star: |
+| render bundles         |        |        |        | :star:    |        |        |                | :star: |
+| compute passes         | :star: |        |        |           |        |        |                |        |
+| optional extensions    |        |        |        |           |        |        |                |        |
```
cwfitzgerald, line 32: This row should probably also have a star.

kvark: I was more thinking in line of blanking out the row completely. But a star works too :)

PR #391, opened 2020-06-18
Change `include_spirv` macro to accept tokens

neachdainn:
The `include_spirv` macros was set to accept literals only. Unfortunately, this means that other macros, such as `concat` and `env` aren't accepted, making it unusable with build script-compiled shaders. This change allows those macros to be used to load the SpirV at the expense of a potentially less helpful error message as the `literal` requirement is differed to the `include_bytes` invocation.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -143,7 +143,7 @@ fn test_vertex_attr_array() {
 /// It ensure the word alignment as well as the magic number.
 #[macro_export]
 macro_rules! include_spirv {
-    ($path:expr) => {
-        $crate::util::make_spirv(&$crate::util::WordAligned(*include_bytes!($path)).0)
+    ($($token:tt)*) => {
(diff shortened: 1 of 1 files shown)


neachdainn:
I'd also like to point out that the following code does *not* work (by any version of this macro) because it creates a temporary, though this isn't addressed by this pull request.

```rust
let spirv = include_spirv!("my.shader.spirv");
let module = device.create_shader_module(spirv);
```

src/macros.rs
```diff
@@ -143,7 +143,7 @@ fn test_vertex_attr_array() {
 /// It ensure the word alignment as well as the magic number.
 #[macro_export]
 macro_rules! include_spirv {
-    ($path:literal) => {
+    ($path:expr) => {
```
cwfitzgerald, line 146: _Strictly_ speaking, this should probably be `tt*` and we just dump everything inside the arguments to include_bytes, but that feels like way too much work.

neachdainn: I made the change.

PR #395, opened 2020-06-20
First Phase of Tracing Transition 

cwfitzgerald:
## Connections

This is the wgpu-rs component of https://github.com/gfx-rs/wgpu/pull/740.

## Notable Changes
- Added annotations for non-trivial wrapper functions
- Enabled building all features on docs.rs
- Took a dep on tracing.
- Fixes an assert that randomly stopped compiling.
- Added env variable WGPU_CHROME_TRACING which will be a path to the output chrome trace, if the feature is enabled.
- Added feature which corresponds directly to wgc feature.

### DIFF
--- a/examples/hello-triangle/main.rs
+++ b/examples/hello-triangle/main.rs
@@ -142,6 +142,7 @@ fn main() {
     #[cfg(not(target_arch = "wasm32"))]
     {
         env_logger::init();
+
         // Temporarily avoid srgb formats for the swapchain on the web
         futures::executor::block_on(run(event_loop, window, wgpu::TextureFormat::Bgra8UnormSrgb));
     }
--- a/src/backend/direct.rs
+++ b/src/backend/direct.rs
@@ -184,6 +184,7 @@ mod pass_impl {
             &mut self,
             render_bundles: I,
         ) {
+            wgc::span!(_guard, TRACE, "RenderPass::execute_bundles wrapper");
             let temp_render_bundles = render_bundles.cloned().collect::<SmallVec<[_; 4]>>();
             unsafe {
                 wgpu_render_pass_execute_bundles(
@@ -450,6 +451,7 @@ impl crate::Context for Context {
         device: &Self::DeviceId,
         desc: &BindGroupDescriptor,
     ) -> Self::BindGroupId {
+        wgc::span!(_guard, TRACE, "Device::create_bind_group wrapper");
         use wgc::binding_model as bm;
 
         let texture_view_arena: Arena<wgc::id::TextureViewId> = Arena::new();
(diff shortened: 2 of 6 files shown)


cwfitzgerald:
Should be ready to review!

examples/hello-triangle/main.rs
```diff
@@ -142,6 +142,13 @@ fn main() {
     #[cfg(not(target_arch = "wasm32"))]
     {
         env_logger::init();
+
+        #[cfg(feature = "subscriber")]
+        {
+            let chrome_tracing_dir = std::env::var("WGPU_CHROME_TRACING");
+            wgpu::util::initialize_default_subscriber(chrome_tracing_dir.ok());
```
kvark, line 149: I don't think we need to show it off in hello- examples. Just framework is OK and most useful anyway.

src/util/mod.rs
```diff
@@ -15,8 +18,8 @@ pub fn make_spirv<'a>(data: &'a [u8]) -> super::ShaderModuleSource<'a> {
     const MAGIC_NUMBER: u32 = 0x0723_0203;
 
     let (pre, words, post) = unsafe { data.align_to::<u32>() };
-    assert_eq!(pre, &[], "data offset is not aligned to words!");
-    assert_eq!(post, &[], "data size is not aligned to words!");
+    assert!(pre.is_empty(), "data offset is not aligned to words!");
```
kvark, line 21: is this needed? I like the previous ones better

src/backend/direct.rs
```diff
@@ -697,6 +701,8 @@ impl crate::Context for Context {
         mode: MapMode,
         range: Range<wgt::BufferAddress>,
     ) -> Self::MapAsyncFuture {
+        wgc::span!(_guard, TRACE, "Buffer::buffer_map_async wrapper");
```
kvark, line 704: dummy question: why are we even tracing wgpu-rs? Most methods here are just pass-through to wgpu-core, which is already traced

examples/hello-triangle/main.rs
cwfitzgerald: The reason I added this here too is because once we switch to `tracing` entirely, env_logger won't do anything, and I would imagine we want logging enabled. I can remove this for now and we can revisit it later if you want.

src/backend/direct.rs
cwfitzgerald: These exist so that non trivial wrappers show up in the chrome tracing trace. This information isn't very useful for logging, so I reduced it to the maximum verbosity.

cwfitzgerald:
Wooooo!

PR #396, opened 2020-06-20
Update wgpu without peek-poke

kvark:
Depends on https://github.com/gfx-rs/wgpu/pull/739
Everything works, however I think it's worth changing the render pass color/depth/stencil descriptors to be more rusty, e.g.
```rust
struct ColorAttachmentDescriptor {
  attachment: &'a TextureView,
  resolve_target: Option<&'a TextureView>,
  clear_color: Option<Color>,
  store_result: bool,
}

enum DepthStencilChannel<T> {
  ReadOnly,
  Mutable { clear_value: Option<T>, store_result: bool },
}
struct DepthStencilAttachmentDescriptor {
}
```
This would also involve wgpu-type/wgpu-core changes, but can be done as a follow-up (don't want to block on them).

There is a value in doing it in this PR, so that end users don't get multiple disruptions, but we need to find a proper shape first.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,14 +28,14 @@ vulkan = ["wgc/gfx-backend-vulkan"]
 package = "wgpu-core"
 version = "0.5"
 git = "https://github.com/gfx-rs/wgpu"
-rev = "a02a56684114da702a64e30af49d0167e273402b"
+rev = "c7be94047d156a3bde88d03cd5c229bedb6efe62"
 features = ["raw-window-handle"]
 
 [dependencies.wgt]
 package = "wgpu-types"
 version = "0.5"
 git = "https://github.com/gfx-rs/wgpu"
-rev = "a02a56684114da702a64e30af49d0167e273402b"
+rev = "c7be94047d156a3bde88d03cd5c229bedb6efe62"
 
 [dependencies]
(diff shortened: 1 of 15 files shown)


kvark:
I'm thinking of going with this API:
```rust
/// Operation to perform to the output attachment at the start of a renderpass.
pub enum LoadOp<V> {
    /// Clear with a specified value.
    Clear(V),
    /// Load from memory.
    Load,
}

/// Pair of load and store operations for an attachment aspect.
pub type Operations<V> = (LoadOp<V>, StoreOp);

/// Describes a color attachment to a [`RenderPass`].
pub struct RenderPassColorAttachmentDescriptor<'a> {
    pub attachment: &'a TextureView,
    pub resolve_target: Option<&'a TextureView>,
    pub ops: Operations<Color>,
}
/// Describes a depth/stencil attachment to a [`RenderPass`].
pub struct RenderPassDepthStencilAttachmentDescriptor<'a> {
    pub attachment: &'a TextureView,
    pub depth_ops: Option<Operations<f32>>,
    pub stencil_ops: Option<Operations<u32>>,
}
```

cwfitzgerald:
I like it!

kvark:
thanks for reviewing quickly!
bors r=cwfitzgerald

PR #397, opened 2020-06-20
Fix the mipmap example

hasenbanck:
The current default value for the lod_max_clamp of a sampler is 0.
This would deactivate mipmaping all together. Setting it to
f32::MAX makes sure to always enable it. It's the default value apple
uses in Metal, it seems to be valid usage in Vulkan (maxLod in
VkSamplerCreateInfo) and in DX12 (MaxLOD in D3D12_SAMPLER_DESC).

Before:
broken

After:
fixed


hasenbanck:
@kvark Sure. I can do that.

kvark:
So this will be fixed automatically upon the next wgpu dependency update, closing

PR #400, opened 2020-06-20
Added test for "capture" example

bfrazho:
I've added a test to validate the capture example. I was planning on using the existing "screenshot.png" that was in there for the assertion for the test, but apparently was slightly different from the one that I was generating. I ended up replacing the screenshot.png with what the test was generating.

I'm a little concerned that they were different from the start, but maybe the png generating library made some small changes over time. I did check to make sure that it generated the same png on Windows and Linux on the same computer.

Let me know if there is anything that you would like me to change!


kvark:
Hi!
Soon, we'd want to transform the example into something richer and covering more ground, see #387 . So I'm not sure if just adding this test will be compatible with future changes.
Notably, the testing of some rendering producing some output is going to be done on a wide range of cases, as a part of the API tracing infrastructure in `wgpu` repository.

examples/capture/main.rs
```diff
@@ -164,12 +167,39 @@ fn main() {
...
+    use std::fs::remove_file;
+
+    #[test]
+    fn ensure_generated_png_matches_expected(){
+        futures::executor::block_on(assert_generated_png_matches_expected(
+            "examples/capture/test_output.png",
+            "examples/capture/screenshot.png"));
+    }
+
+    async fn assert_generated_png_matches_expected(test_file_name: &str, expected_file_name: &str) {
```
kvark, line 194: Let's instead compare the data we read back from `wgpu` (basically, raw pixels). There is no need for the data to go PNG encoding and saving to a file, just so we can load it back and compare.

bfrazho: I'm definitely on board with that. I was thinking about trying to test for that, but wasn't sure if writing out the screenshot to a file was part of the example.

kvark: It's a part of the example, but it doesn't mean that we should be testing that part. We can let PNG folks do their testing, and stdlib likewise with file output :)

bfrazho:
So I think I've got changes that match what you were describing on my fork. I've never had to amend a pull request, do you have any direction around how to do that? Assuming you don't have any other changes in mind! :)

examples/capture/main.rs
```diff
@@ -163,43 +162,60 @@ async fn create_png_for_buffer(png_output_path: &str, width: usize, height: usiz
...
+    async fn assert_generated_data_matches_expected(file_containing_expected_data: &str) {
+        let (device, output_buffer, _) = create_red_image_with_dimensions(100usize, 200usize).await;
+        let buffer_slice = output_buffer.slice(..);
+        let buffer_future = buffer_slice.map_async(wgpu::MapMode::Read);
+        device.poll(wgpu::Maintain::Wait);
+        buffer_future.await.expect("failed to map buffer slice for capture test");
+        let padded_buffer = buffer_slice.get_mapped_range();
+        let actual_buffer_data = padded_buffer.to_vec();
+        let expected_buffer_data = fs::read(file_containing_expected_data).unwrap();
+        assert_eq!(actual_buffer_data, expected_buffer_data);
```
kvark, line 219: that can backfire: once there is an error, it will print out both, and they will spam the whole console.

examples/capture/main.rs
```diff
@@ -163,43 +162,60 @@ async fn create_png_for_buffer(png_output_path: &str, width: usize, height: usiz
...
-        remove_file(test_file_name).expect("failed to remove test file");
+    async fn assert_generated_data_matches_expected(file_containing_expected_data: &str) {
+        let (device, output_buffer, _) = create_red_image_with_dimensions(100usize, 200usize).await;
+        let buffer_slice = output_buffer.slice(..);
+        let buffer_future = buffer_slice.map_async(wgpu::MapMode::Read);
+        device.poll(wgpu::Maintain::Wait);
+        buffer_future.await.expect("failed to map buffer slice for capture test");
+        let padded_buffer = buffer_slice.get_mapped_range();
+        let actual_buffer_data = padded_buffer.to_vec();
+        let expected_buffer_data = fs::read(file_containing_expected_data).unwrap();
```
kvark, line 218: do we actually need this? We know that the buffer needs to be all red. Can we just check if all the 4-byte chunks of the data are red? e.g.
```
for chunk in actual_buffer_data.chunks(4) {
  assert_eq!(chunk, &[0xFFu8, 0, 0, 1]);
}
```

examples/capture/main.rs
bfrazho: I could definitely put a couple assertions for size and making sure it is all red. I was kind of avoiding doing a shortcut like that partially because I wanted to build something that could work for any output easily if it ever changed. And partially because I'm just learning as I'm going ;)
I think I could just change the assertion to assert!(actual_buffer_data == expected_buffer_data, "fail message") or something like that. That way it doesn't bother trying to print the data at all.
Just let me know which you would like me to do! :)

kvark: There is certainly a value in something that would compare the rendering result with different operations. The trick is: we want to have that testing infra in `wgpu` (a bit lower level), so wgpu-rs stays a light wrapper, and it doesn't need to be as rigorously tested. So for this little test it's totally fine to check that everything is red, instead of rolling out a general solution with a reftest in an image.

bfrazho: Sounds good, I'll make the change when I get the chance! Thanks! :)

bfrazho:
No problem! I think it should be all set. Thanks for the help! :)
I'm guessing you'll be the one to close this once it gets merged in?

PR #402, opened 2020-06-21
Re-enable auto-deploy from gecko branch

grovesNL:
As mentioned on Matrix, the `gecko` branch should be safe to use for auto-deployment to GitHub pages for the wgpu-rs examples on wasm.

We can use the `gecko` branch for now and eventually switch back to `master`.


grovesNL:
bors r=kvark

kvark:
Bors r+

PR #406, opened 2020-06-22
Add Battery Feature and Default it On

### LINKED ISSUE #394: Expose Battery Dep By Default

cwfitzgerald:
Closes #394

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,8 @@ all-features = true
 [lib]
 
 [features]
-default = []
+default = ["battery"]
+battery = ["wgc/battery"]
 trace = ["serde", "wgc/trace"]
 replay = ["serde", "wgc/replay"]
 subscriber = ["wgc/subscriber"]


Cargo.toml
```diff
@@ -13,7 +13,8 @@ exclude = ["etc/**/*", "examples/**/*", "tests/**/*", "Cargo.lock", "target/**/*
 [lib]
 
 [features]
-default = []
+default = ["battery"]
+battery = ["wgc/battery"]
```
kvark, line 17: maybe it's easier to just add it to the feature list of near `features = ["raw-window-handle"]` ?
the default feature here doesn't exactly make sense for the Web target, where wgc is not even supposed to compile...
it would be good to check how this affect build times, too

cwfitzgerald: hmm, the problem with doing that is that you can't turn off the battery dep. I'm not sure if you can do different defaults on differnet platforms, will poke around a bit more.

cwfitzgerald: So in the implementation of #484 we learned that this should work fine as is, as cargo will just ignore the `wgc` feature if wgc is not a dep on that target.

cwfitzgerald:
Discussed on matrix, deciding to rip out battery feature as it's likely out of scope for wgpu entirely, and the users can implement this if they desire.

PR #407, opened 2020-06-23
Build Docs with Nightly

cwfitzgerald:
This builds docs with nightly by default. We use intra-doc links which is currently only supported with nightly. docs.rs also uses the latest nightly to build our code, so this will match behavior with docs.rs. I added a fallback to stable in case nightly has issues so we still have working docs.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -13,8 +13,20 @@ jobs:
       - name: Checkout the code
         uses: actions/checkout@v2
 
-      - name: Build the docs
-        run: cargo doc --lib --all-features
+      - name: Install latest nightly
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          override: true
+        continue-on-error: true
+
+      - name: Build the docs (nightly)
+        run: |
+          cargo +nightly doc --lib --all-features
+
+      - name: Build the docs (stable)
+        run: cargo +stable doc --lib --all-features
+        if: ${{ failure() }}
 
       - name: Deploy the docs
         uses: JamesIves/github-pages-deploy-action@releases/v3


.github/workflows/docs.yml
```diff
@@ -13,8 +13,14 @@ jobs:
       - name: Checkout the code
         uses: actions/checkout@v2
 
-      - name: Build the docs
-        run: cargo doc --lib --all-features
+      - name: Build the docs (nightly)
+        run: |
+          rustup install nightly
```
kvark, line 18: why aren't we using the same kind of toolchain setup as on CI:
```
 - name: Install latest nightly
      uses: actions-rs/toolchain@v1
      with:
        toolchain: nightly
        override: true
      continue-on-error: true
```

.github/workflows/docs.yml
```diff
@@ -13,8 +13,14 @@ jobs:
...
+          cargo +nightly doc --lib --all-features
+
+      - name: Build the docs (stable)
+        run: cargo +stable doc --lib --all-features
```
kvark, line 22: why are we building both nightly and stable docs?

.github/workflows/docs.yml
cwfitzgerald: Cause I didn't notice that's how the ci did it, will fix.

.github/workflows/docs.yml
cwfitzgerald: I described this in the dpr description, the stable docs will only fire if nightly fails, as I know we've had issues in the past with nightly not working correctly.

kvark: Are you sure that's how the script is configured? It seems to me that `run: cargo +stable` goes before `if`, where it should be the other way around?

cwfitzgerald: Yay declarative yaml! The order of the arguments doesn't actually matter in this case. https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-status-check-functions. This is the reference I used to make myself believe it worked in the first place.

PR #408, opened 2020-06-23
Staging belt

kvark:
Implementing the belt itself isn't too hard, but integrating it with the example framework and winit's event loop is somewhat challenging. Perhaps, there needs to be a `LocalSpawner` parameter to `render()` like I did [in vange-rs](https://github.com/kvark/vange-rs/blob/e7bc944ef5a5a47d04cedd1c4d299295a1693fd3/bin/boilerplate.rs#L26)?

Edit: code is all ready and shiny... but it does NOT work. Buffers don't come back from mapping for some reason.

### DIFF
--- a/examples/skybox/main.rs
+++ b/examples/skybox/main.rs
@@ -1,3 +1,5 @@
+use futures::task::{LocalSpawn, LocalSpawnExt};
+
 #[path = "../framework.rs"]
 mod framework;
 
@@ -19,6 +21,7 @@ pub struct Skybox {
     bind_group: wgpu::BindGroup,
     uniform_buf: wgpu::Buffer,
     uniforms: Uniforms,
+    staging_belt: wgpu::util::StagingBelt,
 }
 
 impl Skybox {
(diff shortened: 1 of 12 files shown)


examples/skybox/main.rs
```diff
@@ -265,11 +267,14 @@ impl framework::Example for Skybox {
...
+        let raw_uniforms = raw_uniforms(&self.uniforms);
+        self.staging_belt.write_buffer(
             &self.uniform_buf,
             0,
-            bytemuck::cast_slice(&raw_uniforms(&self.uniforms)),
-        );
+            wgpu::BufferSize::new((raw_uniforms.len() * 4) as wgpu::BufferAddress).unwrap(),
+            device,
+        ).copy_from_slice(bytemuck::cast_slice(&raw_uniforms));
+        self.staging_belt.flush(queue, device);
```
lachlansneff, line 277: If the future returned by `StagingBelt::flush` isn't polled, the buffers won't be mapped again, right?

kvark: Right. This is incomplete, I got stuck on figuring out the futures interaction with the event loop.

kvark: actually, your concern still stands. Even with the current code, the poll() isn't happening.

kvark:
bors delegate=kyren

PR #413, opened 2020-06-27
Improve docs for main crate and `hello-triangle` example

GabrielMajeri:
Some minor fixes for the main crate docs, as well as some comments in the `hello-triangle` example.

### DIFF
--- a/examples/hello-triangle/main.rs
+++ b/examples/hello-triangle/main.rs
@@ -12,13 +12,15 @@ async fn run(event_loop: EventLoop<()>, window: Window, swapchain_format: wgpu::
         .request_adapter(
             &wgpu::RequestAdapterOptions {
                 power_preference: wgpu::PowerPreference::Default,
+                // Request an adapter which can render to our surface
                 compatible_surface: Some(&surface),
             },
             wgpu::UnsafeFeatures::disallow(),
         )
         .await
-        .unwrap();
+        .expect("Failed to find an appropiate adapter");
 
+    // Create the logical device and command queue
     let (device, queue) = adapter
         .request_device(
             &wgpu::DeviceDescriptor {
@@ -29,8 +31,9 @@ async fn run(event_loop: EventLoop<()>, window: Window, swapchain_format: wgpu::
             None,
         )
         .await
-        .unwrap();
+        .expect("Failed to create device");
 
+    // Load the shaders from disk
     let vs_module = device.create_shader_module(wgpu::include_spirv!("shader.vert.spv"));
     let fs_module = device.create_shader_module(wgpu::include_spirv!("shader.frag.spv"));
 
@@ -48,13 +51,8 @@ async fn run(event_loop: EventLoop<()>, window: Window, swapchain_format: wgpu::
             module: &fs_module,
             entry_point: "main",
         }),
-        rasterization_state: Some(wgpu::RasterizationStateDescriptor {
-            front_face: wgpu::FrontFace::Ccw,
-            cull_mode: wgpu::CullMode::None,
-            depth_bias: 0,
-            depth_bias_slope_scale: 0.0,
-            depth_bias_clamp: 0.0,
-        }),
+        // Use the default rasterizer state: no culling, no depth bias
+        rasterization_state: None,
         primitive_topology: wgpu::PrimitiveTopology::TriangleList,
         color_states: &[wgpu::ColorStateDescriptor {
             format: swapchain_format,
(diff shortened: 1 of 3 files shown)


examples/hello-triangle/main.rs
```diff
@@ -83,21 +81,13 @@ async fn run(event_loop: EventLoop<()>, window: Window, swapchain_format: wgpu::
     let mut swap_chain = device.create_swap_chain(&surface, &sc_desc);
 
     event_loop.run(move |event, _, control_flow| {
-        // force ownership by the closure
```
GabrielMajeri, line 86: I've removed this and it seems everything still worked fine

src/lib.rs
```diff
@@ -1029,6 +1043,22 @@ impl Instance {
             })
     }
 
+    /// Retrieves an [adapter][Adapter] which matches the given [options][RequestAdapterOptions].
+    ///
+    /// Some options are "soft", so treated as non-mandatory. Others are "hard".
+    ///
+    /// If no adapters are found that suffice all the "hard" options, `None` is returned.
+    pub fn request_adapter(
```
GabrielMajeri, line 1051: I've moved this function to change the order it shows up in the generated documentation; now both adapter-related functions are together.

src/util/mod.rs
```diff
@@ -9,14 +11,15 @@ pub use belt::StagingBelt;
 #[repr(align(4))]
 pub struct WordAligned<Bytes: ?Sized>(pub Bytes);
 
-/// Treat the given by slice as a SPIR-V module.
+/// Treat the given byte slice as a SPIR-V module.
 ///
-/// # Errors
+/// # Panic
```
GabrielMajeri, line 16: This can be changed back to errors if the function is converted to use `Result` instead of `assert!`

GabrielMajeri:
I'm thinking of adding `#[deny(missing_docs)]` to the crate and documenting the few remaining items. Would that be OK?

examples/hello-triangle/main.rs
cwfitzgerald: The problem is that `run` never actually returns, so things aren't properly dropped. It's not necessary (not calling drop is always safe), but useful to have.

src/lib.rs
```diff
@@ -343,21 +345,22 @@ trait Context: Sized {
...
 pub struct Instance {
     context: Arc<C>,
 }
 
 /// Handle to a physical graphics and/or compute device.
 ///
-/// An `Adapter` can be used to open a connection to the corresponding device on the host system,
-/// yielding a [`Device`] object.
+/// Adapters can be used to [open a connection][Adapter::request_device] to
+/// the corresponding [device][Device] on the host system.
```
cwfitzgerald, line 361: This link works as is (here and all other places). I would prefer to just use a raw link even if the case isn't right because it's in a inline code block, so it's clear it's a code reference.

src/lib.rs
```diff
@@ -760,7 +765,7 @@ pub enum BindingResource<'a> {
     TextureView(&'a TextureView),
     /// Binding is backed by an array of textures.
     ///
-    /// [`Capabilities::SAMPLED_TEXTURE_BINDING_ARRAY`] must be supported to use this feature.
+    /// [`Features::SAMPLED_TEXTURE_BINDING_ARRAY`] must be supported to use this feature.
```
cwfitzgerald, line 768: Good catch

examples/hello-triangle/main.rs
GabrielMajeri: As far as I understand, since we set `*control_flow = ControlFlow::Exit` when the user closes the window, `event_loop.run` will stop, and then the big `run` will return?

src/lib.rs
GabrielMajeri: I've updated the PR to match the previous style

examples/hello-triangle/main.rs
cwfitzgerald: Due to restrictions on the web and mobile, `run` never returns (returns `!`). There is `run_return`, but it's non-portable.

kvark: Not dropping the instance and stuff means API validation errors on exit. We don't want that. Let's bring back this ownership hack, until we have a better solution

GabrielMajeri: Thanks for the clarifications! I've added the code back, with a comment explaining why it's necessary.

GabrielMajeri:
@kvark Your opinion on adding `#![deny(missing_docs)]`?

kvark:
I'd rather just `warn` tbh, denying is overly restrictive and harms prototyping.

kvark:
bors r=cwfitzgerald,kvark

GabrielMajeri:
@kvark Alright, I've added a `#![warn(missing_docs)]`. As of this PR, we have all the public APIs documented.

PR #414, opened 2020-06-27
Implement MultiDrawIndirect Extensions

cwfitzgerald:
The wgpu-rs component of https://github.com/gfx-rs/wgpu/pull/754.

I didn't include the modifications to the example as they are particularly awful and ugly. One of my next steps following this is to make an example of both binding indexing and these multi-draw extensions.


cwfitzgerald:
Alrighty, everything should be good to go here.

PR #428, opened 2020-07-09
Split render pass docs into multiple sections, divided by feature

lachlansneff:
This does not affect usage at all.

Here's what the docs look like:

Screen Shot 2020-07-09 at 11 11 30 AM


lachlansneff:
I'm trying to see if I can get the feature sections to be minimized by default, but it doesn't look like there's any way to do it.

cwfitzgerald:
Thanks, looks lovely!
bors r+

Issue #352, opened 2020-06-06
Example `shadow` panics on window resize [Windows, DX12]


krooq:
I'm on the lastest commit d06470e and running the shadow example on Windows 10, I believe I'm using DX12.

When I resize the window I get the following panic trace

```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Extent { width: 1024, height: 769, depth: 1 }`,
 right: `Extent { width: 1024, height: 768, depth: 1 }`: Extent state must match extent from view', C:\Users\Krooq\main\tools\rust\rustup\toolchains\nightly-x86_64-pc-windows-gnu\lib/rustlib/src/rust\src\libstd\macros.rs:16:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[2020-06-06T12:44:33Z ERROR gfx_memory::heaps] Heaps still have 8 types live on drop
[2020-06-06T12:44:33Z ERROR gfx_memory::allocator::general] Memory leak: SizeEntry(11206656) is still used
[2020-06-06T12:44:33Z ERROR gfx_memory::allocator::general] Memory leak: SizeEntry(3276800) is still used
[2020-06-06T12:44:33Z ERROR gfx_memory::allocator::general] Memory leak: SizeEntry(3801088) is still used
[2020-06-06T12:44:33Z ERROR gfx_memory::allocator::general] Memory leak: SizeEntry(256) is still used
[2020-06-06T12:44:33Z ERROR gfx_memory::allocator::general] Memory leak: SizeEntry(2048) is still used
[2020-06-06T12:44:33Z ERROR gfx_memory::allocator::general] Memory leak: SizeEntry(16384) is still used
[2020-06-06T12:44:33Z ERROR gfx_memory::allocator::general] Memory leak: SizeEntry(131072) is still used
[2020-06-06T12:44:33Z ERROR gfx_memory::allocator::general] Memory leak: SizeEntry(1024) is still used
[2020-06-06T12:44:33Z ERROR gfx_memory::allocator::linear] Not all allocation from LinearAllocator was freed
[2020-06-06T12:44:33Z ERROR gfx_descriptor::allocator] DescriptorAllocator is dropped
error: process didn't exit successfully: `target\debug\examples\shadow.exe` (exit code: 101)
```

When I run with extra backtrace it is saying something about the drop of RenderPass on line 814.

kvark:
As reported in #403, same problem is on Vulkan.

[...]

PR #434, opened 2020-07-11
recreate swapchain on framework.rs examples on resize, ref #352

m4b:
There still seem to be residual race condition like errors w.r.t. resizing on at least x11, but this does fix the panic on x11.


cwfitzgerald:
Looks good, my only real concern is that likely the call to the swapchain creation should likely go before resize is called, not after.

kvark:
@cwfitzgerald I think it's fine. The user isn't supposed to hold any relation to swapchain at that point :)
bors r=kvark,cwfitzgerald

PR #437, opened 2020-07-13
Use common types from `wgpu-types`

GabrielMajeri:
The `wgpu-rs` counterpart to https://github.com/gfx-rs/wgpu/pull/791

### DIFF
--- a/examples/boids/main.rs
+++ b/examples/boids/main.rs
@@ -193,16 +193,16 @@ impl framework::Example for Example {
         for i in 0..2 {
             particle_bind_groups.push(device.create_bind_group(&wgpu::BindGroupDescriptor {
                 layout: &compute_bind_group_layout,
-                bindings: &[
-                    wgpu::Binding {
+                entries: &[
+                    wgpu::BindGroupEntry {
                         binding: 0,
                         resource: wgpu::BindingResource::Buffer(sim_param_buffer.slice(..)),
                     },
-                    wgpu::Binding {
+                    wgpu::BindGroupEntry {
                         binding: 1,
                         resource: wgpu::BindingResource::Buffer(particle_buffers[i].slice(..)),
                     },
-                    wgpu::Binding {
+                    wgpu::BindGroupEntry {
                         binding: 2,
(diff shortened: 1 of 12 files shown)


examples/boids/main.rs
```diff
@@ -193,16 +193,16 @@ impl framework::Example for Example {
         for i in 0..2 {
             particle_bind_groups.push(device.create_bind_group(&wgpu::BindGroupDescriptor {
                 layout: &compute_bind_group_layout,
-                bindings: &[
-                    wgpu::Binding {
+                entries: &[
+                    wgpu::BindGroupEntry {
                         binding: 0,
                         resource: wgpu::BindingResource::Buffer(sim_param_buffer.slice(..)),
                     },
```
GabrielMajeri, line 200: I've renamed `bindings` to `entries` and `Binding` to `BindGroupEntry` for consistency with `wgpu-core`. This will cause some breakage unfortunately.

kvark: thank you!

src/lib.rs
```diff
@@ -802,102 +795,25 @@ pub enum BindingResource<'a> {
...
 /// Describes a compute pipeline.
-#[derive(Clone)]
-pub struct ComputePipelineDescriptor<'a> {
-    /// The layout of bind groups for this pipeline.
-    pub layout: &'a PipelineLayout,
-
-    /// The compiled compute stage and its entry point.
-    pub compute_stage: ProgrammableStageDescriptor<'a>,
-}
+pub type ComputePipelineDescriptor<'a> =
```
cwfitzgerald, line 815: For the other descriptor types, we included the the wgt type as XDescriptorBase for docs purposes. Do we still want to do that? If so, these should be imported that way too.

kvark: Oh right, I totally missed that. Fully agree that we need to be consistent with regards to naming here.
Either stop using the `Base` suffix, or use it everywhere. Would the latter cause any problems to the docs? If not, I'm fine with either.
This would need to be changed in wgpu though, so technically this PR to wgpu-rs doesn't need to be blocked. We can get it in, and then just update the naming here.

PR #442, opened 2020-07-16
replace vertex_format_size macro with VertexFormat size function

bootra-dev:
This pull request depends on https://github.com/gfx-rs/wgpu/pull/802

I'm not sure if multiple pull requests is the right way to handle this - let me know if I need to use another workflow.


bootra-dev:
Can this be merged now that https://github.com/gfx-rs/wgpu-rs/commit/be388200dc87ce5c1daedb7cb85cb58cfb4cae9e is in?

kvark:
let's see!
bors r+

PR #444, opened 2020-07-17
Doc link correction

Hscowef:
Hello, I saw that in the documentation almost none of the path or reference to the elements of the crate were linked so I took some time and added the links for most of the things. It may have some typo errors, sorry if it is the case.
(This is the first time I "contribute" to a project so I don't know if I did this pr how I was supposed to 😄 )

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1659,7 +1659,7 @@ impl<'a> RenderPass<'a> {
 
     /// Sets the active index buffer.
     ///
-    /// Subsequent calls to [`draw_indexed`](RenderPass::draw_indexed) on this [`RenderPass`] will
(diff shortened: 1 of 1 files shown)


rukai:
Unfortunately (or maybe fortunately) this already works as is, you just need to use the nightly rust doc to get it to work: `cargo +nightly doc`
You can see the result here: https://wgpu.rs/doc/wgpu/

Hscowef:
Oh neat! I wanted to look at the doc more in detailin the first place so it isn't useless efforts.

kvark:
Oh sad. But it's great that it works in the simpler way.
I suppose we are closing the PR then?

Issue #188, opened 2020-03-03
The recent lifetime changes to `RenderPass` make it difficult to use. 


aloucks:
Related PRs:

https://github.com/gfx-rs/wgpu-rs/pull/155
https://github.com/gfx-rs/wgpu-rs/pull/168

For context, I was looking at updating [imgui-wgpu](https://github.com/Yatekii/imgui-wgpu-rs/blob/master/src/lib.rs#L369) to work with the current master.

Now that `set_index_buffer` (and similar methods) take the a `&'a Buffer` instead of `&Buffer`, it has a ripple effect of "infecting" the surrounding scope with this lifetime.

While attempting to iterate and add draw calls, you end up with either "multiple mutable borrow" errors on wherever you're storing the `Buffer` or similar lifetime errors like "_data from `self` flows into `rpass` here_"

I would think that by calling `set_index_buffer` (and similar methods) that it would increment the ref-count on the buffer so that this lifetime restriction isn't needed.

kvark:
This is indeed a feature that makes it more difficult to use. What we gain from it is very lightweight render pass recording. Explanation follows.

`wgpu` is designed to fully support multi-threading. We do this by having whole storages of different objects locked upon access. So generally, touching anything has a CPU cost. If we had to access each of the refcounts of the resources used by render commands, we'd be paying that cost during the render pass recording, which is hot code. Now with the recent changes, recording a pass is very lightweight, no locks involved.

Most of the dependent resources are meant to outlive the pass anyway. Only a few, like the index buffers you create dynamically, become problematic. Generally, unless you are creating many resources during recording, it's easy to work around. If you are doing that, you aren't on a good path performance wise, and should consider creating one big buffer instead per frame.

Another way to possibly address this is to have `wgpu-rs` ensuring the lifetimes of the objects by other means, like keeping a refcount (note: talking about `wgpu-rs` here, not `wgpu`, which already keeps a refcount, but we'd need to lock an object storage to access it). This is far fetched and not something I'm excited about :)

head's up to @mitchmindtree, who is on 0.4 and will face this issue soon. It would be good to know how much this would affect their case.

aloucks:
> Most of the dependent resources are meant to outlive the pass anyway. Only a few, like the index buffers you create dynamically, become problematic.

[...]

PR #446, opened 2020-07-17
ArcRenderPass in shadow example

StarArawn:
fixes: #188 ??!?

I was recently ran into this problem where I had `Arc`'s that I had to pass to the render pass. I thought shadow was probably the best place to showcase that, although arc isn't really needed for anything there..


StarArawn:
Closing this for now as I think about possible example scenarios and come up with something different.

PR #450, opened 2020-07-19
Implement more derives for `Operations`

### LINKED ISSUE #422: wgpu::Operations and wgpu::LoadOp are missing some useful derives
Could implement `Copy`, `Eq`, `Serialize`, `Deserialize` where `V` has the appropriate trait. (`LoadOp` already has copy but `Operations` doesn't).

`Default` is also a possibility but it depends on whether there's a consensus on what is a sensible default. I'm thinking `LoadOp<V>` would have a default of `LoadOp::Clear(V::default())` (`where V: Default`) and `Operations<V>` could just derive `Default` (clear with the default value for `V` and no store).

GabrielMajeri:
- Implements a `Default` value for `LoadOp` and `Operations` when `V: Default`
- Derives `Eq` for those same traits
- Adds a new optional dependency `serde` which will derive the serialization traits for `LoadOp` and `Operations`, as requested in the issue

Closes #422

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -818,7 +818,7 @@ impl<V: Default> Default for LoadOp<V> {
 }
 
 /// Pair of load and store operations for an attachment aspect.
-#[derive(Copy, Clone, Debug, Default, Hash, Eq, PartialEq)]
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq)]
 #[cfg_attr(feature = "trace", derive(Serialize))]
 #[cfg_attr(feature = "replay", derive(Deserialize))]
 pub struct Operations<V> {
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -797,16 +799,24 @@ pub enum BindingResource<'a> {
 }
 
 /// Operation to perform to the output attachment at the start of a renderpass.
-#[derive(Clone, Copy, Debug, Hash, PartialEq)]
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
```
kvark, line 803: it seems to me that we should make it so `trace` feature derives serialization, and `replay` feature derives deserialization for these types. So `trace` and `replay` should both include `serde`

src/lib.rs
```diff
@@ -797,23 +801,42 @@ pub enum BindingResource<'a> {
...
     pub load: LoadOp<V>,
     /// Whether data will be written to through this attachment.
     pub store: bool,
 }
 
+impl<V: Default> Default for Operations<V> {
+    fn default() -> Self {
+        Self {
+            load: Default::default(),
+            store: true,
```
GabrielMajeri, line 835: I've seen [your comment](https://github.com/gfx-rs/wgpu-rs/issues/422#issuecomment-653777050) again and modified `store` to be `true` by default, as does upstream WebGPU.

PR #452, opened 2020-07-19
Update to latest wgpu

rukai:
public API changes:
*    get_next_frame -> get_current_frame to better reflect the name of webgpu api being called.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,14 +27,14 @@ vulkan = ["wgc/gfx-backend-vulkan"]
 package = "wgpu-core"
 version = "0.5"
 git = "https://github.com/gfx-rs/wgpu"
-rev = "a0ed09a6a83dc0d16c6ad711cf139a5fcfbb9c35"
+rev = "3c6ee8766ae8df8fc5c2b70f7c37d634ff3b79aa"
 features = ["raw-window-handle"]
 
 [dependencies.wgt]
 package = "wgpu-types"
 version = "0.5"
 git = "https://github.com/gfx-rs/wgpu"
-rev = "a0ed09a6a83dc0d16c6ad711cf139a5fcfbb9c35"
+rev = "3c6ee8766ae8df8fc5c2b70f7c37d634ff3b79aa"
 
 [dependencies]
 arrayvec = "0.5"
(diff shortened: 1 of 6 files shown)


rukai:
Ah, just realized that window resizing instantly panics with this PR.

On the hello-triangle example:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: PresentError(OutOfDate)', src/backend/direct.rs:886:9
stack backtrace:
  14: <wgpu::SwapChainTexture as core::ops::drop::Drop>::drop
  15: hello_triangle::run::{{closure}}::{{closure}}
  16: winit::platform_impl::platform::x11::EventLoop<T>::run
  17: winit::platform_impl::platform::EventLoop<T>::run
  18: winit::event_loop::EventLoop<T>::run
  19: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
  20: std::thread::local::LocalKey<T>::with
  21: futures_executor::local_pool::block_on
  22: hello_triangle::main

```

rukai:
https://github.com/gfx-rs/wgpu/commit/314074d13c2d674923117d30ce3061bf5599108c#diff-2f51505f30e3a7256be361e9da896d9fR236

This is the change in wgpu-core that is causing the issue, previously queue.present_surface errors were just ignored, but now they are being returned.

So now we have two options:
*   Change wgpu to have the user call swap_chain_present manually rather than on drop, then we can return the error to them.
*   Change wgpu to ignore the error. (This is undesirable, we want the user to be able to immediately detect the frame failed and redraw it)

rukai:
Oh I see, I should expose the same SwapChainError used by get_current_frame.
I'll do that.

rukai:
Oh wait the webgpu spec doesnt even have a swap_chain_present, its done automatically.
I guess we just ignore the errors and the user just gets their frame silently dropped :/

rukai:
I ended up fixing this in wgpu-core without any changes needed in wgpu: https://github.com/gfx-rs/wgpu/pull/819

kvark:
https://github.com/gfx-rs/wgpu/pull/819 is in now, let's update this and land too

rukai:
Should be ready to go now.

PR #453, opened 2020-07-19
Added Static Lifetime to Statically Loaded SPIR-V Modules

Andful:
The commit enables `include_spirv!` to return a `ShaderModuleSource<'static>`.
This allows `ShaderModuleSource` to outlive the context in which `include_spirv!` was called in.
An example where this implementation would work but the previews would not is the following.
```rust
fn get_vertex_module<'a>() -> wgpu::ShaderModuleSource<'a> {
    wgpu::include_spirv!("shader.vert.spv")
}
let vs_module = device.create_shader_module(get_vertex_module());
```
Also it makes logical sense for a statically loaded module to have a static lifetime.  

The only downside that this change might bring is the redundant presence of the binary string (the non aligned binary string and the aligned binary string) but from the produced assembly I could only find one copy of the binary string so I don't think this is the case.

### DIFF
--- a/src/util/mod.rs
+++ b/src/util/mod.rs
@@ -2,14 +2,16 @@
 
 mod belt;
 
+use std::{
+    borrow::Cow,
+    mem::{align_of, size_of},
+    ptr::copy_nonoverlapping,
+};
+
 #[cfg(all(not(target_arch = "wasm32"), feature = "subscriber"))]
 pub use wgc::logging::subscriber::{initialize_default_subscriber, ChromeTracingLayer};
 
 pub use belt::StagingBelt;
-
-/// Wrapper aligning contents to at least 4.
-#[repr(align(4))]
-pub struct WordAligned<Bytes: ?Sized>(pub Bytes);
 
 /// Treat the given byte slice as a SPIR-V module.
 ///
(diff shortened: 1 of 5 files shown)


Andful:
Yes, this can be addressed by switching `struct ShaderModuleDescriptor` to use `Cow` instead of slices.
I made a commit that implements the change that you suggested in case it would be needed.

src/util/mod.rs
```diff
@@ -2,35 +2,31 @@
...
-    assert_eq!(pre, &[], "data offset is not aligned to words!");
-    assert_eq!(post, &[], "data size is not aligned to words!");
-    assert_eq!(
-        words[0], MAGIC_NUMBER,
-        "wrong magic word {:x}. Make sure you are using a binary SPIRV file.",
-        words[0]
-    );
-    super::ShaderModuleSource::SpirV(words)
+    assert_eq!(data.len() % size_of::<u32>(), 0, "data size is not a multiple of 4");
+    let mut words = vec![0u32; data.len()/size_of::<u32>()];
```
kvark, line 26: can we do this only if needed? i.e. if `data` pointer isn't aligned

Andful: Yes, good idea!

Andful:
Do you want me to squash the commits?

kvark:
Yes, please, let's squash

Andful:
I squashed the commits

PR #460, opened 2020-07-20
Replace borrowed slices in Descriptor structs with copy on write slices

Kimundi:
This is the other half of https://github.com/gfx-rs/wgpu/pull/822

### DIFF
--- a/examples/hello-triangle/main.rs
+++ b/examples/hello-triangle/main.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow::Borrowed;
 use winit::{
     event::{Event, WindowEvent},
     event_loop::{ControlFlow, EventLoop},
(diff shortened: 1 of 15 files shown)


examples/hello-triangle/main.rs
```diff
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
```
kvark, line 1: that's something I never did in my projects, but I wonder if it would make sense to go further and do `use std::borrow::Cow::Borrowed` ?

Kimundi:
Just to document this somewhere: The switch to Cow causes certain descriptor expressions to require explicitly named temporary variables. See https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d4f90482467877cf10f9c0b355736d14 for an enumeration of all relevant cases.

src/backend/direct.rs
```diff
@@ -649,9 +650,9 @@ impl crate::Context for Context {
         wgc::gfx_select!(*device => global.device_create_bind_group(
             *device,
             &bm::BindGroupDescriptor {
-                label: desc.label,
+                label: desc.label.as_ref().map(|label| label[..].into()),
```
kvark, line 653: these should be aligned with explicit `Cow::Borrowed` like the other ones

src/backend/web.rs
```diff
@@ -910,8 +910,8 @@ impl crate::Context for Context {
             .collect::<js_sys::Array>();
 
         let mut mapped_desc = web_sys::GpuBindGroupLayoutDescriptor::new(&mapped_bindings);
-        if let Some(label) = desc.label {
-            mapped_desc.label(label);
+        if let Some(label) = &desc.label {
```
kvark, line 913: this looks strange, we should never match on a reference like that. Perhaps, this instead?
```rust
if let Some(label) = desc.label {
  mapped_desc.label(&*label);
}
```

Kimundi: That would cause a move error, the choices are basically this:

```rust
if let Some(label) = &desc.label {
  mapped_desc.label(label);
}
```

Or

```rust
if let Some(ref label) = desc.label {
  mapped_desc.label(label);
}
```

kvark: ok sure `ref label` it is then!

src/backend/direct.rs
Kimundi: I had hoped I wouldn't have to touch these again because they are in the implementation, but ok. 😄

kvark: thank you!

Issue #449, opened 2020-07-19
Documentation about aligning data before mapping it to a buffer


yutannihilation:
(While I asked this on Matrix and my issue was solved, I'm filing this here as I feel it's worth a documentation)

After https://github.com/gfx-rs/wgpu/pull/758, buffers mapped at creation are required to be aligned to 4. But, iiuc (sorry, I'm still a newbie to computer graphics...), vertices are triangular in most cases, which means the length of index data is a multiple of 3. Besides, a index is `u16`, not a multiple of 4 byte. So, it's a common case the user needs to add some padding to the index data by themselves, especially when the index is generated by external libraries (I got this when I was playing with lyon).

Fortunately (or unfortunately?), current examples use only the even number of length of indices, so they don't hit this problem. I think it's good to show how the user can handle the alignment issue.

Not sure if this is useful, but this is what I'm doing:

``` rust
        // record the range before extending
        let stroke_range = 0..(indices.len() as u32);

        // extend
        let alignment = wgpu::COPY_BUFFER_ALIGNMENT as usize / std::mem::size_of::<u16>();
        let fraction = indices.len() % alignment;
        if fraction > 0 {
            indices
                .extend(std::iter::repeat(0).take(alignment - fraction));
        }
```

kvark:
Thank you for the feedback!
How about we do this in `fn create_buffer_with_data`? We know the size of an element there, and we know the alignment, so we could easily request the underlying buffer to be larger, and just handle the extra elements internally.

yutannihilation:
Oh, nice. Sounds good to me!

PR #462, opened 2020-07-21
create_buffer_with_data automatically handles its own alignment requirements

rukai:
closes https://github.com/gfx-rs/wgpu-rs/issues/449

Does write_buffer/write_texture need similar handling?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1233,15 +1233,24 @@ impl Device {
     /// Creates a new buffer, maps it into host-visible memory, copies data from the given slice,
     /// and finally unmaps it, returning a [`Buffer`].
     pub fn create_buffer_with_data(&self, data: &[u8], usage: BufferUsage) -> Buffer {
-        let size = data.len() as u64;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1233,6 +1233,14 @@ impl Device {
     /// Creates a new buffer, maps it into host-visible memory, copies data from the given slice,
     /// and finally unmaps it, returning a [`Buffer`].
     pub fn create_buffer_with_data(&self, data: &[u8], usage: BufferUsage) -> Buffer {
+        let mut data = Cow::Borrowed(data);
```
kvark, line 1236: can we write this without cloning the whole data vector? it may be costy!

kvark:
Let's not worry about write_buffer just yet :)

src/lib.rs
rukai: How about this?
I could do a copy_from_slice from a constant [0; 4] but I dont think its worth the trouble?

I tested the changes on my game (which breaks when alignment isnt handled) and it works fine.

PR #465, opened 2020-07-21
Fix Vulkan portability support on macOS

kvark:


src/lib.rs
```diff
@@ -1036,16 +1036,19 @@ impl Instance {
...
     pub unsafe fn create_surface_from_core_animation_layer(
         &self,
         layer: *mut std::ffi::c_void,
     ) -> Surface {
         let surface = wgc::instance::Surface {
             #[cfg(feature = "vulkan-portability")]
-            vulkan: self.context.instance.vulkan.as_ref().map(|inst| {
-                inst.create_surface_from_layer(layer as *mut _, cfg!(debug_assertions))
-            }),
+            vulkan: None, //TODO: create_surface_from_layer ?
```
cwfitzgerald, line 1051: Is this not used in most applications?

.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         include:
           - os: macos-10.15
             wasm: false
-            CHECK_COMMAND: cargo check --all-targets
+            CHECK_COMMAND: cargo check --all-targets --features vulkan-portability
```
cwfitzgerald, line 22: Should we only be enabling this feature on mac/osx?

src/lib.rs
kvark: This was not compiling, since Vulkan backend doesn't currently have `create_surface_from_layer` at all. We can add it if requested, but for now leaving that dead code was hurting people who would just do `create_surface`

.github/workflows/ci.yml
kvark: Yes, this is the "- os: macos-10.15" section :)

kvark:
bors r=cwfitzgerald

.github/workflows/ci.yml
cwfitzgerald: Oh the diff didn't show me that :)

Issue #458, opened 2020-07-20
Implement Debug for all public types


kvark:
The problem is that every type stores `Arc<Context>`, so we need to have some kind of manual implementations:
  - either for `Context`, which is something very basic, like `enum Context { Native, Web }`
  - for each type separately, ignoring the context field

kvark:
Fixed by #466

PR #466, opened 2020-07-21
add debug trait for public types

Andful:
This pull request is to resolve the issue #458.
This pull request is still incomplete. The only problematic trait is `RenderBundleEncoder`.
For the native version, `RenderBundleEncoder` does not implement `Debug`. This would have to be implemented in the wgpu project.
 FYI, Context does not need to implement `Debug`. The type that implements `Context` is used directly, i.e. `Arc<C>` is used and not `Arc<dyn Context>`

### DIFF
--- a/src/backend/direct.rs
+++ b/src/backend/direct.rs
@@ -12,7 +12,43 @@ use smallvec::SmallVec;
 use std::{ffi::CString, fmt, marker::PhantomData, ops::Range, ptr, slice};
 use typed_arena::Arena;
 
-pub struct Context(pub wgc::hub::Global<wgc::hub::IdentityManagerFactory>);
+pub struct Context(wgc::hub::Global<wgc::hub::IdentityManagerFactory>);
+
+impl Context {
+    pub fn adapter_get_info(&self, id: wgc::id::AdapterId) -> wgc::instance::AdapterInfo {
+        let global = &self.0;
+        wgc::gfx_select!(id => global.adapter_get_info(id))
+    }
+
+    pub fn enumerate_adapters(&self, backends: wgt::BackendBit) -> Vec<wgc::id::AdapterId> {
+        self.0
+            .enumerate_adapters(wgc::instance::AdapterInputs::Mask(backends, |_| {
+                PhantomData
+            }))
+    }
+
+    #[cfg(any(target_os = "ios", target_os = "macos"))]
+    pub unsafe fn create_surface_from_core_animation_layer(
+        &self,
+        layer: *mut std::ffi::c_void,
+    ) -> crate::Surface {
+        let surface = wgc::instance::Surface {
+            #[cfg(feature = "vulkan-portability")]
+            vulkan: None, //TODO: create_surface_from_layer ?
+            metal: self.0.instance.metal.as_ref().map(|inst| {
+                inst.create_surface_from_layer(layer as *mut _, cfg!(debug_assertions))
+            }),
+        };
+
+        crate::Surface {
+            id: self.0.surfaces.register_identity(
+                PhantomData,
+                surface,
+                &mut wgc::hub::Token::root(),
+            ),
+        }
+    }
+}
 
 impl fmt::Debug for Context {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
--- a/src/backend/direct.rs
+++ b/src/backend/direct.rs
@@ -9,26 +9,16 @@ use crate::{
 use arrayvec::ArrayVec;
 use futures::future::{ready, Ready};
 use smallvec::SmallVec;
-use std::{ffi::CString, marker::PhantomData, ops::Range, ptr, slice};
+use std::{ffi::CString, fmt, marker::PhantomData, ops::Range, ptr, slice};
 use typed_arena::Arena;
 
-macro_rules! gfx_select {
-    ($id:expr => $global:ident.$method:ident( $($param:expr),+ )) => {
-        match $id.backend() {
-            #[cfg(any(not(any(target_os = "ios", target_os = "macos")), feature = "vulkan-portability"))]
-            wgt::Backend::Vulkan => $global.$method::<wgc::backend::Vulkan>( $($param),+ ),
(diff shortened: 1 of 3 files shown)


kvark:
> FYI, Context does not need to implement Debug. The type that implements Context is used directly, i.e. Arc<C> is used and not Arc<dyn Context>

We don't want the contents of the `Context` to be printed whenever anything in this context is printed.

Andful:
That would mean I have to implement `Debug` for every type and cannot use `#[derive(Debug)]`. I can implement `Debug` only for types defined in the wgpu-rs crate and not in the wgpu crate i.e. I cannot implement Debug for direct::Context.

kvark:
We have choices:
  - add `Debug` upstream to `wgpu-core`
  - wrap the `wgpu` thing around a newtype in wgpu-rs and implement `Debug` for it
  - custom-implement `Debug` for the types in wgpu-rs

Andful:
I think it is a good idea to implement `Debug` for `RenderBundleEncoder` upstream to `wgpu-core` but I would personally not know how I would feel about implementing a custom `Debug` to `direct::Context` upstream to `wgpu-core`. To me it does not seem right to make a change in `wgpu-core` which might negatively affect its development experience for a feature in `wgpu-rs`.

Regarding the second option, I like that option and I think it will require less overall code to be achieved and also very scalable. A change in a type will not result in needing a change in the `Debug` implementation.

I personally think the best option would be to implement `Debug` for `RenderBundleEncoder` and wrap `direct::Context` and implement `Debug` in the wrapper.

src/lib.rs
```diff
@@ -147,27 +147,27 @@ trait RenderPassInner<Ctx: Context>: RenderInner<Ctx> {
...
+    type BufferId: Debug + Send + Sync + 'static;
+    type TextureId: Debug + Send + Sync + 'static;
+    type PipelineLayoutId: Debug + Send + Sync + 'static;
+    type RenderPipelineId: Debug + Send + Sync + 'static;
+    type ComputePipelineId: Debug + Send + Sync + 'static;
+    type CommandEncoderId: Debug;
+    type ComputePassId: Debug + ComputePassInner<Self>;
+    type RenderPassId: Debug + RenderPassInner<Self>;
+    type CommandBufferId: Debug + Send + Sync;
     type RenderBundleEncoderId: RenderInner<Self>;
```
Andful, line 167: Changes is `wgpu-core` are needed for `direct::Context::RenderBundleEncoderId` to implement `Debug`

src/backend/web.rs
```diff
@@ -23,14 +24,26 @@ use wasm_bindgen::prelude::*;
...
+
+#[derive(Clone, Debug)]
 pub(crate) struct Sendable<T>(T);
 unsafe impl<T> Send for Sendable<T> {}
 unsafe impl<T> Sync for Sendable<T> {}
 
-pub(crate) type Context = Sendable<web_sys::Gpu>;
+pub(crate) struct Context(web_sys::Gpu);
+unsafe impl Send for Context {}
+unsafe impl Sync for Context {}
```
Andful, line 36: Is this necessary?

src/backend/direct.rs
```diff
@@ -9,26 +9,16 @@ use crate::{
...
-            wgt::Backend::Metal => $global.$method::<wgc::backend::Metal>( $($param),+ ),
-            #[cfg(windows)]
-            wgt::Backend::Dx12 => $global.$method::<wgc::backend::Dx12>( $($param),+ ),
-            #[cfg(windows)]
-            wgt::Backend::Dx11 => $global.$method::<wgc::backend::Dx11>( $($param),+ ),
-            _ => unreachable!()
-        }
-    };
-}
+pub struct Context(pub wgc::hub::Global<wgc::hub::IdentityManagerFactory>);
```
kvark, line 15: does it need to have `pub` contents?

kvark:
Thank you!

> I personally think the best option would be to implement Debug for RenderBundleEncoder upstream in wgpu-core and wrap direct::Context and implement Debug in the wrapper.

That's reasonable, let's proceed with this. Would you be interested in making that upstream PR? Could be as simple as `derive(Debug)` :)

src/backend/web.rs
kvark: Good question! It was introduced in #272, which appears to be focused on associated types being Send/Sync. Maybe the context doesn't need it? cc @kyren 
Let's remove this Send/Sync from the Context for now and see?

kyren: It might not be necessary since `Context` is strictly an internal type?  If it's not visible at all to the user and anything that contains `Context` is itself forcibly made `Send + Sync` then I guess It shouldn't matter?  I don't remember specifically but I probably made `Context` Send just to follow the pattern of forcibly making the all the web_sys handle types `Send + Sync` and focusing my effort from the bottom up, but I'd 100% believe that it might not have been actually necessary.

Andful:
I made a pull request to `wgpu-core` to make `RenderBundleEncoder` implement `Debug`. Is this pull request going to be accepted before the changes are made to `wgpu-core` or should I wait the changes in `wgpu-core` to occur and make the last changes in this pull request?
 
`Instance.request_adapter` requires `Context` to be `Send + Sync` therefore I made `crate::Context` require  `Send + Sync`.

Regarding making the fields of `direct::Context` private, it requires some moving of code and I would like your confirmation before including that change (currently `Adapter.get_info`, `Instance.enumerate_adapters` and `Instance.create_surface_from_core_animation_layer`). I.e. tell me if I should include [this commit](https://github.com/gfx-rs/wgpu-rs/pull/466/commits/019e7513b3401fad6a8895fd7400fb4486140a07)

kvark:
@Andful thank you for the careful consideration!
>  Is this pull request going to be accepted before the changes are made to wgpu-core or should I wait the changes in wgpu-core to occur and make the last changes in this pull request?

It's accepted, but you are going to be blocked on #460 if you try to use it. So I think we should merge this first, and then follow-up with picking up the render bundle changes.

> I.e. tell me if I should include this commit.

Even though it adds a bit of code, I think it makes sense: it encapsulates all the "direct" logic in the direct module. So I think we should go for it.

PR #468, opened 2020-07-23
Expose Buffer::usage information

DasEtwas:


src/lib.rs
```diff
@@ -1438,6 +1438,11 @@ impl Buffer {
         self.map_context.lock().reset();
         Context::buffer_unmap(&*self.context, &self.id);
     }
+    
+    /// Returns the BufferUsage supplied to the BufferDescriptor upon creating this Buffer.
```
kvark, line 1442: We can't really (and we don't want to) expose *all* the information that was passed by the user in `BufferDescriptor`. So if we start exposing some, like this one, the question arises: where do we draw the line on what we expose and what not?
Currently, the logic is simple: we don't expose anything the user would otherwise already know.

DasEtwas: In my use case the solution would be to store the supplied BufferUsage alongside the Buffer, which wouldn't be as elegant as this (and a waste of space).

kvark: Right. This is OK though. The space consideration is really minor here, and the inconvenience is still there for other things, like the buffer size.
For us, not exposing this means that we can freely consider this field an implementation detail, and we'd like to keep it this way.

kvark:
Going to close this as the change is undesired in wgpu-rs, and there is a clear path for the users to get that if they need.

PR #471, opened 2020-07-23
Change deployment target to wgpu-rs.github.io

kvark:
Using "gh-pages" branch for the website is not idiomatic to Git. This PR changes the deployment to https://github.com/gfx-rs/wgpu-rs.github.io instead


kvark:
I guess the only way to test this is by merging.
I've switched the source of `wgpu.rs` already. Now we just need to make sure that this deployment works.

PR #473, opened 2020-07-24
Update wgpu, use defaults for rasterizer states, use depth-clamping

kvark:
Includes https://github.com/gfx-rs/wgpu/pull/833 last


kvark:
bors r=straightforward

PR #476, opened 2020-07-24
Convert to tracing for logging and add fmt logger

cwfitzgerald:
wgpu-rs half of https://github.com/gfx-rs/wgpu/pull/835

### DIFF
--- a/examples/framework.rs
+++ b/examples/framework.rs
@@ -72,8 +72,13 @@ struct Setup {
 }
 
 async fn setup<E: Example>(title: &str) -> Setup {
-    #[cfg(not(target_arch = "wasm32"))]
-    env_logger::init();
+    #[cfg(feature = "subscriber")]
+    {
+        let chrome_tracing_dir = std::env::var("WGPU_CHROME_TRACE");
+        wgpu::util::initialize_default_subscriber(
+            chrome_tracing_dir.as_ref().map(std::path::Path::new).ok(),
+        );
+    };
 
     #[cfg(target_arch = "wasm32")]
     console_log::init().expect("could not initialize logger");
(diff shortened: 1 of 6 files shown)


examples/framework.rs
```diff
@@ -69,8 +69,13 @@ struct Setup {
 }
 
 async fn setup<E: Example>(title: &str) -> Setup {
-    #[cfg(not(target_arch = "wasm32"))]
-    env_logger::init();
+    #[cfg(feature = "subscriber")]
```
kvark, line 72: I think the examples should enable this feature in `wgpu-core` unconditionally

src/util/mod.rs
```diff
@@ -35,8 +35,8 @@ pub fn make_spirv<'a>(data: &'a [u8]) -> super::ShaderModuleSource<'a> {
     // otherwise copy the byte array in an owned vector and use that instead.
     let words = if data.as_ptr().align_offset(align_of::<u32>()) == 0 {
         let (pre, words, post) = unsafe { data.align_to::<u32>() };
-        debug_assert_eq!(pre, &[]);
-        debug_assert_eq!(post, &[]);
+        debug_assert!(pre.is_empty());
```
kvark, line 38: what's happening here?

examples/framework.rs
cwfitzgerald: I was doing this to try to avoid having three different places to update the ref, but I can do that.

src/util/mod.rs
cwfitzgerald: This change is the same change that #479 addressed. Accidentally did it twice.

cwfitzgerald:
Alright, updated with our conversations on the matrix in considerations.

cwfitzgerald:
Updated refs, seems we're good to go.
bors r=kvark

PR #479, opened 2020-07-24
Split framework limit situation into requested/required

cwfitzgerald:
Requiring individual examples be responsible for panicing if their features aren't supported is a bit bug-prone, so this encodes requirements in the framework and enforces it in the framework.

Additionally made the verbs consistent.

### DIFF
--- a/examples/framework.rs
+++ b/examples/framework.rs
@@ -30,10 +30,13 @@ pub enum ShaderStage {
 }
 
 pub trait Example: 'static + Sized {
-    fn needed_features() -> wgpu::Features {
+    fn optional_features() -> wgpu::Features {
         wgpu::Features::empty()
     }
-    fn needed_limits() -> wgpu::Limits {
+    fn required_features() -> wgpu::Features {
+        wgpu::Features::empty()
+    }
+    fn required_limits() -> wgpu::Limits {
         wgpu::Limits::default()
     }
     fn init(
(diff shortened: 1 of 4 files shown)


examples/framework.rs
```diff
@@ -30,10 +30,13 @@ pub enum ShaderStage {
 }
 
 pub trait Example: 'static + Sized {
-    fn needed_features() -> wgpu::Features {
+    fn requested_features() -> wgpu::Features {
```
kvark, line 33: maybe have it the other way, and call `optional_features`? This would let examples avoiding to repeat themselves if they need to implement both functions

cwfitzgerald:
Addressed. Also addressed that weird type error when dealing with features on. Need to write up a rustc bug report...

src/util/mod.rs
```diff
@@ -35,8 +35,8 @@ pub fn make_spirv<'a>(data: &'a [u8]) -> super::ShaderModuleSource<'a> {
     // otherwise copy the byte array in an owned vector and use that instead.
     let words = if data.as_ptr().align_offset(align_of::<u32>()) == 0 {
         let (pre, words, post) = unsafe { data.align_to::<u32>() };
-        debug_assert_eq!(pre, &[]);
-        debug_assert_eq!(post, &[]);
+        debug_assert!(pre.is_empty());
```
kvark, line 38: why this change?

cwfitzgerald: Remember that issue I had like 2 weeks ago or so, where if you build with the subscriber feature this line of code fails to typecheck? That apparently made it into master, so this fixes this. We need to add the subscriber feature to ci.

PR #481, opened 2020-07-25
Remove screenshots from README

kvark:
Screenshots cover a ton of space, and now we have a dedicated hosted gallery for them.
I'm open to the idea of having *some* smaller screenshots here. It's just difficult to draw the line. Maybe README should show a few screenshots of examples only, and the gallery would not have them?


kvark:
ok, let's proceed and rethink this
bors r=grovesNL

cwfitzgerald:
As an extension to this, we might consider posting screenshots that are hosted by the gallery to prevent them from taking up space here.

PR #482, opened 2020-07-25
Add label parameter to `create_buffer_with_data`

OptimisticPeach:
Improves QOL by allowing you to attach a label to a buffer when created with predetermined data:
```
let uniform_buf = device.create_buffer_with_data(
    bytemuck::cast_slice(my_data),
    wgpu::BufferUsage::UNIFORM,
    Some(Cow::Borrowed("My Uniforms")),
);
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1241,32 +1241,6 @@ impl Device {
         }
     }
 
-    /// Creates a new buffer, maps it into host-visible memory, copies data from the given slice,
-    /// and finally unmaps it, returning a [`Buffer`].
-    pub fn create_buffer_with_data(&self, data: &[u8], usage: BufferUsage) -> Buffer {
-        let align = COPY_BUFFER_ALIGNMENT;
-        let unpadded_size = data.len() as u64;
-        let padding = (align - unpadded_size % align) % align;
-        let padded_size = unpadded_size + padding;
-
-        let buffer = self.create_buffer(&BufferDescriptor {
-            label: None,
-            size: padded_size,
-            usage,
-            mapped_at_creation: true,
-        });
-
-        let range =
-            Context::buffer_get_mapped_range_mut(&*self.context, &buffer.id, 0..padded_size);
-        range[0..unpadded_size as usize].copy_from_slice(&data);
-        for i in unpadded_size..padded_size {
-            range[i as usize] = 0;
-        }
-
-        buffer.unmap();
-        buffer
-    }
-
     /// Creates a new [`Texture`].
     ///
     /// `desc` specifies the general format of the texture.
--- a/src/util/mod.rs
+++ b/src/util/mod.rs
@@ -12,6 +12,7 @@ use std::{
 pub use wgc::logging::subscriber::{initialize_default_subscriber, ChromeTracingLayer};
 
 pub use belt::StagingBelt;
+use std::sync::Arc;
 
 /// Treat the given byte slice as a SPIR-V module.
 ///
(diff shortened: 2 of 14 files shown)


grovesNL:
I wonder if `create_buffer_with_data` would be doing too much by automatically padding and including labels. Originally the idea was to keep it as a small convenience function to handle the copy only, so we decided not to include labels here in the a while ago (in https://github.com/gfx-rs/wgpu-rs/pull/222#discussion_r399742905).

> My biggest concern here is seeing `None` on the argument list is bad

FWIW some IDEs are starting to annotate arguments with the argument name now (I think rust-analyzer and CLion do this for Rust, maybe some others).

src/lib.rs
```diff
@@ -513,6 +513,49 @@ pub struct BufferSlice<'a> {
     size: Option<BufferSize>,
 }
 
+/// Describes the contents of a [`Buffer`] at creation.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub enum BufferInitContents<'a> {
+    /// Allocate memory, but do not provide values for the contents.
+    Allocation(BufferAddress),
```
kvark, line 520: let's have this as `Allocation { size: BufferAddress, mapped: bool }`, since `mapped_at_creation` would otherwise make no sense for the `CopyData` variant

src/lib.rs
```diff
@@ -513,6 +513,49 @@ pub struct BufferSlice<'a> {
...
+    /// Copy the values from the slice into the [`Buffer`] on creation.
+    CopyData(&'a [u8]),
+}
+
+/// Describes a [`Buffer`] when allocating.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct BufferInitDescriptor<'a, 'data> {
```
kvark, line 527: Let's just call it `BufferDescriptor`

src/lib.rs
```diff
@@ -513,6 +513,49 @@ pub struct BufferSlice<'a> {
     size: Option<BufferSize>,
 }
 
+/// Describes the contents of a [`Buffer`] at creation.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub enum BufferInitContents<'a> {
```
kvark, line 518: Let's call it `BufferContents` or `BufferMemory`

src/lib.rs
```diff
@@ -1227,44 +1266,47 @@ impl Device {
...
-        });
-
-        let range =
-            Context::buffer_get_mapped_range_mut(&*self.context, &buffer.id, 0..padded_size);
-        range[0..unpadded_size as usize].copy_from_slice(&data);
-        for i in unpadded_size..padded_size {
-            range[i as usize] = 0;
+    /// Creates a [`Buffer`] with or without data to initialize it with.
+    pub fn create_buffer(&self, descriptor: &BufferInitDescriptor<'_, '_>) -> Buffer {
+        let wgt_descriptor = descriptor.to_buffer_descriptor();
```
kvark, line 1271: do we need `to_buffer_descriptor` at all? I think we end up matching on `descriptor.contents` the whole 3 times in the process, including this function!
Can we perhaps just do a single match in here without having a separate `to_buffer_descriptor` function?

grovesNL:
@kvark :+1: it seems ok to me. I think we could probably omit labels from most of the examples once we're using builders though.

I have a slight concern about divergence from the JS API because it might make it difficult to translate between JS and Rust examples. For example, the `CopyData` helper doesn't exist in the upstream spec and applies padding automatically. Maybe we could try to keep helpers like the staging belt and `CopyData` separate from the main API when possible? I don't think it's too bad in this case, but in general we might want to be careful about diverging.

kvark:
Thank you for feedback, @grovesNL , it makes a lot of sense to me. How do you see it exposed as a helper? Should there be a standalone `create_buffer_with_data(device: &Device, desc: &BufferInitDescriptor)` function in the `util` module, or something else?

grovesNL:
@kvark yeah exactly – moving `create_buffer_with_data` into `util` should make it more obvious that it's an optional helper (e.g. users are able to create their own custom version of it if they'd like)

kvark:
@OptimisticPeach sorry about getting you to do the other thing 😓 . Would you be open to try adding this method to `util` module, and stripping `wgpu` main API from this magic?

src/util/mod.rs
```diff
@@ -53,3 +54,95 @@ pub fn make_spirv<'a>(data: &'a [u8]) -> super::ShaderModuleSource<'a> {
...
+    CopyData(&'a [u8]),
+}
+
+/// Describes a [`Buffer`] when allocating.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct BufferInitDescriptor<'a, 'data> {
+    /// Debug label of a buffer. This will show up in graphics debuggers for easy identification.
+    pub label: Option<Cow<'a, str>>,
+    /// Contents of a buffer on creation.
+    pub contents: BufferContents<'data>,
```
kvark, line 143: Sorry for not being clear about this: we want `data: &'a [u8]` here instead of the contents enum. If the user is not providing the data, they should use the original `Device::create_buffer`.

src/util/mod.rs
```diff
@@ -53,3 +54,95 @@ pub fn make_spirv<'a>(data: &'a [u8]) -> super::ShaderModuleSource<'a> {
...
+    },
+    /// Copy the values from the slice into the [`Buffer`] on creation.
+    CopyData(&'a [u8]),
+}
+
+/// Describes a [`Buffer`] when allocating.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct BufferInitDescriptor<'a, 'data> {
+    /// Debug label of a buffer. This will show up in graphics debuggers for easy identification.
+    pub label: Option<Cow<'a, str>>,
```
kvark, line 141: also, this could be `Option<&'a str>`, and I think it should be.
The cows make most sense for things coming from `wgpu`, while this one doesn't.

src/util/mod.rs
```diff
@@ -53,3 +54,57 @@ pub fn make_spirv<'a>(data: &'a [u8]) -> super::ShaderModuleSource<'a> {
...
+        }
+
+        buffer.unmap();
+        buffer
+    }
+}
+
+/// Describes a [`Buffer`] when allocating.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct BufferInitDescriptor<'a, 'data> {
```
kvark, line 101: sorry for being pendantic, please use a single `'a` lifetime, no need to have two different lifetimes here

OptimisticPeach: Those lifetimes may be disjoint though, wouldn't it be more restrictive to only have one?

PR #484, opened 2020-07-26
Enable features in CI to prevent compiler errors

cwfitzgerald:
Simple change, this also builds features in CI.

I added trace and replay features to wasm, even though wasm doesn't support tracing as is, because people may enable them to get serde support for wgpu-types stuff.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,18 +19,18 @@ jobs:
         include:
           - os: macos-10.15
             wasm: false
-            CHECK_COMMAND: cargo check --all-targets --features vulkan-portability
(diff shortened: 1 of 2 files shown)


cwfitzgerald:
I think for now min/max is probably okay as we don't really have intersecting features. I didn't use `--all-features` because it is my impression that vulkan-portability is only supposed to be used on mac.

kvark:
I guess it's a no-op to enable Vulkan Portability on other platforms, so it's ok.

cwfitzgerald:
One of the `#[cfg(feature = subscriber)]` guards weren't behind a platform guard as well. Everything should work fine now.
bors retry

Issue #478, opened 2020-07-24
[meta] Solving the cow-gonomics


kvark:
(short for `Cow` API ergonomics)
With #460 and #477, our API is updated to use `Cow` in places we had borrowing of `[]` and `str`. This is needed for `wgpu` mostly, to have a single struct declaration used in different parts (`wgpu-core`, `wgpu-types`, `player`). But now it affects the user-facing API, and it hurts the ergonomics: `Cow::Borrowed(&[...])` is longer than `&[...]`.

We have the following proposals to improve this:
  1. do nothing, leave `Cow` in the API
  2. #470 - wrap it in a builder pattern
  3. redefine structures in wgpu-rs with just plain slices `&[..]`.

Both (2) and (3) have the benefit of making `wgpu-types` types to not be generic any more. But they involve more code, obviously.

zakarumych:
I'm not have much of a context. But both '&[T]' and 'Vec<T>' implement 'Into<Cow<[T]>'

kvark:
We've been discussing the problem with `Cow` in the API on the matrix, and we realize the urgency of this change, since it affects a lot of code. Current consensus is converging around #470 with a custom trait instead of `Into<Cow<>>` bound.

Another choice to make is:
  a) totally hide `wgpu-types` structures in wgpu-rs behind the new builders
  b) just add builder methods to `wgpu-types` structs

With (a) we can make `wgpu-types` non-generic and lighter, but we'll end up with more code in total.
It seems like we are going to try (b) first, @cwfitzgerald was about to start experimenting on that.

kvark:
Good news is, @Andful implemented the builders in `wgpu-types` 🎉 
Bad news is, we are now reconsidering if the builders are truly the way to go in `wgpu-rs`. We want to try out redefining plain structures and see what comes out of it. If anybody wants to help, let's talk on the matrix!

kvark:
Fixed by #501

Issue #470, opened 2020-07-23
Proposal: Introduce builder pattern


MatusT:
This proposal argues for introduction of the builder pattern for most of the structs of the public API. It is a reaction to the recent introduction of the usage of `Cow` in the API which results in quite a lot of noise.

Consider the following example:
```Rust
let pipeline_layout = device.create_pipeline_layout(&PipelineLayoutDescriptor {
  bind_group_layouts: Borrowed(&[&camera_bind_group_layout, &per_molecule_bind_group_layout]),
  push_constant_ranges: Borrowed(&[]),
});
```
versus
```Rust
let pipeline_layout = device.create_pipeline_layout(PipelineLayoutDescriptor::new()
    .bind_group_layouts(&[&camera_bind_group_layout, &per_molecule_bind_group_layout])
); 
```

Notice several things:
1. With builder pattern there can be some reasonable defaults, like in the case of **push_constant_ranges**. In contrast to just introducing default implementation everywhere, builder pattern doesn't introduce noise with ``` ... Default::default() ```.
2. All of the ```Borrowed``` can be omitted. This is taken from article https://jwilm.io/blog/from-str-to-cow/ and would be done by annotating setter functions with ```Input: Into<Cow<ActualInput>```.
3. We can get rid of **&** at the beginning by implementing Deref (see how Ash does It) and also **{ }** parantheses.

All of this creates, in my opinion: cleaner, more readable, easily writable(especially Borrowed could be confusing for beginners to Rust) WebGPU code. 

This builder pattern is inspired by the Ash crate.

kvark:
Thank you for filing this!
Benefits:
- `wgpu-types` struct no longer need to be generic (since wgpu-rs API would no longer try to reuse them)
- good ability to add fields later without breaking the API
- better ergonomics

Downsides: more code

Kimundi:
Note that that setter API is not actually going to work: If you accept a `Input: Into<Cow<'a, [T]>`, then passing an `&[T; N]` argument will not do a deref coercion. That is in fact why we used `Borrowed()` there: If you tried to use `From` or `Into`, you would have to explicitly slice the array literal with `[..]` as well.

See https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=976b42c8df8ac0103044d1b8e86860df

Andful:
[...]

PR #485, opened 2020-07-26
Use builder for TextureDescriptor

cwfitzgerald:
This isn't meant to be merged as is, but this demos one usage of the new builders.

#478 #470 https://github.com/gfx-rs/wgpu/pull/840

### DIFF
--- a/examples/cube/main.rs
+++ b/examples/cube/main.rs
@@ -173,15 +173,17 @@ impl framework::Example for Example {
             height: size,
             depth: 1,
         };
-        let texture = device.create_texture(&wgpu::TextureDescriptor {
-            label: None,
-            size: texture_extent,
-            mip_level_count: 1,
-            sample_count: 1,
-            dimension: wgpu::TextureDimension::D2,
-            format: wgpu::TextureFormat::Rgba8UnormSrgb,
-            usage: wgpu::TextureUsage::SAMPLED | wgpu::TextureUsage::COPY_DST,
-        });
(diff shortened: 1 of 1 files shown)


examples/cube/main.rs
```diff
@@ -173,15 +173,17 @@ impl framework::Example for Example {
...
-            usage: wgpu::TextureUsage::SAMPLED | wgpu::TextureUsage::COPY_DST,
-        });
+        let texture = device.create_texture(
+            wgpu::TextureDescriptor::new(
+                texture_extent,
+                wgpu::TextureDimension::D2,
+                wgpu::TextureFormat::Rgba8UnormSrgb,
+                wgpu::TextureUsage::SAMPLED | wgpu::TextureUsage::COPY_DST,
+            )
+            .label("label")
```
kvark, line 183: Is rustfmt really formatting it this way?

cwfitzgerald: Yeah. Would you've formatted it differently?

kvark: Typically if you chain calls, they get aligned with indentation relative to the init line, i.e.
```rust
something
  .do_something()
  .and_something()
```

In our case, it looks like there is no indentation, so I'm surprised

multun: that's the expected behavior inside parenthesized expressions I think

cwfitzgerald:
Closing as it has served its purpose.

PR #490, opened 2020-07-29
Panic using error Display rather than unwrapping

### LINKED ISSUE #489: Descriptive panics on errors
We are currently unwrapping all errors returned by `wgpu-core`, see https://github.com/gfx-rs/wgpu/issues/638. We should instead panic with the actual description of the error.

cormac-obrien:
Fixes #489.

### DIFF
--- a/src/backend/direct.rs
+++ b/src/backend/direct.rs
@@ -1,15 +1,15 @@
 use crate::{
-    backend::native_gpu_future, BindGroupDescriptor, BindGroupLayoutDescriptor, BindingResource,
-    BufferDescriptor, CommandEncoderDescriptor, ComputePipelineDescriptor, Features, Limits,
-    LoadOp, MapMode, Operations, PipelineLayoutDescriptor, RenderPipelineDescriptor,
-    SamplerDescriptor, ShaderModuleSource, SwapChainStatus, TextureDescriptor,
-    TextureViewDescriptor,
+    backend::native_gpu_future, BindGroupDescriptor, BindGroupLayoutDescriptor,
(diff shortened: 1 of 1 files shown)


src/backend/direct.rs
```diff
@@ -546,7 +546,12 @@ impl crate::Context for Context {
         trace_dir: Option<&std::path::Path>,
     ) -> Self::RequestDeviceFuture {
         let global = &self.0;
-        let device_id = wgc::gfx_select!(*adapter => global.adapter_request_device(*adapter, desc, trace_dir, PhantomData)).unwrap();
+        let device_id = match wgc::gfx_select!(
+            *adapter => global.adapter_request_device(*adapter, desc, trace_dir, PhantomData)
+        ) {
+            Ok(d) => d,
+            Err(e) => panic!("{}", e),
+        };
```
GabrielMajeri, line 554: Since this pattern is repeated in a lot of places, would it be possible to use a generic function? Or use a macro?

cwfitzgerald: Discussed in the matrix that a function like `fn on_error(err: impl Error) { panic!("{}", err) )` and calling `.unwrap_or_else(on_error)` is the desired way to go.

src/lib.rs
```diff
@@ -2338,3 +2338,7 @@ impl SwapChain {
         }
     }
 }
+
+fn on_error<T>(err: impl Error) -> T {
```
kvark, line 2342: since it's only used inside `direct.rs`, let's move it there

PR #493, opened 2020-07-30
Use BCn textures if possible in skybox example

cwfitzgerald:
This converts the skybox to using compressed textures if available. I have designed the code to be extensible to other compressed formats as they are added.

Additionally I added srgb to the texture types, as we were using a srgb framebuffer without properly converting into  linear on the texture reads.


examples/skybox/main.rs
```diff
@@ -79,13 +83,11 @@ impl framework::Example for Skybox {
...
         let aspect = sc_desc.width as f32 / sc_desc.height as f32;
         let uniforms = Self::generate_uniforms(aspect);
-        let uniform_buf = device.create_buffer_init(
-            &wgpu::util::BufferInitDescriptor {
-                label: Some("Uniform Buffer"),
-                contents: bytemuck::cast_slice(&raw_uniforms(&uniforms)),
-                usage: wgpu::BufferUsage::UNIFORM | wgpu::BufferUsage::COPY_DST,
-            }
-        );
+        let uniform_buf = device.create_buffer_init(&wgpu::util::BufferInitDescriptor {
```
kvark, line 86: Looks like we have different versions of the formatter in play. Don't know what the last version was used here, but please check if you can update yours and it gives the same result.

cwfitzgerald: I think my stable is a version behind (my main proj is nightly) so I'll make sure everything is updated.

PR #497, opened 2020-08-06
gecko: Update to latest web-sys

grovesNL:
The Cargo.toml patches are temporary until a new version of web-sys is published.

Once a new version of web-sys is published I'll cherry-pick this commit over to `master` too.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,6 +74,10 @@ test = true
 #gfx-backend-dx11 = { version = "0.5.0", path = "../gfx/src/backend/dx11" }
 #gfx-descriptor = { version = "0.1.0", path = "../gfx-extras/gfx-descriptor" }
 #gfx-memory = { version = "0.1.0", path = "../gfx-extras/gfx-memory" }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -74,6 +74,10 @@ test = true
 #gfx-backend-dx11 = { version = "0.5.0", path = "../gfx/src/backend/dx11" }
 #gfx-descriptor = { version = "0.1.0", path = "../gfx-extras/gfx-descriptor" }
 #gfx-memory = { version = "0.1.0", path = "../gfx-extras/gfx-memory" }
+wasm-bindgen = { git = "https://github.com/rustwasm/wasm-bindgen" }
```
kvark, line 77: shouldn't we pick a concrete revision? Otherwise, I'm not sure if Cargo is even obligated to grab the latest git, it can still use some older revisions for some users.

grovesNL: Yeah definitely, we should include `rev`

PR #501, opened 2020-08-10
Fighting back the bovine forces

Kimundi:
"Good work boys, we'll get 'em next time!"

### DIFF
--- a/src/backend/direct.rs
+++ b/src/backend/direct.rs
@@ -1,21 +1,17 @@
 use crate::{
     backend::native_gpu_future, BindGroupDescriptor, BindGroupLayoutDescriptor, BindingResource,
     CommandEncoderDescriptor, ComputePipelineDescriptor, Features, Limits, LoadOp, MapMode,
-    Operations, PipelineLayoutDescriptor, RenderPipelineDescriptor, SamplerDescriptor,
-    ShaderModuleSource, SwapChainStatus, TextureDescriptor, TextureViewDescriptor,
+    Operations, PipelineLayoutDescriptor, RenderBundleEncoderDescriptor, RenderPipelineDescriptor,
+    SamplerDescriptor, ShaderModuleSource, SwapChainStatus, TextureDescriptor,
+    TextureViewDescriptor,
 };
 
 use arrayvec::ArrayVec;
 use futures::future::{ready, Ready};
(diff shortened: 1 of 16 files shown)


src/backend/direct.rs
```diff
@@ -723,9 +722,20 @@ impl crate::Context for Context {
                     module: fs.module.id,
                     entry_point: Borrowed(&fs.entry_point),
                 });
+        // TODO: How to avoid this allocation?
```
kvark, line 725: use `ArrayVec<[_; wgc::device::MAX_VERTEX_BUFFERS]>` for the type

src/backend/direct.rs
```diff
@@ -723,9 +722,20 @@ impl crate::Context for Context {
...
+            .map(|vertex_buffer| wgt::VertexBufferDescriptor {
+                stride: vertex_buffer.stride,
+                step_mode: vertex_buffer.step_mode,
+                attributes: Borrowed(vertex_buffer.attributes),
+            })
+            .collect();
         let vertex_state = wgt::VertexStateDescriptor {
             index_format: desc.vertex_state.index_format,
-            vertex_buffers: Borrowed(&desc.vertex_state.vertex_buffers),
+            vertex_buffers: vertex_buffers.into(),
```
kvark, line 738: And here we should be more explicit, i.e. `Borrowed(&vertex_buffers)`

src/lib.rs
```diff
@@ -889,64 +889,218 @@ pub use wgt::RequestAdapterOptions as RequestAdapterOptionsBase;
...
+    /// How to filter between mip map levels
+    pub mipmap_filter: FilterMode,
+    /// Minimum level of detail (i.e. mip level) to use
+    pub lod_min_clamp: f32,
+    /// Maximum level of detail (i.e. mip level) to use
+    pub lod_max_clamp: f32,
+    /// If this is enabled, this is a comparison sampler using the given comparison function.
+    pub compare: Option<CompareFunction>,
+    /// Valid values: 1, 2, 4, 8, and 16.
+    pub anisotropy_clamp: Option<u8>,
```
kvark, line 1007: this would probably make more sense to make `NonZeroU8`, since any non-zero value can be more or less ignored by the implementation, but 0 makes very little sense.

PR #508, opened 2020-08-16
Run event loop outside async function

### LINKED ISSUE #454: Examples: The winit::EventLoop should not be run in wasm_bindgen_futures::spawn_local
Running the event loop inside of an async function in spawn_local appears to block new JS promises from being run if they are created in an event callback:

https://github.com/rustwasm/wasm-bindgen/issues/2246

I am not sure exactly how to restructure the examples, but figured I would post this here in case anyone knows the right way to do this or comes across the same issue as I did.

### LINKED ISSUE #1457: Examples: The winit::EventLoop should not be run in wasm_bindgen_futures::spawn_local
Running the event loop inside of an async function in spawn_local appears to block new JS promises from being run if they are created in an event callback:

https://github.com/rustwasm/wasm-bindgen/issues/2246

I am not sure exactly how to restructure the examples, but figured I would post this here in case anyone knows the right way to do this or comes across the same issue as I did.

dragly:
Running the event loop inside an async function blocks new futures from
being spawned in the examples using spawn_local. This change moves the
event loop out of the async function and runs it directly in main. To
still be able to initialize the wgpu-rs adapter and device asynchronously,
we therefore need to introduce a custom event that is sent to the event
loop once the wgpu-rs adapter and device have been created.

Fixes #454

### DIFF
--- a/examples/framework.rs
+++ b/examples/framework.rs
@@ -3,7 +3,7 @@ use futures::task::LocalSpawn;
 use std::time::{Duration, Instant};
 use winit::{
     event::{self, WindowEvent},
-    event_loop::{ControlFlow, EventLoop},
+    event_loop::{ControlFlow, EventLoop, EventLoopProxy},
 };
 
 #[cfg_attr(rustfmt, rustfmt_skip)]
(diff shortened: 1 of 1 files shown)


dragly:
I am unable to test this against the `wasm32-unknown-unknown` target because the current master branch fails with an `Uncaught (in promise) Error: undefined` when trying to build and run it with the instructions from https://gfx-rs.github.io/2020/04/21/wgpu-web.html

However, I have tested a similar approach against an earlier version in a different project and it seems to work just fine.

The way this is set up feels a bit complex and hacky. Please feel free to suggest other approaches :)

grovesNL (commented):
Thank you! This seems alright to me but I'd be interested if anyone else has other ideas on how to make this work. I mentioned it on Matrix in case somebody else could take a look too.

We'd also need to apply this change to any other examples that aren't using `framework`.

dragly:
That sounds like a good idea. I feel like the solution I propose here is more of a workaround since it adds some complexity and makes it harder to reason about the order of operations. If someone figures out exactly what is happening when the `spawn_local` fails to run, maybe a better solution can be found.

grovesNL:
Ok great :+1: in that case we can probably proceed by looking into winit/wasm-bindgen-futures interactions to see if we can fix this upstream. If there's no good place to resolve this upstream, then the workaround seems alright, or at least for now until there's a better solution.

On Matrix there was some discussion about the exception being used for control flow for winit and how it interacts with wasm-bindgen-futures, and whether that might be a problem, or if something else is going on here. Hopefully it could work without custom user events if we can track down where this is breaking.

kvark:
Project moved to `wgpu`

PR #509, opened 2020-08-17
Convert StagingBelt to take a &mut CommandEncoder

cwfitzgerald:
This allows `StagingBelt` to be `Send`, very important for async/highly multithreaded applications. Also added a bit more documentation.

Tested on the skybox example.

### DIFF
--- a/src/util/belt.rs
+++ b/src/util/belt.rs
@@ -1,9 +1,9 @@
 use crate::{
-    Buffer, BufferAddress, BufferDescriptor, BufferSize, BufferUsage, BufferViewMut, CommandBuffer,
-    CommandEncoder, CommandEncoderDescriptor, Device, MapMode,
+    Buffer, BufferAddress, BufferDescriptor, BufferSize, BufferUsage, BufferViewMut,
+    CommandEncoder, Device, MapMode,
 };
 use futures::{future::join_all, FutureExt};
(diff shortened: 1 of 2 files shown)


src/util/belt.rs
```diff
@@ -105,29 +110,23 @@ impl StagingBelt {
             .get_mapped_range_mut()
     }
 
-    /// Produce a command buffer with all the accumulated transfers.
+    /// Prepare for command buffers to be submitted.
     ///
     /// At this point, all the partially used staging buffers are closed until
     /// the GPU is done copying the data from them.
-    pub fn flush(&mut self, device: &Device) -> CommandBuffer {
+    pub fn flush(&mut self) {
```
kvark, line 117: I think with the new signature it should no longer be called `flush`. Maybe `finish()?`

src/util/belt.rs
```diff
@@ -105,29 +110,23 @@ impl StagingBelt {
             .get_mapped_range_mut()
     }
 
-    /// Produce a command buffer with all the accumulated transfers.
+    /// Prepare for command buffers to be submitted.
```
kvark, line 113: since there is nothing about "command buffers" in the signature, we shouldn't be documenting it like that.
Maybe "prepare mapped buffers to be used in submission" or something like that?

cwfitzgerald:
Addressed!
bors r=kvark

PR #511, opened 2020-08-17
Add an mdBook

cormac-obrien:
So far, the book contains a brief introduction and a fleshed-out tutorial to the "Hello, Triangle!" example. I'm open to any and all feedback!

This PR also doesn't have any of the build-and-deploy infrastructure set up, which I imagine we'll want before merging it.


book/src/01-05-hello-triangle.md
```diff
@@ -0,0 +1,256 @@
...
+    });
+
+    // let size = ...
+}
+```
+
+This code uses `Device::create_buffer_init`, a convenience method provided by `wgpu` to initialize a buffer with data at creation.
+Since this method is not part of the WebGPU standard, it requires the `DeviceExt` trait to be imported.
+The `contents` of the buffer are provided with `bytemuck::cast_slice`, which cleanly converts a `&[Vertex]` to a `&[u8]`.
+The buffer requires the `COPY_DST` usage so that the provided data can be copied into it, and of course the `VERTEX` usage because it's a vertex buffer.
```
JMS55, line 150: COPY_DST isn't needed here. It's only needed when copying new data to the buffer after creating it. See https://docs.rs/wgpu/0.6.0/wgpu/struct.BufferUsage.html#associatedconstant.COPY_DST.

book/src/01-02-instances-etc.md
```diff
@@ -0,0 +1,190 @@
...
+
+Now that you have an `Instance`, you can query for the available adapters on a system with `Instance::enumerate_adapters`.
+This returns an iterator over a series of `Adapter` objects.
+Try printing out the info for each adapter (`AdapterInfo`), like this:
+
+```rust,no_run,noplayground
+fn main() {
+    // let instance = ...
+
+    for adapter in instance.enumerate_adapters(wgpu::BackendBit::PRIMARY) {
```
kvark, line 34: I think we should focus on the API that is portable to the Web. `enumerate_adapters` can be polyfilled by returning a single adapter, but essentially `request_adapter` is most portable

book/src/01-04-clearing-the-screen.md
```diff
@@ -0,0 +1,153 @@
+# Clearing the Screen
+
+In this section, you'll learn how to clear the screen and handle window resizes.
+
+## The swap chain
+
+`wgpu` represents the framebuffers that will be presented to the screen with a data structure called a `SwapChain`.
+If you're coming from Vulkan or D3D12, you'll already be familiar with this term.
+Briefly, the `SwapChain` determines the format, size and presentation mode of the frames drawn by `wgpu`.
```
kvark, line 9: might be worth saying something in line with "SwapChain links a surface together with a logical device to draw on it"

book/src/01-04-clearing-the-screen.md
```diff
@@ -0,0 +1,153 @@
...
+    let mut swap_chain = device.create_swap_chain(&surface, &sc_desc);
+
+    // event_loop.run(...);
+}
+```
+
+A few notes on this code:
+- The `width` and `height` fields of `SwapChainDescriptor` take their values from the window size.
+  This is not maintained automatically, so the swap chain needs to be recreated when the window size changes.
+- The choice of `format` is platform-dependent. `Bgra8UnormSrgb` is ubiquitous on desktop Vulkan implementations ([Windows], [Linux]), [required on D3D12], and [the default for Metal] regardless of platform. When developing for mobile or the web, a different format may be required.
```
kvark, line 37: we actually have `swap_chain_get_preferred_format` in wgpu-core. We should expose this in `wgpu-rs`

book/src/01-05-hello-triangle.md
```diff
@@ -0,0 +1,256 @@
...
+When uploading data to the GPU, `wgpu` expects a `&[u8]`.
+You can obtain this from a `&[T]` a couple ways:
+- Call `as_ptr()` on the slice, cast to `*const T` and then `*const u8`, and then reconstruct a `&[u8]` from the pointer with `slice::from_raw_parts`. This can be tricky if you're not confident about the particulars of `unsafe` code.
+- Use a library for performing this conversion, such as `bytemuck`. This is the approach taken by these examples, as it minimizes the amount of `unsafe` code being written.
+
+Add `bytemuck` to your `Cargo.toml`:
+
+```toml
+[dependencies]
+bytemuck = "1.3"
```
kvark, line 110: does this have to be "1.3" versus just "1"?

kvark (approved):
Amazing work, thank you!
Note that there is no need to finish the book in the PR. We can land the first sections you got, and then follow-up.

kvark:
Project moved to `wgpu`
A book would still be great to have!

PR #515, opened 2020-08-18
Add logging information to README

cwfitzgerald:


README.md
```diff
@@ -76,6 +76,34 @@ In the future [WGSL](https://gpuweb.github.io/gpuweb/wgsl.html) will be the shad
...
+
+```rust
+tracing_log::LogTracer::init().unwrap()
+``` 
+
+#### `tracing` events -> `log` events
+
+The `tracing` crate has a `log` feature which will automatically use `log` if no subscriber is added.
+
+If you want events to be handled both by `tracing` and `log`, enable the `log-always` feature of `tracing`.
```
kvark, line 105: we should specify this as something to put into a cargo manifest, i.e.
```toml
tracing = { version = "0.1", features = ["log-always"] }
```

README.md
```diff
@@ -76,6 +76,34 @@ In the future [WGSL](https://gpuweb.github.io/gpuweb/wgsl.html) will be the shad
 
 For now, the shaders can be compiled to SPIR-V by running `make`, which requires you to have `glslang`s `glslangValidator` binary.
 
+## Logging
+
+`wgpu-core` uses `tracing` for logging and `wgpu-rs` uses `log` for logging.
```
kvark, line 81: maybe a comma is missing before "and"?

cwfitzgerald: I was always taught that `and` doesn't get a comma when used with only two elements, but english comma rules are nothing if not confusing

kvark: I'm not nearly a native speaker, so please free to ignore this suggestion :)

PR #516, opened 2020-08-18
Depend on less of futures

JMS55:


Cargo.toml
```diff
@@ -59,6 +59,7 @@ rand = { version = "0.7.2", features = ["wasm-bindgen"] }
 bytemuck = "1"
 noise = "0.6"
 ddsfile = "0.4"
+futures = "0.3"
```
cwfitzgerald, line 62: This should also only have the features the examples need.

cwfitzgerald:
Thank you!
bors r+

PR #517, opened 2020-08-19
Added some explanatory comments to hello-compute example

JonathanWoollett-Light:
Simply added some more comments explaing the basic compute example.

Also moved the lines:
```rust
let slice_size = numbers.len() * std::mem::size_of::<u32>();
let size = slice_size as wgpu::BufferAddress;
```
To be just before where they are used.

### DIFF
--- a/examples/hello-compute/main.rs
+++ b/examples/hello-compute/main.rs
@@ -20,10 +20,11 @@ async fn run() {
 }
 
 async fn execute_gpu(numbers: Vec<u32>) -> Vec<u32> {
-    let slice_size = numbers.len() * std::mem::size_of::<u32>();
-    let size = slice_size as wgpu::BufferAddress;
-
+    // Instantiates instance of WebGPU
+    // 'wgpu::BackendBit::PRIMARY' selects one of 'Vulkan + Metal + DX12 + Browser WebGPU' as the backend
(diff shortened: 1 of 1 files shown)


kvark:
This is very similar to #511 . Do you think it makes sense to duplicate this documentation across:
  - the compute example
  - actual docs
  - the git book

JonathanWoollett-Light:
@kvark  I think it is okay to have duplicated explanations, and examples should have explanations as readily availableas possible. It is far better to have 2 explanations than 1 you can't find.

https://github.com/gfx-rs/wgpu-rs/pull/511 seems a bit more complex. This just adding comments to an example, something sorta very simple and just helpful when using an example to learn. 

Generaly I think most examples should be pretty thoroughly commented, especially the introduction ones.

kvark:
My concern is then - that hello-compute is chosen arbitrarily here. It's not directly discoverable. Some may look at hello-triangle, or at the `cube` example, and will be confused about why the docs are not there as well.

JonathanWoollett-Light:
I can understand that, but effectively I've chosen it since I'm doing compute stuff and not graphical stuff. I think it is better 1 out of 3 examples are well commented than 0 out of 3, I believe it is atleast progress.

kvark:
Alright, there is no harm to keep it. Thank you for elaborating!
bors r+

PR #523, opened 2020-08-21
Change buffer binding to not take a slice

kvark:
This has been a constant source of confusion w.r.t dynamic offsets.
What it looks like it's doing: user provides a slice of buffer visible to the shader. Just like they provide slices to vertex and index buffer bindings.
What it's actually doing: it takes the offset and size separately, modifying the offset later on based on the dynamic offset.

We don't need this confusion. Semantics is different from `BufferSlice` and so we should have a different API here.


src/lib.rs
```diff
@@ -1646,6 +1654,15 @@ impl Drop for BufferViewMut<'_> {
 }
 
 impl Buffer {
+    /// Return the bindnig view of the whole buffer.
```
cwfitzgerald, line 1657: Typo

src/lib.rs
```diff
@@ -1646,6 +1654,15 @@ impl Drop for BufferViewMut<'_> {
...
+    pub fn as_binding(&self) -> BindingResource {
```
cwfitzgerald, line 1658: This function is great but the name doesn't quite match the usage to me. It needs a `whole` or `entire` or something to show that this generates a binding for the entire buffer.

PR #534, opened 2020-08-25
Removed erroneous zero size check related conditional in water example resize()

### LINKED ISSUE #519: Water example validation error when resize
Test on wgpu-rs 0.6, windows 10, GTX2070s
When resizing the water example, the Vulkan validation comes this log:

[1.956601 ERROR]()(no module):
VALIDATION [VUID-VkPresentInfoKHR-pImageIndices-01296 (-945112042)] : Validation Error: [ VUID-VkPresentInfoKHR-pImageIndices-01296 ] Object 0: handle = 0x23a9c726310, type = VK_OBJECT_TYPE_QUEUE; | MessageID = 0xc7aabc16 | Images passed to present must be in layout VK_IMAGE_LAYOUT_PRESENT_SRC_KHR or VK_IMAGE_LAYOUT_SHARED_PRESENT_KHR but is in VK_IMAGE_LAYOUT_UNDEFINED. The Vulkan spec states: Each element of pImageIndices must be the index of a presentable image acquired from the swapchain specified by the corresponding element of the pSwapchains array, and the presented image subresource must be in the VK_IMAGE_LAYOUT_PRESENT_SRC_KHR layout at the time the 
operation is executed on a VkDevice (https://github.com/KhronosGroup/Vulkan-Docs/search?q=)VUID-VkPresentInfoKHR-pImageIndices-01296)
object info: (type: QUEUE, hndl: 2450756100880)

### LINKED ISSUE #532: Water example viewport goes black momentarily on resize
Looking at the resize function for this example, I see:

```
        if sc_desc.width == 0 && sc_desc.height == 0 {
            // Stop rendering altogether.
            self.active = None;
            return;
        } else {
            // The next frame queued is the wrong size: (0, 0),
            // so we skip a frame to avoid crashes where our
            // textures are the correct (window) size, and the
            // frame is still (0, 0).
            self.active = Some(self.current_frame + 1);
        }
```

That doesn't seem to make sense, and changing to the following resolves the issue for me (viewport no longer goes black momentarily on resize):

```
       if sc_desc.width == 0 && sc_desc.height == 0 {
            // Stop rendering altogether.
            self.active = None;
            return;
        } 

        self.active = Some(self.current_frame);
```

crispweed:
The conditional at the start of resize() already tests for and exits on the zero width and height condition.

The conditional removed here was then getting run on every other call to resize(), which doesn't seem logical, doesn't correspond to the comment, and seemed to be causing at least one issue with rendered viewport going black during resize.

Fixes #532
(probably) Fixes #519

### DIFF
--- a/examples/framework.rs
+++ b/examples/framework.rs
@@ -250,8 +250,8 @@ fn start<E: Example>(
                 ..
             } => {
                 log::info!("Resizing to {:?}", size);
-                sc_desc.width = size.width;
-                sc_desc.height = size.height;
+                sc_desc.width = if size.width == 0 { 1 } else { size.width };
+                sc_desc.height = if size.height == 0 { 1 } else { size.height };
                 example.resize(&sc_desc, &device, &queue);
                 swap_chain = device.create_swap_chain(&surface, &sc_desc);
             }
--- a/examples/water/main.rs
+++ b/examples/water/main.rs
@@ -648,13 +648,8 @@ impl framework::Example for Example {
             // Stop rendering altogether.
             self.active = None;
             return;
-        } else {
-            // The next frame queued is the wrong size: (0, 0),
-            // so we skip a frame to avoid crashes where our
-            // textures are the correct (window) size, and the
-            // frame is still (0, 0).
-            self.active = Some(self.current_frame + 1);
         }
+        self.active = Some(self.current_frame);
 
         // Regenerate all of the buffers and textures.
 


kvark:
I'm thinking that a proper application would not even *try* to create a swapchain when it's minimized. It serves no use. It should not try to render anything.
But for a short-term fix, your PR is good. We've had a lot of reports for crashing on minimizing the examples...

PR #537, opened 2020-08-26
Update wgpu with DepthComparison component and better shader validation

kvark:
Depends on https://github.com/gfx-rs/wgpu/pull/898

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,14 +26,14 @@ vulkan-portability = ["wgc/gfx-backend-vulkan"]
 package = "wgpu-core"
 #version = "0.6"
 git = "https://github.com/gfx-rs/wgpu"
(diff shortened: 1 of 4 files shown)


cwfitzgerald:
What's the motivation for this?

kvark:
Forgot to mention that, sorry!
GLSL 450 generates `gl_CullDistance` semantic in SPIR-V that Naga doesn't handle because it's not portable, and it's not in WGSL. Simply switching to 440 resolves that with no obvious downsides.

kvark:
bors r=cwfitzgerald

kvark:
The latest Naga allows us to have `gl_CullDistance` and eat it, too. So shader changes are now reverted 🎉 
bors r=cwfitzgerald

PR #540, opened 2020-08-29
Improve and fix examples/texture-arrays

im-0:
Original examples/texture-arrays produces following output on my machine (Linux, Mesa/RADV 20.2-rc3, Radeon VII):

texture-arrays-non-uniform-bug

Quick investigation showed that only shaders with non-uniform indexing are problematic (first commit helped with this). Then, after searching on the web, I figured out that the problem is in missing `nonuniformEXT`. Second commit fixes this.

I am new to graphics programming, and `nonuniformEXT` may work just by accident. So please confirm my finding before merging.


im-0:
Also see https://github.com/gfx-rs/wgpu/pull/908

cwfitzgerald:
Interesting findings! It seems like only drivers of a certain red persuasion seem to actually care about this, hence why I never caught this on my green and blue machines.

An aside, we really should provide a way to switch between shaders at runtime in that example.

bors r+

PR #547, opened 2020-09-01
Implement on_uncaptured_error 

scoopr:
This is very crude at the moment, I'm mostly just checking if the directions is at all viable.

So I set out to implement `on_caught_error`, focusing solely on the direct backend. It just accepts a `Fn(GPUError)` (and I named it GPUError because I didn't yet want to rename the import of `std::error::Error`).

I store the handler in a `ErrorSinkRaw` struct, which is then wrapped in Arc+Mutex, with the idea that other resources can point to the same Arc, so when `Device::on_uncaught_error` is called mid application, all resources will point to the right place. Otherwise I think something like `Buffer::unmap` would have to find the `Device` it was created with, to find the uncaught handler.

Now I store the `ErrorSink` in the `Device` struct, because it is a convenient place to hold it. But I guess it is a bit nonsensical with the web backend, so either it would need to be `cfg(not(wasm32))`, or move the whole thing in the backend side, but in the direct backend, I don't see a convenient place to store the closure. But the unwrapping is done in direct backend, so it can't really be moved up in to wgpu-core either, unless the error handling overall is moved there.

Also related, I'm now passing the `error_sink` argument to the few methods where I implemented the error handling, which I think is a bit ugly, but it works.

### DIFF
--- a/src/backend/direct.rs
+++ b/src/backend/direct.rs
@@ -497,9 +497,14 @@ fn map_pass_channel<V: Copy + Default>(
     }
 }
 
+#[derive(Debug)]
+pub(crate) struct Device {
+    id: wgc::id::DeviceId,
+}
+
 impl crate::Context for Context {
     type AdapterId = wgc::id::AdapterId;
-    type DeviceId = wgc::id::DeviceId;
+    type DeviceId = Device;
     type QueueId = wgc::id::QueueId;
     type ShaderModuleId = wgc::id::ShaderModuleId;
     type BindGroupLayoutId = wgc::id::BindGroupLayoutId;
@@ -567,7 +572,8 @@ impl crate::Context for Context {
             *adapter => global.adapter_request_device(*adapter, desc, trace_dir, PhantomData)
         )
         .unwrap_pretty();
-        ready(Ok((device_id, device_id)))
+        let device = Device { id: device_id };
+        ready(Ok((device, device_id)))
     }
 
     fn adapter_features(&self, adapter: &Self::AdapterId) -> Features {
@@ -582,12 +588,12 @@ impl crate::Context for Context {
 
     fn device_features(&self, device: &Self::DeviceId) -> Features {
         let global = &self.0;
-        wgc::gfx_select!(*device => global.device_features(*device)).unwrap_pretty()
+        wgc::gfx_select!(device.id => global.device_features(device.id)).unwrap_pretty()
     }
 
     fn device_limits(&self, device: &Self::DeviceId) -> Limits {
         let global = &self.0;
-        wgc::gfx_select!(*device => global.device_limits(*device)).unwrap_pretty()
+        wgc::gfx_select!(device.id => global.device_limits(device.id)).unwrap_pretty()
     }
 
     fn device_create_swap_chain(
@@ -598,7 +604,7 @@ impl crate::Context for Context {
     ) -> Self::SwapChainId {
         let global = &self.0;
         wgc::gfx_select!(
-            *device => global.device_create_swap_chain(*device, *surface, desc)
+            device.id => global.device_create_swap_chain(device.id, *surface, desc)
         )
         .unwrap_pretty()
     }
@@ -614,7 +620,7 @@ impl crate::Context for Context {
         };
         let global = &self.0;
         wgc::gfx_select!(
-            *device => global.device_create_shader_module(*device, desc, PhantomData)
+            device.id => global.device_create_shader_module(device.id, desc, PhantomData)
         )
         .unwrap_pretty()
     }
@@ -626,7 +632,7 @@ impl crate::Context for Context {
     ) -> Self::BindGroupLayoutId {
         let global = &self.0;
         wgc::gfx_select!(
-            *device => global.device_create_bind_group_layout(*device, &wgc::binding_model::BindGroupLayoutDescriptor {
+            device.id => global.device_create_bind_group_layout(device.id, &wgc::binding_model::BindGroupLayoutDescriptor {
                 label: desc.label.map(Borrowed),
                 entries: Borrowed(desc.entries),
             }, PhantomData)
@@ -673,8 +679,8 @@ impl crate::Context for Context {
             .collect::<Vec<_>>();
 
         let global = &self.0;
-        wgc::gfx_select!(*device => global.device_create_bind_group(
-            *device,
+        wgc::gfx_select!(device.id => global.device_create_bind_group(
+            device.id,
             &bm::BindGroupDescriptor {
                 label: desc.label.as_ref().map(|label| Borrowed(&label[..])),
                 layout: desc.layout.id,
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -2610,3 +2621,62 @@ impl SwapChain {
...
+}
+
+impl Debug for ErrorSinkRaw {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f,"ErrorSink")
+    }
+}
+
+fn default_error_handler(err: GPUError) {
+    eprintln!("WGPU Error: {}",err);
```
kvark, line 2680: any reason this isn't a `panic` right away?

src/lib.rs
```diff
@@ -2610,3 +2621,62 @@ impl SwapChain {
...
     }
 }
+
+type ErrorSinkCallback = dyn Fn(GPUError) ;
+type ErrorSink = Arc<Mutex<ErrorSinkRaw>>;
+
+
+/// TODO:
+#[derive(Debug)]
+pub struct GPUError {
```
kvark, line 2631: we don't have any `GPU` prefixed types. Could this be `AnyError`?

src/lib.rs
```diff
@@ -436,6 +438,7 @@ pub struct Adapter {
 pub struct Device {
     context: Arc<C>,
     id: <C as Context>::DeviceId,
+    error_sink: ErrorSink
```
kvark, line 441: It seems to me that `ErrorSink` is only needed in the direct backend, right?
We could make it a part of the direct objects that have methods. I.e. in the backend, instead of this:
```rust
type BufferId = wgc::id::BufferId;
```
We could have:
```rust
struct Buffer {
  id: wgc::id::BufferId,
  error_sink: ErrorSink,
}
```
I do wonder how verbose it would be comparing to the current code here, but at least it's going to be properly concealed in the backend.

examples/hello-triangle/main.rs
```diff
@@ -30,8 +30,11 @@ async fn run(event_loop: EventLoop<()>, window: Window, swapchain_format: wgpu::
         .await
         .expect("Failed to create device");
 
+    device.on_uncaught_error(move |err| eprintln!("Gosh darn it! {}", err));
     // Load the shaders from disk
-    let vs_module = device.create_shader_module(wgpu::include_spirv!("shader.vert.spv"));
+    let vs_module = device.create_shader_module(wgpu::ShaderModuleSource::Wgsl(std::borrow::Cow::Borrowed("")));
```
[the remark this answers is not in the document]
scoopr: Yes naturally, I left it here for showing that it works :)

src/lib.rs
scoopr: For whatever reason, I couldn't get it to print anything but "Box<Any>" for the panic message. Though I kind of prefer it like this anyway, the actual error is somewhat easier to read, but I would be fine either way.

src/lib.rs
scoopr: Ah yes, as I mentioned in the description, I couldn't call it `Error` because the std `Error` is imported, and was gonna rename that import later. But I was gonna call it just `Error`, for being consistent with the `GPUFoo` -> `wgpu::Foo` convention. 

It does kind of look like it should impl std `Error` as well for being more rusty, though it doesn't matter much as it is not used as a `Result::Err` in the public interface.

src/lib.rs
scoopr: Ah, so direct `impl Context` could just have `type BufferId = Buffer` instead of `type BufferId = wgc::id::BufferId`? That could work, then it would be contained in the right place within the direct backend, that would make sense!

Yeah, the wrapping structs feel a bit annoyingly verbose, but if you are ok with it, I'll try with that

kvark: hopefully, it's not for too many of them, and not too annoying?

src/lib.rs
kvark: I don't mind either way

scoopr:
Alright, I feel pretty good about this one so far. It is not yet the full implementation, but the plumbing should be there.

Also, did the spec change from uncaught to uncaptured in the meantime? :D

scoopr:
If the approach looks good, then I could go ahead and plough through rest of the unwrap sites and types

src/backend/direct.rs
```diff
@@ -614,9 +628,12 @@ impl crate::Context for Context {
         };
         let global = &self.0;
         wgc::gfx_select!(
-            *device => global.device_create_shader_module(*device, desc, PhantomData)
+            device.id => global.device_create_shader_module(device.id, desc, PhantomData)
```
kvark, line 631: we could probably figure out a slightly shorter way of doing this, although I'm not sure if we should (given the complexity). I.e. this could be
```rust
device.id => device.create_wrap(
    &self.0,
    |g| g.device_create_shader_module(device.id, desc, PhantomData),
    |g| g.shader_module_error(PhantomData),
)
```
It slightly feels like fighting against the `gfx_select` macro limitations.
It's up to you to decide if the current code needs to stay, or something like this used instead.

src/backend/direct.rs
```diff
@@ -497,9 +501,15 @@ fn map_pass_channel<V: Copy + Default>(
     }
 }
 
+#[derive(Debug)]
+pub(crate) struct Device {
+    pub id: wgc::id::DeviceId,
```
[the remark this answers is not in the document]
scoopr: Oh yeah, must be some leftover

src/backend/direct.rs
scoopr: Hmm. It is certainly tempting to try to hide some of this. The closure in the unwrap is certainly a little awkward. But it this also has a little more implicit feel.

I started thinking, could I implement some `ErrorId` trait for the various id types, so the unwrap could just call `T::error_id()`, and not have the closure? But more work and also has a little implicit feel, especially as each trait impl probably has a single place of use.

I think I'm gonna leave it as is, unless something clearly better comes to mind

scoopr:
Addressed the issues.
I also added a new commit where I implement std Error for the Error. I made it rely more on the `source()`.
Default handler error looks something like this now:
```
wgpu error: Validation error

Caused by:
    error in stage VERTEX: unable to find an entry point matching the Vertex execution model

thread 'main' panicked at 'Handling wgpu errors as fatal by default', src/backend/direct.rs:1462:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I like it, but if you like terser errors, the default can be, as I can now have my own formatting as well :)

scoopr:
Tweaked the errors a little in wgpu-core as a test, and I got that to this. The messages would need to be cleaned up a little, but I find there is great potential here :)

```
wgpu error: Validation error

Caused by:
    Creating render pipeline (label: Some("hello-triangle pipeline"))
    error in stage VERTEX
    unable to find an entry point matching the Vertex execution model

thread 'main' panicked at 'Handling wgpu errors as fatal by default', src/backend/direct.rs:1457:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

kvark (approved):

cwfitzgerald (approved):
Awesome!

r=kvark,cwfitzgerald

cwfitzgerald:
bors r=kvark,cwfitzgerald

PR #549, opened 2020-09-03
Pass [u8] instead of [u32] for push constants

### LINKED ISSUE #544: Use &[u8] instead of &[u32] for push constants
So, we are talking about [`set_push_constants`](https://docs.rs/wgpu/0.6.0/wgpu/struct.RenderPass.html#method.set_push_constants). Although it's possible to cast the slices from and to each other, I would like to point out that with a slice of `u8` you can prepare the data with [`bytes`](https://docs.rs/bytes/0.5.6/bytes/index.html); and it is much cleaner if you want to dynamically assign values. Currently we have to cast the types for multiple times, and as I found there are already people got [lost in the wild](https://github.com/gfx-rs/wgpu-rs/issues/524). Since `&[u8]` is generally recognized as a type for buffer, I would suggest using `&[u8]` instead of `&[u32]` to pass the push constant buffer. If it's acceptable, I would be happily opening a PR for this. :)

PENGUINLIONG:
Closes #544.
I wonder if we need to `assert` and ensure the inputs are aligned to four bytes?

### DIFF
--- a/examples/texture-arrays/main.rs
+++ b/examples/texture-arrays/main.rs
@@ -329,9 +329,9 @@ impl framework::Example for Example {
         rpass.set_vertex_buffer(0, self.vertex_buffer.slice(..));
         rpass.set_index_buffer(self.index_buffer.slice(..));
         if self.uniform_workaround {
-            rpass.set_push_constants(wgpu::ShaderStage::FRAGMENT, 0, bytemuck::cast_slice(&[0]));
+            rpass.set_push_constants(wgpu::ShaderStage::FRAGMENT, 0, bytemuck::cast_slice(&[0u32]));
             rpass.draw_indexed(0..6, 0, 0..1);
-            rpass.set_push_constants(wgpu::ShaderStage::FRAGMENT, 0, bytemuck::cast_slice(&[1]));
+            rpass.set_push_constants(wgpu::ShaderStage::FRAGMENT, 0, bytemuck::cast_slice(&[1u32]));
             rpass.draw_indexed(6..12, 0, 0..1);
         } else {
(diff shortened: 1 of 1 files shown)


src/backend/direct.rs
```diff
@@ -84,15 +84,13 @@ mod pass_impl {
...
             unsafe {
                 wgpu_compute_pass_set_push_constant(
                     self,
                     offset,
-                    (data.len() * std::mem::size_of::<u32>())
-                        .try_into()
-                        .unwrap(),
-                    data.as_ptr(),
+                    data.len().try_into().unwrap(),
+                    data.as_ptr() as *const u32,
```
cwfitzgerald, line 93: This api should be changed in wgpu-core to take a pointer to u8. As of right now this is unsound as the data may not be aligned to 4 bytes, causing ub on dereference.

cwfitzgerald (changes requested):

kvark:
@PENGUINLIONG why closing?

PENGUINLIONG:
> @PENGUINLIONG why closing?

😥 wow i didn't actually. probably an accident

PENGUINLIONG:
I'm reopening this. It would fail the checks for now but it should work with the upgraded `wgpu`.

cwfitzgerald:
Could you rebase/update this?

PENGUINLIONG:
Should be good to go :)

cwfitzgerald (approved):
Could you squash this, then we're good to go.

PENGUINLIONG:
Oh, I just realized @kvark has already updated the push constant interfaces. Closing this.

Issue #190, opened 2020-03-05
Add user-friendly buffer type casting support?


lachlansneff:
Currently, the buffers returned by `Buffer::map_read` and `Buffer::map_write` return types which can be safely converted to `&[u8]`. If a buffer contains a more complex struct, you either have to cast unsafely or build it manually. 

If the `BufferReadMapping` and `BufferWriteMapping` structs had an type parameter (it could even be set to `u8` by default for backwards compatibility) that was bounded on `zerocopy::FromBytes` or `bytemuck::Pod`, it'd be more convenient to safely cast the slice to a slice of an arbitrary type.

For example, `BufferReadMapping` could be implemented as:

```rust
pub struct BufferReadMapping<T> {
    data: *const u8,
    size: usize,
    buffer_id: wgc::id::BufferId,
    _marker: PhantomData<T>,
}
//TODO: proper error type
pub type BufferMapReadResult<T> = Result<BufferReadMapping<T>, ()>;

impl<T: bytemuck::Pod> BufferReadMapping<T>
{
    pub fn as_slice(&self) -> &[T] {
        bytemuck::cast_slice(unsafe {
            slice::from_raw_parts(self.data, self.size)
        })
    }
}

impl<T> Drop for BufferReadMapping<T> {
    fn drop(&mut self) {
        wgn::wgpu_buffer_unmap(self.buffer_id);
    }
}

```

This would work if `BufferReadMapping` implemented `Deref` as well, which might be even more user-friendly.

grovesNL:
FWIW we did have a zeropcopy bound previously but the safety is still questionable because being POD isn't enough (some level still needs to be aware of `repr` and alignment of fields compared to their representation in shaders, etc.). There's some good discussion in https://github.com/gfx-rs/wgpu-rs/pull/119.

So for now we decided to expose `&[u8]` and let the caller decide whether the cast is actually safe

lachlansneff:
Ah, okay, that makes sense. That thread is pretty dense, but the outcome is clear haha.

I would like to see this general problem of matching up data-types between the cpu and gpu sides revisited at some point, but it looks like it's not too important right now.

PR #551, opened 2020-09-07
Replace unsafe impl Pod with safe derive

rukai:
bytemuck now includes derives that will implement the Pod trait, failing to compile if the struct cannot safely be a Pod.
Lets use it to remove most of the unsafe usage from the examples.
closes https://github.com/gfx-rs/wgpu-rs/issues/190

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ log = "0.4"
 png = "0.16"
 winit = { version = "0.22.1", features = ["web-sys"] }
 rand = { version = "0.7.2", features = ["wasm-bindgen"] }
-bytemuck = "1"
+bytemuck = { version = "1.4", features = ["derive"] }
 noise = "0.6"
 ddsfile = "0.4"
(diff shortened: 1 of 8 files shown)


cwfitzgerald (approved):
Nice!

Cargo.toml
```diff
@@ -56,7 +56,7 @@ log = "0.4"
 png = "0.16"
 winit = { version = "0.22.1", features = ["web-sys"] }
 rand = { version = "0.7.2", features = ["wasm-bindgen"] }
-bytemuck = "1"
+bytemuck = { version = "1", features = ["derive"] }
```
kvark, line 59: I assume this feature is only available since some `1.x` minor version?

kvark (changes requested):
One small change is needed, otherwise looks awesome!

Cargo.toml
rukai: Good point! fixed.

kvark (approved):
bors r+

PR #553, opened 2020-09-08
Add hello-windows example

qthree:
# `hello-windows`
## Description

This is demonstration of ability to create multiple windows with differently colored backgrounds.

## Screenshot
screenshot

# `shared-textures` (old version)
## Description

This is demonstration of ability to create multiple windows and share single texture across all of them.
Available in ad3030ae8f61e3ea3f7b913cd03c366119d058fb commit.

## Screenshot
screenshot


examples/shared-textures/main.rs
```diff
@@ -0,0 +1,339 @@
...
+        )
+        .await
+        .expect("Failed to create device");
+
+    let bind_group_layout = device.create_bind_group_layout(&wgpu::BindGroupLayoutDescriptor {
+        label: None,
+        entries: &[
+            wgpu::BindGroupLayoutEntry {
+                binding: 0,
+                visibility: wgpu::ShaderStage::FRAGMENT,
```
kvark, line 110: let's try to strip away all the parts of the example that are not related or critical to demonstrating the main aspect of it - multi-window. So there needs to be no shaders at all here: instead, you can just clear the color in each window to a different color (thus making a picture of windows?)

kvark (commented):
Nice, thank you for contributing this!
Just a few notes to address

examples/shared-textures/main.rs
```diff
@@ -0,0 +1,339 @@
...
+    }
+}
+
+async fn run(
+    event_loop: EventLoop<()>,
+    mut windows: Vec<Window>,
+    swapchain_format: wgpu::TextureFormat,
+) {
+    let instance = wgpu::Instance::new(wgpu::BackendBit::PRIMARY);
+    let first_window = windows.remove(0);
```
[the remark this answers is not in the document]
qthree: We need `first_surface` to create `adapter`, but we can't create `Viewport` without `device`. But once `first_surface` is created we probably shouldn't create another surface for the same window, so we remove it from vector and handle separately from another windows.

examples/shared-textures/README.md
```diff
@@ -0,0 +1,13 @@
+# shared-textures
```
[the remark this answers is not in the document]
qthree: Renamed to `hello-windows` as suggested.

examples/shared-textures/main.rs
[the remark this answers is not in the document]
qthree: Refactored to create all surfaces at once, before creation of adapter. But i done this with type-level state and encapsulated unsafe.

examples/shared-textures/main.rs
qthree: All irrelevant parts are removed: texture creation, bind group and even pipeline.

examples/hello-windows/main.rs
```diff
@@ -0,0 +1,220 @@
...
+        self.swap_chain
+            .get_current_frame()
+            .expect("Failed to acquire next swap chain texture")
+            .output
+    }
+}
+
+async fn run(
+    event_loop: EventLoop<()>,
+    viewports: Vec<ViewportDesc>,
```
kvark, line 79: I think it would be easier to not make `ViewportDesc` generic, and just have `Vec<(Window, wgpu::Color)>` here

kvark (approved):
It looks great! I have one review note, which is optional for you to address.
Let me know what you think, and we'll proceed

kvark (approved):

kvark:
Would you mind squashing and turning this from a draft into a real PR?

examples/hello-windows/main.rs
qthree: Ok, done. It's fine to simplify example, but in real-life situation there would be much more parameters other than just `background`.

kvark:
hmm, errors on the web backend are unrelated and weird, as if the web-sys was updated with a breaking change
@grovesNL do you know anything about this?

grovesNL:
@kvark a new version of web-sys was published (including the changes we are using on the gecko branch), so we can cherry-pick https://github.com/gfx-rs/wgpu-rs/pull/497 now

grovesNL:
@kvark I also forgot to mention that breaking changes in web-sys in patch releases are allowed for unstable APIs (including WebGPU). Unstable APIs are only available by passing the `RUSTFLAGS=--cfg=web_sys_unstable_apis` environment variable which is meant to acknowledge that there could be breaking changes when new versions are published.

web-sys is setup this way so web-sys doesn't have to release a new minor version whenever any web API changes a signature (this generally only happens for unstable APIs in active development, like WebGPU).

We could allow failures for wasm builds in wgpu-rs, or be sure to patch any wgpu-rs branches using it as soon as new versions of web-sys are available.

rukai:
Can't we pin to a specific version then "=0.3.45" ?

grovesNL:
@rukai yeah definitely, we could pin the version here. One downside of pinning the version is that users need to use a matching version of wasm-bindgen-cli for wgpu (instead of just using the latest version for everything), but maybe that's ok for now.

PR #555, opened 2020-09-10
Update to latest web-sys

grovesNL:
Cherry-pick #497 into `master` now that new versions of web-sys/wasm-bindgen/js-sys/wasm-bindgen-futures are published.

We could also do this for `v0.6` too if we want CI to succeed there, or alternatively consider disabling CI for wasm builds on `v0.6`.


kvark (approved):
bors r+

PR #557, opened 2020-09-10
Pin web-sys version

rukai:
As mentioned in https://github.com/gfx-rs/wgpu-rs/pull/553 this will prevent the wasm build from breaking.
@grovesNL Is pinning just web-sys and wasm-bindgen sufficient?


grovesNL (approved):
Yeah I think just pinning these two should be ok for now. Thank you!
bors r+

PR #559, opened 2020-09-11
Update wgpu with the polygon modes support, use in the cube example

kvark:
Uses https://github.com/gfx-rs/wgpu/pull/921, cc @manuel-woelker


cwfitzgerald (approved):
One minor concern but good to go afterwards.

examples/cube/main.rs
```diff
@@ -269,30 +294,56 @@ impl framework::Example for Example {
...
+                rasterization_state: Some(wgpu::RasterizationStateDescriptor {
+                    front_face: wgpu::FrontFace::Ccw,
+                    cull_mode: wgpu::CullMode::Back,
+                    polygon_mode: wgpu::PolygonMode::Line,
+                    ..Default::default()
+                }),
+                primitive_topology: wgpu::PrimitiveTopology::TriangleList,
+                color_states: &[wgpu::ColorStateDescriptor {
+                    format: sc_desc.format,
+                    color_blend: wgpu::BlendDescriptor {
```
cwfitzgerald, line 328: Do we actually need alpha blending here? I would imagine it would only overwrite the parts of the polygon that are actually visible.

kvark:
Not required, but still good to demonstrate in more than just the water example :)

Bors r=cwfitzgerald

PR #562, opened 2020-09-12
More errorsinkifying

scoopr:
Continuing on with errorsinkifying more places, these wrap Buffer, Texture and CommandEncoder with a struct that includes the `error_sink`, and uses it for the error handling.

I think most of the apis are errorsinkified with this. The remaining ones may need some more thinking. 

Queue would be wrapped mostly for the submit, the write_ stuff could in theory piggyback on the texture/buffer error-sink.
Others like swapchain, some mapping stuff, etc. needs some discussion or other implementation work.

### DIFF
--- a/src/backend/direct.rs
+++ b/src/backend/direct.rs
@@ -121,10 +121,10 @@ mod pass_impl {
         }
         fn dispatch_indirect(
             &mut self,
-            indirect_buffer: &wgc::id::BufferId,
+            indirect_buffer: &super::Buffer,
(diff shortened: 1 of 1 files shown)


scoopr:
Uh, of course `command_encoder` apis are missing also. I hope it is ok to wrap those ids with a struct containing error_sink too? even if those are create slightly more often than most?

kvark (approved):
Thank you!
the "xxx.id.id" parts look a bit unfortunate, but I don't have a better idea :)
bors r+

PR #565, opened 2020-09-14
Add context to errors

scoopr:
ContextError is just a simple wrapper that contains a string and the original error, and it is used to decorate the error to give it more context.


kvark (approved):
Nice, thank you!
bors r+

PR #566, opened 2020-09-14
Report out of memory errors

scoopr:


scoopr:
Finding for the right error kind is a bit.. brute force, I guess, but I think it might work fine if the `DeviceError` happens to be always used.

kvark (approved):
looks reasonable, thank you!
bors r+

PR #569, opened 2020-09-16
Reorder framework canvas setup for web

grovesNL:
winit only appends the canvas element once the window is built, and we need the canvas to exist before `create_surface` can use it. So we can just reorder the calls to build the window before using `create_surface` on the web backend in framework.

Also temporarily no-op debug markers for the web until we have them in gecko. Even though they're technically unimplemented, we can still proceed without them.

### DIFF
--- a/examples/framework.rs
+++ b/examples/framework.rs
@@ -80,9 +80,6 @@ async fn setup<E: Example>(title: &str) -> Setup {
         );
     };
 
-    #[cfg(target_arch = "wasm32")]
-    console_log::init().expect("could not initialize logger");
-
     let event_loop = EventLoop::new();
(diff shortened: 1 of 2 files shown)


examples/framework.rs
```diff
@@ -93,6 +90,22 @@ async fn setup<E: Example>(title: &str) -> Setup {
     }
     let window = builder.build(&event_loop).unwrap();
 
+    #[cfg(target_arch = "wasm32")]
+    {
+        use winit::platform::web::WindowExtWebSys;
+        console_log::init().expect("could not initialize logger");
```
kvark, line 96: we aren't going to get any console messages for window creation any more?

kvark (approved):
bors r+

PR #570, opened 2020-09-17
web: temporarily polyfill `create_buffer_init`

grovesNL:
Temporarily polyfill `create_buffer_init` on the web backend by using the old buffer mapping API. This is just a temporary fix until the new buffer mapping API is available in Nightly.

This polyfill allows us to run `cube` from `master` on Nightly (there's still lots missing from other examples). But at least some simple use cases (e.g. projects only using `create_buffer_init`/`write_buffer` and no other buffer mapping functions) can run with `master`.

### DIFF
--- a/src/backend/web.rs
+++ b/src/backend/web.rs
@@ -42,6 +42,30 @@ impl fmt::Debug for Context {
     }
 }
 
+impl Context {
+    pub fn create_buffer_init_polyfill(
+        device: &Sendable<web_sys::GpuDevice>,
+        desc: &BufferDescriptor<'_>,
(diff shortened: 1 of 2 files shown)


kvark (approved):
Quite a hack!
bors r+

grovesNL:
Missed a `cfg`
bors r=kvark

kvark:
It sounds like we need that in the gecko branch, too?

grovesNL:
The `gecko` branch still uses the old buffer mapping API (e.g. `create_buffer_mapped`) for now, so I don't think we need it there yet

PR #574, opened 2020-09-19
Implement AsRef for BufferView

de-vri-es:
This PR implements `AsRef` and `AsMut` in addition to `Deref` and `DerefMut` for `BufferView` and `BufferViewMut`.

This allows the buffer views to be used directly by generic code that wants an `AsRef<[u8]>`.

It's also subjectively a small win when you want to pass the views to non-generic code. I find `buffer.as_ref()` clearer than `&*buffer`. That also goes for `buffer.deref()`, but `Deref` is not in the prelude.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1658,6 +1658,18 @@ impl std::ops::DerefMut for BufferViewMut<'_> {
     }
 }
 
+impl AsRef<[u8]> for BufferView<'_> {
+    fn as_ref(&self) -> &[u8] {
+        self.data
+    }
+}
+
+impl AsMut<[u8]> for BufferViewMut<'_> {
+    fn as_mut(&mut self) -> &mut [u8] {
+        self.data
+    }
+}
+
 impl Drop for BufferView<'_> {
     fn drop(&mut self) {
         self.slice


kvark:
Thank you!
I wonder if we should remove the Deref implementations then. Thoughts?
@lachlansneff @cwfitzgerald @grovesNL

lachlansneff:
I agree that `&*buffer` isn't too clear, but you can actually just do `buffer[..]`, and `buffer.copy_from_slice`, or anything you just do with a slice. I suggest not removing the `Deref` impl, or if you do want to, add a function specifically for getting the slice, like `data()` and `data_mut()`.

de-vri-es:
I agree, the `Deref` impl is still useful if you just want to use the view as a slice directly.

I didn't think of it myself, but `&buffer[..]` is indeed already better than `&*buffer`. For generic code `AsRef` is still very useful though :)

src/lib.rs
```diff
@@ -1658,6 +1658,24 @@ impl std::ops::DerefMut for BufferViewMut<'_> {
     }
 }
 
+impl AsRef<[u8]> for BufferView<'_> {
+    fn as_ref(&self) -> &[u8] {
+        self.data
+    }
+}
+
+impl AsRef<[u8]> for BufferViewMut<'_> {
```
kvark, line 1667: can we remove that one?

de-vri-es: Oh, good point. I didn't notice that the deref impl for the write view panics. But that makes sense.

Will remove :)

cwfitzgerald:
I concur, keeping both Deref and AsRef makes sense to me.

src/lib.rs
de-vri-es: Hmm, but doesn't Rust still allow read access through a mut reference? That still leaves a soundness hole, but that's also true for `DerefMut` then.

lachlansneff: I don't think reading a write-only buffer is unsafe. It may just be very slow.

kvark: Rust allows that, but we don't want to pretend it's ok for this case. People can jump hoops to do it, still, but hopefully they will not :)

kvark:
Could you squash commits please?

de-vri-es:
> Could you squash commits please?

Sure :) Wasn't sure which you prefer.

Squashed.

src/lib.rs
de-vri-es: Right, as long as it's not unsafe. Otherwise it might be a good idea to make the functions that create buffer views unsafe.

But I don't know enough about GPU memory mapping to know what makes sense, so I'll happily defer to your collective wisdom :)

PR #576, opened 2020-09-20
Convert swapchain.get_current_frame to a immutable reference

cwfitzgerald:
Per our discussion on matrix, converts swapchain.get_current_frame to an immutable reference as it's internally synchronized.

Rationale: will allow me to take a non-pretty RW mutex as read instead of write in my application


kvark (approved):
So that would allow you to get multiple frames in flight, right? Is there anything else blocking this?
bors r+

cwfitzgerald:
No, this actually wasn't related to my attempt to get multiple frames in flight working. The big thing blocking that is calling `get_current_frame` on the swapchain before the last frame drops panics. I can file an issue about that.

PR #585, opened 2020-10-05
Demonstrate buffer write issue with DX11+IntelHD

qthree:
Demonstration of buffer write issue with DX11 + Intel HD. Not intended for merging.
Frame rendering starts before the end of uniform buffer update.
That results in a glitches.
If you see only one rectangle then you're OK.
RUIqXmDZge
`main.rs` has constants at the top for tweaking:
```
SUBMITS - how many times to redraw using different render pass, command buffer and separate submit
DRAWCALLS - how many times to call draw inside one render pass
AFTER_SUBMIT_DELAY_MS - sleeping between different submits.
```
I have this issue with:
```
SUBMITS = 10
DRAWCALLS = 1
AFTER_SUBMIT_DELAY_MS = 10
```
or
```
SUBMITS = 1
DRAWCALLS = 1000
AFTER_SUBMIT_DELAY_MS = 0
```
This demonstration is done using modified `hello-triangle` example.


kvark:
I'm able to reproduce. The Renderdoc command stream shows the following:
- buffer A is updated with `UpdateSubresource`
- buffer A gets copied into B with `CopySubresourceRegion`
- buffer B gets bound on each draw call with `VSSetConstantBuffers`

So it looks like one of the copies doesn't get synchronized to the draw call properly.

cwfitzgerald:
Can repro on latest dx11, investigating.

cwfitzgerald:
So after some discussion on the DX discord and on the matrix, we have decided that this is a intel driver bug. I'll report this upstream. I have some ideas on how to work around this, but it is going to be an extensive amount of work. 

Would it be possible to refile this issue on https://github.com/gfx-rs/wgpu as a normal issue with a link to the branch with the repro? It will help us better keep track of these upstream issues.

cwfitzgerald:
Closing in favor of https://github.com/gfx-rs/wgpu/issues/1060

PR #594, opened 2020-10-12
Fix mipmap example which was y-flipped

Uriopass:
Mipmap rendering was y-flipped, hence making one out of every two mipmaps wrong. See:

image

(look at the black dots)


kvark (approved):
Sweet!
bors r+

PR #604, opened 2020-11-04
[Blocked] Relax render/compute pass lifetimes

### LINKED ISSUE #188: The recent lifetime changes to `RenderPass` make it difficult to use.

### LINKED ISSUE #603: Demonstrate TypedArena use in an example
There are cases where resources are:
  - temporarily created
  - borrowed from a temporary `Arc<T>`

In these cases, a `TypedArena` can help extending the lifetimes of these structures. We should show it in one of the examples.

### LINKED ISSUE #1453: Lifetimes on `RenderPass` make it difficult to use.
Related PRs:

https://github.com/gfx-rs/wgpu-rs/pull/155
https://github.com/gfx-rs/wgpu-rs/pull/168

For context, I was looking at updating [imgui-wgpu](https://github.com/Yatekii/imgui-wgpu-rs/blob/master/src/lib.rs#L369) to work with the current master.

Now that `set_index_buffer` (and similar methods) take the a `&'a Buffer` instead of `&Buffer`, it has a ripple effect of "infecting" the surrounding scope with this lifetime.

While attempting to iterate and add draw calls, you end up with either "multiple mutable borrow" errors on wherever you're storing the `Buffer` or similar lifetime errors like "_data from `self` flows into `rpass` here_"

I would think that by calling `set_index_buffer` (and similar methods) that it would increment the ref-count on the buffer so that this lifetime restriction isn't needed.

### LINKED ISSUE #1462: Demonstrate TypedArena use in an example
There are cases where resources are:
  - temporarily created
  - borrowed from a temporary `Arc<T>`

In these cases, a `TypedArena` can help extending the lifetimes of these structures. We should show it in one of the examples.

kvark:
Mostly reverts #168 given that `wgpu-core` is now properly handling error IDs.
Closes #603
Closes #188

Has a major positive usability impact 🚆  It is no longer required by the type system to keep resources alive for the duration of a render/compute pass encoding. It's still required by the runtime: dropping a pass will produce an error if some of the dependencies are dropped before it.

Looking at the error messages may be difficult until an approach similar to https://github.com/gfx-rs/wgpu/pull/931#issuecomment-721386402 is rolled out in wgpu-rs (cc @scoopr). We'll make it a priority before releasing wgpu-0.7.

Edit: actually, the IDs here are not "error IDs" that `wgpu-core` handles. It will panic straight upon seeing such a stale ID...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,14 +26,14 @@ vulkan-portability = ["wgc/gfx-backend-vulkan", "gfx-backend-vulkan"]
 package = "wgpu-core"
 #version = "0.6"
 git = "https://github.com/gfx-rs/wgpu"
-rev = "7ac706f0a9d90ebdcffeae48e3c530b9921bff2c"
+rev = "7b9abf0fef76eea363869bc940fda1783dc12bab"
 features = ["raw-window-handle"]
 
 [dependencies.wgt]
 package = "wgpu-types"
 #version = "0.6"
 git = "https://github.com/gfx-rs/wgpu"
-rev = "7ac706f0a9d90ebdcffeae48e3c530b9921bff2c"
+rev = "7b9abf0fef76eea363869bc940fda1783dc12bab"
 
 [dependencies]
 arrayvec = "0.5"
--- a/src/backend/direct.rs
+++ b/src/backend/direct.rs
@@ -1375,10 +1375,10 @@ impl crate::Context for Context {
         .unwrap_error_sink(&encoder.error_sink, || ())
     }
 
-    fn command_encoder_begin_render_pass<'a>(
+    fn command_encoder_begin_render_pass(
         &self,
         encoder: &Self::CommandEncoderId,
-        desc: &crate::RenderPassDescriptor<'a, '_>,
+        desc: &crate::RenderPassDescriptor,
     ) -> Self::RenderPassId {
         wgc::span!(_guard, TRACE, "CommandEncoder::begin_render_pass wrapper");
         let colors = desc
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -368,10 +368,10 @@ trait Context: Debug + Send + Sized + Sync {
         encoder: &Self::CommandEncoderId,
         pass: &mut Self::ComputePassId,
     );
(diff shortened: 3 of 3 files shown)


scoopr:
I think the "stale IDs" will need to be handled more gracefully anyway. Consider, `pushErrorScope, loadScene, popErrorScope`, I would imagine should work without paniccing, no matter if some of the IDs are invalid on creation or dropped at the wrong time.

I'm so conflicted about this. I do love when the compiler tells me something is wrong, but in this particular case, the papercut is so such a big wound that I do hope good documentation and error messages will be enough.

kvark:
It's blocked because of this:
>Edit: actually, the IDs here are not "error IDs" that wgpu-core handles. It will panic straight upon seeing such a stale ID...

We don't want the panics for this, since WebGPU API doesn't allow panics there. At the same time, we don't want to add special handling of this in `wgpu-core` because it would have no way to tell if the ID was wrong to begin with (in which case we should panic, since this can't happen in WebGPU proper), or just got removed after it was used...

So @DefinitelyNotRobot should probably not wait for this. If you are constrained by the lifetimes here, consider `typed_arena` to keep stuff that needs extended lifetimes.

pythonesque:
I actually really like the lifetime discipline enforced here, I'd much rather get an error at compile time than runtime for stuff like this.

kvark:
Project moved to `wgpu`

PR #182, opened 2020-02-25
Wgpu gl support

zicklag:
Discussion: https://github.com/gfx-rs/wgpu/issues/450.

PR #606, opened 2020-11-05
Enable GL backend

kvark:
Replaces #182
Depends on https://github.com/gfx-rs/wgpu/pull/907


VincentFTS:
Thanks @kvark for this !
Do you plan to add this on the v0.6 branch ?

kvark:
@VincentFTS not until most of the examples can run at least on Linux with this backend. We aren't there yet, it's too rough atm to expose to users of the released crate.

PR #613, opened 2020-11-09
Add more context to errors originating in Passes and Encoders

Kimundi:
**Description**
This adds some top-level error context information to indicate wether an error originated in a `{Render,Compute}Pass` or `{Command,RenderBundle}Encoder` to make error messages more clear.

Combined with https://github.com/gfx-rs/wgpu/pull/1018, errors in passes will look like this:
```
wgpu error: Validation error

Caused by:
    In a RenderPass
    In a draw command
    index 3600 extends beyond limit 38
```


kvark (approved):
Thank you!

PR #615, opened 2020-11-11
Update web bindings

grovesNL:
- Use newer buffer mapping API everywhere and remove polyfill.
  - There's a weird thing here where we hand out `&[u8]`/`&mut [u8]`s for users to write into, but later have to copy back into the browser-provided `ArrayBuffer`. It's done during `Drop` here but open to some other ideas.
- Add in `web_sys::GpuTextureComponentType::DepthComparison` though it looks like we don't have it in Gecko yet:
  > GPUDevice.createBindGroupLayout: 'depth-comparison' (value of 'textureComponentType' member of GPUBindGroupLayoutEntry) is not a valid value for enumeration GPUTextureComponentType.
- `mipmap` fails because `layout` is still required in Gecko:
  > GPUDevice.createRenderPipeline: Missing required 'layout' member of GPUPipelineDescriptorBase.
- Add in some parts that we can implement now, including labels, debug markers, render bundles, and `get_bind_group_layout`


grovesNL:
This uses my local branch temporarily – I'll make a PR to wasm-bindgen soon so we can pin to that commit until it's published

kvark:
> GPUDevice.createRenderPipeline: Missing required 'layout' member of GPUPipelineDescriptorBase.

Implicit pipeline layouts are not trivial to implement in a multi-process scheme of things. I'm working on it now, mostly blocked by https://github.com/gfx-rs/naga/pull/267

kvark (approved):
Nothing rings a bell for me in these changes. Thank you for amazing work updating the Web backend! Please proceed when ready.

grovesNL:
@kvark if we want to proceed with this now we'll probably have to switch to a git dependency for wasm-bindgen and related crates. Is that alright or should we wait for the next wasm-bindgen release?

kvark:
Totally ok

grovesNL:
FYI I had to leave it in `patch` as a git dependency instead of a regular git dependency. For some reason cargo tries to match dependency versions between the direct web-sys dependency and the transitive older version of web-sys used in wgpu-core/gfx-backend-gl/glow, even though these dependencies aren't used in the wasm build.

I don't think this `patch` section should cause problems for anyone, and it's still only meant to be a temporary workaround until a new version of wasm-bindgen is available. If it causes issues, we can remove this and find some other way to make it work for now.

bors r=kvark

scoopr:
I think the `patch` section always need to be in the root Cargo.toml, which makes using the current `master` in a project a little bit of a hassle. Just copy-pasting the patch section worked for me though.

kvark:
The cause of the problem is web-sys breaking SemVer. It shouldn't update patch versions with breaking changes

grovesNL:
I don't think breaking changes were causing issues here. It just looks like we can't have both a git dependency on web-sys from wgpu-rs and a transitive dependency on a published version of web-sys (through glow) because of how package resolution works in cargo. The bigger issue is that the transitive dependency shouldn't matter, because we don't use gfx-backend-gl on the wasm target.

kvark:
My understanding is that you can't have those 2 dependencies because they are supposed to be SemVer compatible. If they were actual breaking versions of web-sys, cargo would have no issue here.

grovesNL:
Oh ok, in that case maybe it's because the [`~` in glow's dependencies](https://github.com/grovesNL/glow/blob/main/Cargo.toml#L44). We didn't always have that and I wasn't sure about adding it there, but it seemed to be necessary based on https://github.com/grovesNL/glow/pull/47 I wonder if we can remove it?

kvark:
I thought we are required to use "=version" for web-sys, specifically because it's not friends with semver...

grovesNL:
Yep that's the case for unstable web APIs (WebGPU in this case), but `=version` isn't needed for stable web APIs (e.g. WebGL for glow). I was trying to use it as a git dependency without the `=version` here though

PR #617, opened 2020-11-11
Add labels for some error messages

scoopr:
This places some pieces on how to transform the ids in the errors to a label, and adds them to the validation message.

This changes the formatting of the validation error to happen when constructing the error, rather than in the unhandled error handler.
It also requires some code for all the error variants for extracting the ids.

### DIFF
--- a/src/backend/direct.rs
+++ b/src/backend/direct.rs
@@ -1,7 +1,8 @@
 use crate::{
-    backend::native_gpu_future, BindGroupDescriptor, BindGroupLayoutDescriptor, BindingResource,
-    CommandEncoderDescriptor, ComputePipelineDescriptor, Features, Limits, LoadOp, MapMode,
-    Operations, PipelineLayoutDescriptor, RenderBundleEncoderDescriptor, RenderPipelineDescriptor,
+    backend::error::format_error, backend::native_gpu_future, BindGroupDescriptor,
+    BindGroupLayoutDescriptor, BindingResource, CommandEncoderDescriptor,
+    ComputePipelineDescriptor, Features, Limits, LoadOp, MapMode, Operations,
+    PipelineLayoutDescriptor, RenderBundleEncoderDescriptor, RenderPipelineDescriptor,
     SamplerDescriptor, ShaderModuleSource, SwapChainStatus, TextureDescriptor,
     TextureViewDescriptor,
 };
@@ -19,6 +20,10 @@ use typed_arena::Arena;
 pub struct Context(wgc::hub::Global<wgc::hub::IdentityManagerFactory>);
 
 impl Context {
+    pub(crate) fn global(&self) -> &wgc::hub::Global<wgc::hub::IdentityManagerFactory> {
+        &self.0
+    }
+
     pub fn adapter_get_info(&self, id: wgc::id::AdapterId) -> wgc::instance::AdapterInfo {
         let global = &self.0;
         wgc::gfx_select!(id => global.adapter_get_info(id)).unwrap_pretty()
@@ -651,9 +656,10 @@ impl crate::Context for Context {
         wgc::gfx_select!(
             device.id => global.device_create_shader_module(device.id, desc, PhantomData)
         )
-        .map_err(|err| err.with_context("In Device::create_shader_module".to_string()))
+        .map_err(|err| err.with_context("In Device::create_shader_module", None))
         .unwrap_error_sink(
             &device.error_sink,
+            &self,
             || wgc::gfx_select!( device.id => global.shader_module_error(PhantomData)),
         )
     }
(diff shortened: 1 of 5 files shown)


scoopr:
As discussed in gfx-rs/wgpu#931 and on matrix, this tries to do the label formatting on the wgpu-rs side. Depends on the changes on #931.

I'm not entirely happy with it, but it certainly is workable.

src/backend/error.rs
```diff
@@ -0,0 +1,83 @@
...
+
+fn fmt_pretty_any(error: &(dyn Error + 'static), context: &Context) -> String {
+    if let Some(pretty_err) = error.downcast_ref::<wgc::command::TransferError>() {
+        pretty_err.fmt_pretty(context)
+    } else {
+        format_error_line(error.as_display())
+    }
+}
+
+fn format_error_line(err: &dyn Display) -> String {
```
kvark, line 30: these should probably work with `fmt::Arguments` parameter instead of `&dyn Display`, see https://doc.rust-lang.org/std/fmt/struct.Arguments.html

src/backend/direct.rs
```diff
@@ -16,9 +16,15 @@ use std::{
 };
 use typed_arena::Arena;
 
+use crate::backend::error::format_error;
```
kvark, line 19: nit: can we merge this include with the other `use crate::`?

kvark (commented):
Thank you!
It looks to be a fine way to proceed, just a few notes to address :+1:

scoopr:
The latest commit now implements the labels for the errors I found that had Ids to resolve.
Also formats the labels in the same format with the `ContextError`. Also had to make ContextError box the source error, as I needed a concrete type to downcast to.
This requires gfx-rs/wgpu#1033 to land first

With this, a error could look like this
```
wgpu error: Validation Error

Caused by:
    In CommandEncoder::copy_buffer_to_buffer
    Copy error
    destination buffer/texture is missing the `COPY_DST` usage flag
      note: destination = `staging_buffer`
```

The same in release mode looks like
```
wgpu error: Validation Error

Caused by:
    In CommandEncoder::copy_buffer_to_buffer
    Copy error
    destination buffer/texture is missing the `COPY_DST` usage flag
      note: destination = `<Buffer-(0, 1, Metal)>`
```

The creation/ContextError looks like
```
wgpu error: Validation Error

Caused by:
    In Device::create_render_pipeline
      note: label = `hello-triangle pipeline`
    error in stage VERTEX
    unable to find an entry point at VERTEX stage
```

scoopr:
Oh it seems I neglected few points above that might still be relevant

scoopr:
Addressed most of the comments above, except for the `fmt::Arguments` thing. I see the value in it, but I don't have the stamina to work out the details of it right now, it seems to need some larger restructuring to be useful here.

src/backend/error.rs
```diff
@@ -0,0 +1,83 @@
+use std::{error::Error, fmt::Display};
+
+use super::Context;
+
+pub(crate) fn format_error(err: &(dyn Error + 'static), context: &Context) -> String {
```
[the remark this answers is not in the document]
scoopr: I'm fine either way, so I changed it to `impl`. I guess I originally thought that this code doesn't need to be monomorphised, and I figured it should emit only single body with the `dyn`. Though I guess with more vtable handling.

src/backend/error.rs
```diff
@@ -0,0 +1,242 @@
...
+        err_descs.push(fmt_pretty_any(source, context));
+        source_opt = source.source();
+    }
+
+    let desc = format!("Validation Error\n\nCaused by:\n{}", err_descs.join(""));
+    desc
+}
+
+fn fmt_pretty_any(error: &(dyn Error + 'static), context: &Context) -> String {
+    if let Some(pretty_err) = error.downcast_ref::<super::direct::ContextError>() {
```
kvark, line 21: can this be casted to a `PrettyError` trait object?

src/backend/error.rs
```diff
@@ -0,0 +1,242 @@
...
+            }
+            Self::InvalidPipeline(id) => {
+                let name = wgc::gfx_select!(id => global.render_pipeline_label(id));
+                ret.push_str(&format_label_line("render pipeline", &name));
+            }
+            Self::Buffer(id, ..) => {
+                let name = wgc::gfx_select!(id => global.buffer_label(id));
+                ret.push_str(&format_label_line("buffer", &name));
+            }
+            Self::DestroyedBuffer(id) => {
```
kvark, line 106: should probably merge this with `Self::Buffer` path?

kvark (commented):
Thank you! We are almost there!
I'll be fine following up with the FormatArgs. In fact, this code may need to be redone after https://github.com/gfx-rs/wgpu/issues/1032 is done, anyway.

kvark:
> This requires gfx-rs/wgpu#1033 to land first

we can wait for it to land in a few minutes :)

src/backend/error.rs
scoopr: Needs to be a concrete type
> the `downcast_ref` method cannot be invoked on a trait object

scoopr:
update `wgpu-core` rev, merged the match arm.

kvark:
Great, thank you!
bors r+

PR #618, opened 2020-11-12
Removed unusued Uniform struct in texture-arrays example.

Stygeon:
The Uniform struct in the texture-arrays example is unused.


cwfitzgerald:
Thanks!

bors r+

PR #624, opened 2020-11-15
Remove implicit conversion hello-triangle shader.vert

Napokue:
Remove implicit conversion to float, and change it to explicit conversion. Reason for this change is to prepare the hello-triangle example for Naga glsl-spirv path.


kvark (approved):
bors r+

PR #627, opened 2020-11-17
Fixed size in "boids" example

infinitesnow:
Buffer size is specified in number of items instead of bytes and fails validation against compiled SPIR-V


kvark (approved):
Oh, great, thank you!
bors r+

PR #631, opened 2020-11-21
Temporarily use `BackendBit::PRIMARY` in examples

grovesNL:
`BackendBit::all()` allows GL/surfman to be used on macOS which appears to prevent rendering with Metal (e.g. hello-triangle displays an empty window).

surfman will be removed in later versions of gfx, which will allow us to switch back to `BackendBit::all()` in the examples.


grovesNL:
cc @scoopr

kvark (approved):
bors r+

PR #632, opened 2020-11-21
Improve error messages

scoopr:
Improve error messages
depends on gfx-rs/wgpu#1037

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,14 +26,14 @@ vulkan-portability = ["wgc/gfx-backend-vulkan", "gfx-backend-vulkan"]
 package = "wgpu-core"
 #version = "0.6"
 git = "https://github.com/gfx-rs/wgpu"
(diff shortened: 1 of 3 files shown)


scoopr:
An example error message,
before:
```
wgpu error: Validation Error

Caused by:
    In a RenderPass
    In a set_pipeline command
    render pipeline is incompatible, incompatible sample count: 1 != 4
    incompatible sample count: 1 != 4
```
after:
```
Caused by:
    In a render pass
      note: encoder = `frame encoder`
    In a set_pipeline command
      note: render pipeline = `render pipe`
    Render pipeline is incompatible with render pass
    Incompatible sample count: 1 != 4
```

kvark (approved):
Needs to pick up the latest wgpu

kvark (approved):
Thank you!
Rebasing on top of this is going to be hard :)
bors r+

PR #635, opened 2020-11-24
Staging belt: do not panic by unwrapping send result in recall (fixes #634)

### LINKED ISSUE #634: StagingBelt SendError (recall future finishes after the main gpu user is dropped)
I'm using a slightly customized [iced](https://github.com/hecrj/iced) with a custom windowing integration in a GLib-async-based app.

In upstream iced [the `Compositor` owns a `LocalPool`](https://github.com/hecrj/iced/blob/8f081bad7756e16a4eb0ca4dcb1144fc47de5e9b/wgpu/src/window/compositor.rs#L15) where the `self.staging_belt.recall()` future is spawned, and then `run_until_stalled` is used. I can't use `LocalPool` because the GLib executor is already running (unlike tokio it doesn't have the special magic integration that allows a `LocalPool` to still work).

I modified the code to [spawn `recall` on the main GLib executor, adding `Abortable` to avoid the `SendError` situation in `recall`](https://github.com/myfreeweb/iced/commit/a9fb968cc3d2eb064ae36d021e894f300e6dbad1), which seemed to work, but now I have another situation where closing a window panics on the `SendError`.

So I'm wondering.. is it even correct that `recall` unwraps the send result? Should it ever panic on a `SendError` if it's an asynchronous process that might be finished right as the main graphics context is being dropped?

myfreeweb:


kvark (approved):
Thank you!
would you mind logging this at least? Something like `log::error("Unable to send the staging belt chunk: {:?}", error)`

kvark (approved):
thank you!
bors r+

myfreeweb:
Oh wait, `log` is a dev dependency?

kvark:
Oh snap, indeed we don't have `log`, so we shouldn't try to log the error :/ Sorry for misleading you into this!

kvark (approved):
bors r+

PR #639, opened 2020-11-25
Make BindingResource::TextureViewArray more flexible (addresses #588)

### LINKED ISSUE #588: Use &[&TextureView] instead of &[TextureView] in the TextureArray bind group entry
Not every user store TextureView's in Vec<TextureView>, someone can store it in the structs, which makes it basically "unmovable" into a slice.
But if we replace it with &[&TextureView], then we can collect a slice of references to pass it, which easier to handle.
Right now a workaround is to create an intermediate view and push it into an intermediate array, which is not great.

98devin:
Fixes #588 

Rather than having a field of `&[TextureView]`, which requires either keeping all TextureViews in a vector in the first place or moving them into a temporary vector or array and back, use `&[&TextureView]` instead. Obviates the need for hacky workarounds if you keep your TextureViews in non-contiguous structs, maps, or can't move them, etc.

For those using a vector previously, the required transformation is merely `.iter().collect()`.

### DIFF
--- a/examples/texture-arrays/main.rs
+++ b/examples/texture-arrays/main.rs
@@ -218,8 +218,8 @@ impl framework::Example for Example {
                 wgpu::BindGroupEntry {
                     binding: 0,
                     resource: wgpu::BindingResource::TextureViewArray(&[
-                        red_texture_view,
-                        green_texture_view,
+                        &red_texture_view,
+                        &green_texture_view,
                     ]),
                 },
                 wgpu::BindGroupEntry {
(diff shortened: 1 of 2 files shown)


grovesNL (approved):
Thank you! This looks good to me but I'll wait to see if @kvark or @cwfitzgerald have any concerns about the API change before merging

kvark:
The agreement has been reached on #588. I edited the description to link to it.
Thank you for the PR!
bors r=grovesNL

PR #649, opened 2020-11-28
Adopt the updated bind group entry API

kvark:
Depends on https://github.com/gfx-rs/wgpu/pull/1047


kvark:
Sorry about all the disruption in the API, folks!
bors r+

PR #656, opened 2020-12-02
Publish examples to subdirectory

grovesNL:
Avoid removing existing directory contents by publishing wasm examples to a subdirectory instead.


grovesNL:
Working on this currently so merging now – I'll followup with another PR if anything is wrong with this setup

PR #659, opened 2020-12-02
Temporarily skip debug labels for web backend

grovesNL:
Debug labels aren't currently supported by gecko. We can re-enable these later.


kvark (approved):
bors r+

PR #662, opened 2020-12-04
API and examples update for index_format

kejor:
__Connections__
#608 https://github.com/gfx-rs/wgpu/issues/978 https://github.com/gfx-rs/wgpu/pull/1019

__Description__
Use new API from https://github.com/gfx-rs/wgpu/commit/4513fb2b83fe2cd19682ccca9c55a6995f061699 where `wgpu_render_pass_set_index_buffer` and `wgpu_render_bundle_set_index_buffer` receive an `index_format` and `VertexStateDescriptor`'s `index_format` is an `Option`. The examples were updated accordingly.

PR #663, opened 2020-12-04
Shader module descriptor

### LINKED ISSUE #665: viewportCount set to 0 instead of 1, failing validation on Vulkan backend
I am not using multiview features, but it seems that the viewport count is incorrectly set at pipeline creation:

```
[0.110121 ERROR]()(no module): 
VALIDATION [VUID-VkPipelineViewportStateCreateInfo-viewportCount-01216 (-376016417)] : Validation Error: [ VUID-VkPipelineViewportStateCreateInfo-viewportCount-01216 ] Object 0: handle = 0x5628bad9a280, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0xe99671df | vkCreateGraphicsPipelines: The VkPhysicalDeviceFeatures::multiViewport feature is disabled, but pCreateInfos[0].pViewportState->viewportCount (=0) is not 1. The Vulkan spec states: If the multiple viewports feature is not enabled, viewportCount must not be greater than 1 (https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VUID-VkPipelineViewportStateCreateInfo-viewportCount-01216)
object info: (type: DEVICE, hndl: 94732933505664)
```

Attaching `vulkaninfo`, `vulkan trace` and `wgpu trace`
[vkinfo.txt](https://github.com/gfx-rs/wgpu-rs/files/5646967/vkinfo.txt)
[trace.tar.gz](https://github.com/gfx-rs/wgpu-rs/files/5646968/trace.tar.gz)
[vk_apidump.log](https://github.com/gfx-rs/wgpu-rs/files/5646969/vk_apidump.log)

kvark:
Depends on https://github.com/gfx-rs/wgpu/pull/1071
Fixes #665
Based on #662

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -210,7 +210,7 @@ trait Context: Debug + Send + Sized + Sync {
     fn device_create_shader_module(
         &self,
         device: &Self::DeviceId,
-        source: ShaderModuleSource,
+        desc: &ShaderModuleDescriptor,
     ) -> Self::ShaderModuleId;
     fn device_create_bind_group_layout(
         &self,
@@ -672,7 +672,7 @@ impl Drop for ShaderModule {
 }
 
 /// Source of a shader module.
-pub enum ShaderModuleSource<'a> {
+pub enum ShaderSource<'a> {
     /// SPIR-V module represented as a slice of words.
     ///
     /// wgpu will attempt to parse and validate it, but the original binary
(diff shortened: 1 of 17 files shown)


kvark:
I imagine this may become a bit of a problem when rust-gpu decides that it can provide a Naga Module directly to wgpu-rs.
Not sure what to do with it, we need a separate discussion.

src/lib.rs
```diff
@@ -687,6 +687,16 @@ pub enum ShaderModuleSource<'a> {
...
 }
 
+/// Descriptor for a shader module.
+pub struct ShaderModuleDescriptor<'a> {
+    /// Debug label of the shader module. This will show up in graphics debuggers for easy identification.
+    pub label: Label<'a>,
+    /// Source code for the shader.
+    pub source: ShaderSource<'a>,
+    /// Experimental translation path attempts to avoid SPIR-V and work with Naga IR directly.
+    pub experimental_translation: bool,
```
grovesNL, line 697: If we're mostly using this internally, then another option is to enable this through an environment variable. Then we wouldn't need to change the API later when we're ready to remove the SPIR-V path

grovesNL (approved):

cwfitzgerald (approved):
LGTM!

kvark:
Bors r=cwfitzgerald,grovesNL

PR #664, opened 2020-12-05
Integrate ETC and ASTC textures in Skybox example

cwfitzgerald:
Depends on https://github.com/gfx-rs/wgpu/pull/1074

Converts skybox example to mip-mapped files in:
 - Bgra8
 - Bc1
 - ETC2
 - ASTC4x4

Looks quite good now with mipmapping and trilinear filtering even at small resolutions.


kvark (approved):
Looks great! My only concern is the texture size here

cwfitzgerald:
bors r=kvark

PR #669, opened 2020-12-07
Update wgpu with Naga changes, replace hello-triangle shaders with WGSL

kvark:


cwfitzgerald (approved):
Very fancy! First wgsl shader!

bors r+

PR #672, opened 2020-12-11
Add write_whole_texture_mips Helper

cwfitzgerald:
Adds a helper in QueueExt for uploading an entire texture with all of its mips. Name very bikeshedable.

### DIFF
--- a/examples/framework.rs
+++ b/examples/framework.rs
@@ -168,7 +168,7 @@ async fn setup<E: Example>(title: &str) -> Setup {
                 label: None,
                 features: (optional_features & adapter_features) | required_features,
                 limits: needed_limits,
-                shader_validation: true,
+                shader_validation: false,
             },
             trace_dir.ok().as_ref().map(std::path::Path::new),
         )
--- a/examples/skybox/main.rs
+++ b/examples/skybox/main.rs
@@ -165,16 +165,6 @@ impl framework::Example for Skybox {
         let layer_size = wgpu::Extent3d { depth: 1, ..size };
         let max_mips = layer_size.max_mips();
 
-        let texture = device.create_texture(&wgpu::TextureDescriptor {
-            size,
-            mip_level_count: max_mips as u32,
-            sample_count: 1,
-            dimension: wgpu::TextureDimension::D2,
-            format: skybox_format,
-            usage: wgpu::TextureUsage::SAMPLED | wgpu::TextureUsage::COPY_DST,
-            label: None,
-        });
-
         log::debug!(
             "Copying {:?} skybox images of size {}, {}, 6 with {} mips to gpu",
             skybox_format,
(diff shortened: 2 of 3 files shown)


src/util/mod.rs
```diff
@@ -99,3 +99,92 @@ pub struct BufferInitDescriptor<'a> {
...
+
+        let (layer_iterations, mip_extent) =
+            if texture.texture_dimension == crate::TextureDimension::D3 {
+                (1, size)
+            } else {
+                (size.depth, crate::Extent3d { depth: 1, ..size })
+            };
+
+        let mut binary_offset = 0;
+        for layer in 0..layer_iterations {
```
kvark, line 136: one important question here is - what guides the decision on the order between mipmap levels and array layers? I.e. why does this code iterate the layers in the outer loop, and levels in the inner loop, specifically?

src/util/mod.rs
```diff
@@ -99,3 +99,92 @@ pub struct BufferInitDescriptor<'a> {
...
+                );
+
+                binary_offset = end_offset;
+            }
+        }
+    }
+}
+
+/// View of a texture which can be used to copy to/from a buffer/texture.
+pub struct MipmappedTextureCopyView<'a> {
```
kvark, line 181: Looking how this repeats the `TextureDescriptor` I can't help but wonder. Maybe a better API here would be to have `DeviceExt::create_texture_with_data(desc, queue, data)`?

src/util/mod.rs
```diff
@@ -99,3 +99,92 @@ pub struct BufferInitDescriptor<'a> {
...
+                            x: 0,
+                            y: 0,
+                            z: layer,
+                        },
+                    },
+                    &data[binary_offset..end_offset],
+                    crate::TextureDataLayout {
+                        offset: 0,
+                        bytes_per_row,
+                        rows_per_image: 0,
```
kvark, line 169: this, in particular, is demonstrating the benefit of iterating over the mipmaps in the outer loop. This way, you'd not need to iterate the layers at all, and just upload all layers at once.

cwfitzgerald:
All comments addressed/discussed :)

examples/framework.rs
```diff
@@ -168,7 +168,7 @@ async fn setup<E: Example>(title: &str) -> Setup {
                 label: None,
                 features: (optional_features & adapter_features) | required_features,
                 limits: needed_limits,
-                shader_validation: true,
+                shader_validation: false,
```
kvark, line 171: I managed to get Naga to a point where it can handle the skybox example when ported to WGSL, so we'll be able to turn this back on with this upcoming change

examples/skybox/main.rs
```diff
@@ -193,60 +183,19 @@ impl framework::Example for Skybox {
...
-                        bytes_per_row,
-                        rows_per_image: 0,
-                    },
-                    mip_physical,
-                );
-
-                binary_offset = end_offset;
-            }
-        }
+        let texture = device.create_texture_with_data(
```
kvark, line 186: this looks slick!

kvark (approved):
thanks!
I'm a bit sore from the tools doing layers then mipmaps. If the order was different, it would be much easier here for the users, but oh well...
bors r+

examples/framework.rs
cwfitzgerald: fwiw, I didn't intend on committing this, it just got caught up in my force pushing nonsense

PR #677, opened 2020-12-14
Update to latest wgpu master

yzsolt:
- Update to latest `wgpu` master
- Fix build error caused by https://github.com/gfx-rs/wgpu/pull/1077

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,13 +25,13 @@ vulkan-portability = ["wgc/gfx-backend-vulkan"]
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies.wgc]
 package = "wgpu-core"
 git = "https://github.com/gfx-rs/wgpu"
-rev = "4846e41eb0ffcaafe148833bc25c3a18abce1b26"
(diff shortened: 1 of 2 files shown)


yzsolt:
Haven't dug myself into how `RenderInner` and the ID machinery works, so let me know if there's a better fix!

kvark (commented):
Thank you!
There is also one nasty hidden place that needs to be updated - https://github.com/gfx-rs/wgpu-rs/blob/2ef725065e68164cced1551c7a2540523eb0ca77/Cargo.toml#L51

yzsolt:
> There is also one nasty hidden place that needs to be updated -

Shouldn't it point to the same revision of `gfx` as `wgpu-core` master? Because it already does that, I'm not sure what it should be updated to otherwise.

kvark:
> > There is also one nasty hidden place that needs to be updated -
> 
> Shouldn't it point to the same revision of `gfx` as `wgpu-core` master? Because it already does that, I'm not sure what it should be updated to otherwise.

Yes, it needs to point to the same thing. If it already does, then we are all good.
bors r+

PR #679, opened 2020-12-17
WGSL shader conversion for the examples

kvark:
Blocked on:
  - ~~https://github.com/gfx-rs/naga/issues/335~~
  - ~~https://github.com/gfx-rs/naga/issues/336~~
  - ~~https://github.com/gfx-rs/wgpu/issues/1095~~
  - ~~https://github.com/gfx-rs/naga/pull/360~~
  - https://github.com/gfx-rs/gfx/pull/3622
  - ~~GL backend doing the program link check earlier~~

Converted:
- [x] boids
- [x] cube
- [x] mipmap
- [x] msaa-line
- [x] shadow
- [x] skybox
- [ ] texture-array
- [ ] water


baptistemanson:
I tested all examples on develop and on this branch on Windows 10 using a RTX 2080 SUPER.
I have 3 regressions:

Cube seems to have some coords issues:
image


Hello triangle
thread 'main' panicked at 'Failed to find an appropriate adapter', examples\hello-triangle\main.rs:19:10

MipMap doesn't display
image

They all work properly on origin/develop.

I'll dig more into those later!

baptistemanson:
Thanks for the force push, it solved the hello-triangle issue.
When removing the projection matrix from the cube example, everything works as expected. I suspect changing from GLSL to WGSL has maybe change the convention around matrices (row first <> column first or something similar?)
image

examples/cube/main.rs
```diff
@@ -274,19 +271,27 @@ impl framework::Example for Example {
             }],
         };
 
-        let vs_module = device.create_shader_module(&wgpu::include_spirv!("shader.vert.spv"));
-        let fs_module = device.create_shader_module(&wgpu::include_spirv!("shader.frag.spv"));
+        let mut flags = wgpu::ShaderFlags::VALIDATION;
```
grovesNL, line 274: nit: we could temporarily have a helper function for this in framework to remove backend matching from the example code

examples/cube/main.rs
```diff
@@ -274,19 +271,27 @@ impl framework::Example for Example {
...
+        match adapter.get_info().backend {
+            wgpu::Backend::Metal => flags |= wgpu::ShaderFlags::EXPERIMENTAL_TRANSLATION,
+            _ => (), //TODO: https://github.com/gfx-rs/naga/issues/335
+        }
+        let shader = device.create_shader_module(&wgpu::ShaderModuleDescriptor {
+            label: None,
+            source: wgpu::ShaderSource::Wgsl(Cow::Borrowed(include_str!("shader.wgsl"))),
```
grovesNL, line 281: nit: this will break the examples on the web backend for a while, though maybe we could start generating SPIR-V internally soon or temporarily keep the spv files as a fallback path for the web

grovesNL (approved):
Awesome :+1: great to see this coming along

examples/cube/main.rs
kvark: I considered this, but the problem is that each example can be limited by the specifics of the shaders it needs. So one can enable experimental translation on Metal, for example, another can not. So for now I think having them separate is better.

examples/cube/main.rs
kvark: Good point! I'll hook Naga up to the web side and convert it.

grovesNL (approved):

examples/cube/main.rs
grovesNL: Yeah for now it's probably fine like this – it might be nice to have a helper once it's ready on all examples

kvark:
Going to land this right after #734

kvark:
A little functional changes here
  - switched "boids" to use read-only storage for the source array
  - switched "shadow" to use the storage with runtime-sized array instead of a uniform structure

bors r=grovesNL

examples/shadow/shader.wgsl
```diff
@@ -0,0 +1,114 @@
...
+fn vs_main() {
+    const w: mat4x4<f32> = u_entity.world;
+    out_normal_vs = mat3x3<f32>(w.x.xyz, w.y.xyz, w.z.xyz) * vec3<f32>(in_normal.xyz);
+    out_position_vs = u_entity.world * vec4<f32>(in_position);
+    out_position = u_globals.view_proj * out_position_vs;
+}
+
+// fragment shader
+
+[[block]]
```
dneto0, line 48: FYI. It's invalid to declare an array of block-decorated struct.
WGSL doesn't have array-of-buffers, but Vulkan SPIR-V does

kvark: Right. At the time this was written, our `[[block]]` semantics was a bit different. Today, there is no `[[block]]` on it https://github.com/gfx-rs/wgpu-rs/blob/69be1ff88a073b3a7f9470044834e75c22f7b4ad/examples/shadow/shader.wgsl#L54

PR #680, opened 2020-12-17
Update wgpu with Naga fixes and OpenGL backend

### LINKED ISSUE #607: Panic on drop of a surface with the GL backend
`thread 'main' panicked at 'assertion failed: (left != right)`

I notice that on the master branch with the latest commit enabling the GL backend, I get the above panic with the following code:

```rust
let instance = Instance::new(BackendBit::GL);
let surface = Surface::new(&window);
drop(surface); // panic here
```

The panic occurs on line 135 of the surfman module "platform/generic/egl/context."

Display server: X11
WM: i3

kvark:
This PR makes it easier to build wgpu-rs on Linux, since it doesn't require pkg-config into EGL.
Fixes #607
Fixes https://github.com/gfx-rs/wgpu/issues/1063

### DIFF
--- a/examples/hello-triangle/main.rs
+++ b/examples/hello-triangle/main.rs
@@ -25,7 +25,6 @@ async fn run(event_loop: EventLoop<()>, window: Window, swapchain_format: wgpu::
                 label: None,
                 features: wgpu::Features::empty(),
                 limits: wgpu::Limits::default(),
-                shader_validation: true,
             },
             None,
         )
@@ -36,7 +35,7 @@ async fn run(event_loop: EventLoop<()>, window: Window, swapchain_format: wgpu::
     let shader = device.create_shader_module(&wgpu::ShaderModuleDescriptor {
         label: None,
         source: wgpu::ShaderSource::Wgsl(Cow::Borrowed(include_str!("shader.wgsl"))),
-        experimental_translation: true,
+        flags: wgpu::ShaderFlags::all(),
     });
 
     let pipeline_layout = device.create_pipeline_layout(&wgpu::PipelineLayoutDescriptor {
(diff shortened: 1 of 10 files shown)


examples/hello-triangle/main.rs
```diff
@@ -7,7 +7,7 @@ use winit::{
 
 async fn run(event_loop: EventLoop<()>, window: Window, swapchain_format: wgpu::TextureFormat) {
     let size = window.inner_size();
-    let instance = wgpu::Instance::new(wgpu::BackendBit::PRIMARY);
+    let instance = wgpu::Instance::new(wgpu::BackendBit::GL);
```
cwfitzgerald, line 10: This seems to have snuck into the commit

src/macros.rs
```diff
@@ -52,7 +52,7 @@ macro_rules! include_spirv {
             $crate::ShaderModuleDescriptor {
                 label: Some($($token)*),
                 source: $crate::util::make_spirv(include_bytes!($($token)*)),
-                experimental_translation: false,
+                flags: $crate::ShaderFlags::VALIDATION,
```
cwfitzgerald, line 55: Does validation still fall through if it can't parse stuff? If not I'd rather this be user controllable.

cwfitzgerald (changes requested):

src/macros.rs
kvark: It still falls, yes. But at the same time - the flag is fully user-controllable. The macro is just a helper, it's not the actual wgpu-rs API.

cwfitzgerald: Yeah I know, it just seems like if there is a macro available, users use it, so it's worthwhile keeping it in mind. But it failing out is fine.

kvark:
The warning on exit will be fixed in next updates, see upstream in https://github.com/zakarumych/gpu-descriptor/pull/9
The WGSL conversion is also going to happen as the next step, see #679 
bors r=cwfitzgerald

kvark:
bors retry

kvark:
bors r=cwfitzgerald

PR #684, opened 2020-12-19
First step for WebGL support.

VincentFTS:
This permits to test WebGL backend in Wgpu-rs.
Next step is to test examples to make them work with WebGL backend !

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,12 +21,20 @@ trace = ["serde", "wgc/trace"]
 replay = ["serde", "wgc/replay"]
 # Make Vulkan backend available on platforms where it is by default not, e.g. macOS
 vulkan-portability = ["wgc/gfx-backend-vulkan"]
+webgl = ["wgc"]
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies.wgc]
 package = "wgpu-core"
 git = "https://github.com/gfx-rs/wgpu"
 rev = "4ebe1f50b057046e4d4f015eb006330d62f5fe91"
 features = ["raw-window-handle"]
+
+[target.'cfg(target_arch = "wasm32")'.dependencies.wgc]
+package = "wgpu-core"
+git = "https://github.com/gfx-rs/wgpu"
+rev = "4ebe1f50b057046e4d4f015eb006330d62f5fe91"
+features = ["raw-window-handle"]
+optional = true
 
 [dependencies.wgt]
 package = "wgpu-types"
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -1,15 +1,15 @@
-#[cfg(target_arch = "wasm32")]
+#[cfg(all(target_arch = "wasm32", not(feature = "webgl")))]
 mod web;
-#[cfg(target_arch = "wasm32")]
+#[cfg(all(target_arch = "wasm32", not(feature = "webgl")))]
 pub(crate) use web::{BufferMappedRange, Context};
 
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(any(not(target_arch = "wasm32"), feature = "webgl"))]
 mod direct;
-#[cfg(not(target_arch = "wasm32"))]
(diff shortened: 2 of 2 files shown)


examples/framework.rs
```diff
@@ -108,6 +108,7 @@ async fn setup<E: Example>(title: &str) -> Setup {
 
     log::info!("Initializing the surface...");
 
+    #[cfg(not(webgl))]
     let backend = if let Ok(backend) = std::env::var("WGPU_BACKEND") {
```
kvark, line 112: we should still allow `BROWSER_WEBGPU` being picked if compiling for wasm

build.rs
```diff
@@ -0,0 +1,11 @@
+fn main() {
+    // Setup cfg aliases
+    cfg_aliases::cfg_aliases! {
+        // Vendors/systems
+        wasm: { target_arch = "wasm32" },
```
grovesNL, line 5: Because we only use these in a few places, maybe we could just write them inline instead of having build.rs with aliases

examples/framework.rs
VincentFTS: webgl alias is not wasm, it is wasm PLUS webgl feature, so if compiled for wasm without webgl feature, WebGPU will be allowed.

build.rs
VincentFTS: I thought `#[cfg(all(target_arch = "wasm32", not(feature = "webgl")))]` was not very explicit for WebGPU cases.
Let me know if you still want me to remove this.

examples/framework.rs
grovesNL: Ideally I think we'd like to allow both WebGPU and WebGL (as a fallback) to be used in a single wasm file, instead of only WebGL or WebGPU

build.rs
grovesNL: Maybe we could avoid explicit checks for WebGPU because it can always be included (i.e. try to use WebGPU first, then WebGL if the feature is enabled)

VincentFTS: Outch, it seems that Wgpu-rs has not been designed to permit this …
backend/mod.rs is clearly a toggle between WebGPU backend and others, so making a build enabling WebGPU AND another backend is kind of a big rework …

examples/framework.rs
VincentFTS: Outch, it seems that Wgpu-rs has not been designed to permit this …
backend/mod.rs is clearly a toggle between WebGPU backend and others, so making a build enabling WebGPU AND another backend is kind of a big rework …

grovesNL: Oh right, I forgot about that. We can consider doing that later then, but let's just be sure to write the feature checks in a way that would still allow it

examples/framework.rs
```diff
@@ -108,6 +108,7 @@ async fn setup<E: Example>(title: &str) -> Setup {
 
     log::info!("Initializing the surface...");
 
+    #[cfg(not(feature = "webgl"))]
```
grovesNL, line 111: I think we can probably leave this as it was originally, because examples can use WebGL by allowing secondary backends like GL (e.g. `wgpu::BackendBit::all()` for examples that support DX11 or GL)

grovesNL (approved):
Looks great, thank you! Just noticed one small change we could probably revert

examples/framework.rs
VincentFTS: Tell me if I’m wrong, I think we need WebGPU to be "deactivated", if not, WebGL will never be used.
I’ve never succeeded in using the environment variable in a browser context anyway …

grovesNL: I think it should already be deactivated because of how `web` is guarded by `#[cfg(all(target_arch = "wasm32", not(feature = "webgl")))]`, right?

In other words, when `wgpu::BackendBit::all()` and `--features webgl` are used, we will search through the possible backends to see which one can be used. In that case, we should only see that the GL backend (WebGL) is available, so we'll choose that backend.

grovesNL:
Thanks!
bors r+

PR #685, opened 2020-12-20
Replace futures crate with pollster.

### LINKED ISSUE #628: Reduce `futures` dependencies
Looks like the `futures` drag a bunch of other stuff. Perhaps, we can make them optional? Or only depend on `futures-core`? or nothing at all, and implement the needed stuff directly on `std`.

parasyte:
This PR removes all of the `futures` dependencies. `std::future` does not contain a lot of useful helpers available in `futures`. The obvious ones are `join_all`, `FutureExt::map`, and `block_on`.

- `join_all` is replaced with a `Vec<T>` and an `async` block.
- `FuturesExt::map` in the web backend is replaced by rolling the `map` function into the `MakeSendFuture` type.
- `block_on` is provided by `pollster`.

The original code using `join_all` ignored the result type yielded by the Future from `map_async`. This code does the same, but makes dropping the result a little more obvious.

These should not be troublesome. Figured I would call them out anyway.

The last big change is replacing `futures-executor` in the examples with `async-executor`. A new concrete `Spawner` type is used in the example framework instead of an implementation of `futures_task::LocalSpawn`.

Fixes #628

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,6 @@ rev = "4ebe1f50b057046e4d4f015eb006330d62f5fe91"
 
 [dependencies]
 arrayvec = "0.5"
-futures = { version = "0.3", default-features = false, features = ["std"] }
 parking_lot = "0.11"
 raw-window-handle = "0.3"
 smallvec = "1"
@@ -60,7 +59,10 @@ rand = { version = "0.7.2", features = ["wasm-bindgen"] }
 bytemuck = { version = "1.4", features = ["derive"] }
 noise = "0.6"
 ddsfile = "0.4"
-futures = { version = "0.3", default-features = false, features = ["std", "executor"] }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
+async-executor = "1.0"
+pollster = "0.2"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies.wgpu-subscriber]
 version = "0.1"
--- a/src/backend/web.rs
+++ b/src/backend/web.rs
@@ -7,7 +7,6 @@ use crate::{
     TextureViewDescriptor, TextureViewDimension,
 };
 
-use futures::FutureExt;
 use std::{
     fmt,
     future::Future,
@@ -55,19 +54,34 @@ pub(crate) struct RenderBundleEncoder(web_sys::GpuRenderBundleEncoder);
 // This is safe on wasm32 *for now*, but similarly to the unsafe Send impls for the handle type
 // wrappers, the full story for threading on wasm32 is still unfolding.
 
-pub(crate) struct MakeSendFuture<F>(F);
+pub(crate) struct MakeSendFuture<F, M> {
+    future: F,
+    map: M,
+}
 
-impl<F: Future> Future for MakeSendFuture<F> {
-    type Output = F::Output;
+impl<F: Future, M: Fn(F::Output) -> T, T> Future for MakeSendFuture<F, M> {
+    type Output = T;
 
     fn poll(self: Pin<&mut Self>, cx: &mut task::Context) -> Poll<Self::Output> {
         // This is safe because we have no Drop implementation to violate the Pin requirements and
         // do not provide any means of moving the inner future.
-        unsafe { self.map_unchecked_mut(|s| &mut s.0) }.poll(cx)
+        unsafe {
+            let this = self.get_unchecked_mut();
+            match Pin::new_unchecked(&mut this.future).poll(cx) {
+                task::Poll::Ready(value) => task::Poll::Ready((this.map)(value)),
+                task::Poll::Pending => task::Poll::Pending,
+            }
+        }
     }
 }
 
-unsafe impl<F> Send for MakeSendFuture<F> {}
+impl<F, M> MakeSendFuture<F, M> {
+    fn new(future: F, map: M) -> Self {
+        Self { future, map }
+    }
+}
+
+unsafe impl<F, M> Send for MakeSendFuture<F, M> {}
 
 impl crate::ComputePassInner<Context> for ComputePass {
     fn set_pipeline(&mut self, pipeline: &Sendable<web_sys::GpuComputePipeline>) {
@@ -800,7 +814,6 @@ fn map_map_mode(mode: crate::MapMode) -> u32 {
 }
 
 type JsFutureResult = Result<wasm_bindgen::JsValue, wasm_bindgen::JsValue>;
-type FutureMap<T> = futures::future::Map<wasm_bindgen_futures::JsFuture, fn(JsFutureResult) -> T>;
 
 fn future_request_adapter(result: JsFutureResult) -> Option<Sendable<web_sys::GpuAdapter>> {
     match result {
@@ -850,11 +863,18 @@ impl crate::Context for Context {
 
     type SwapChainOutputDetail = SwapChainOutputDetail;
 
-    type RequestAdapterFuture = MakeSendFuture<FutureMap<Option<Self::AdapterId>>>;
-    type RequestDeviceFuture = MakeSendFuture<
-        FutureMap<Result<(Self::DeviceId, Self::QueueId), crate::RequestDeviceError>>,
+    type RequestAdapterFuture = MakeSendFuture<
+        wasm_bindgen_futures::JsFuture,
+        fn(JsFutureResult) -> Option<Self::AdapterId>,
     >;
-    type MapAsyncFuture = MakeSendFuture<FutureMap<Result<(), crate::BufferAsyncError>>>;
+    type RequestDeviceFuture = MakeSendFuture<
+        wasm_bindgen_futures::JsFuture,
+        fn(JsFutureResult) -> Result<(Self::DeviceId, Self::QueueId), crate::RequestDeviceError>,
+    >;
+    type MapAsyncFuture = MakeSendFuture<
+        wasm_bindgen_futures::JsFuture,
+        fn(JsFutureResult) -> Result<(), crate::BufferAsyncError>,
+    >;
 
     fn init(_backends: wgt::BackendBit) -> Self {
         Context(web_sys::window().unwrap().navigator().gpu())
@@ -900,8 +920,10 @@ impl crate::Context for Context {
         };
         mapped_options.power_preference(mapped_power_preference);
         let adapter_promise = self.0.request_adapter_with_options(&mapped_options);
-        MakeSendFuture(
-            wasm_bindgen_futures::JsFuture::from(adapter_promise).map(future_request_adapter),
+
+        MakeSendFuture::new(
+            wasm_bindgen_futures::JsFuture::from(adapter_promise),
+            future_request_adapter,
         )
     }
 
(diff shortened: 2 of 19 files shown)


src/backend/web.rs
```diff
@@ -800,7 +799,7 @@ fn map_map_mode(mode: crate::MapMode) -> u32 {
 }
 
 type JsFutureResult = Result<wasm_bindgen::JsValue, wasm_bindgen::JsValue>;
-type FutureMap<T> = futures::future::Map<wasm_bindgen_futures::JsFuture, fn(JsFutureResult) -> T>;
+type MakeBoxedSendFuture<T> = MakeSendFuture<std::pin::Pin<Box<dyn Future<Output = T>>>>;
```
grovesNL, line 802: Does this allocate now? Could we keep the old signature somehow?

parasyte: It does allocate now. I have to use `Box<dyn Future<...>>` to return an `async` block. Which I am only doing because it's nicer to use than creating a concrete type that implements `Future`. This is definitely something than can be improved though.

parasyte: @grovesNL I copy-pasted `FuturesExt::Map` to get this working. Fair tradeoff, I suppose.

kvark:
@parasyte thank you for taking a stab at this! I'm excite to see the reduced dependencies list. This and the formely surfman dependencies were identified as the biggest sources of dependencies for us.
@grovesNL would you be able to make a final call on this? I believe you are in the best position to do so, but of course it depends on your availability.

src/backend/web.rs
```diff
@@ -69,6 +69,76 @@ impl<F: Future> Future for MakeSendFuture<F> {
...
 
+fn future_map<Fut, F, O>(fut: Fut, f: F) -> Map<Fut, F>
+where
+    Fut: Future,
+    F: FnOnce(Fut::Output) -> O,
+{
+    Map::new(fut, f)
+}
+
+pin_project! {
```
grovesNL, line 80: Could we avoid `pin_project` here too? I think we should be able to do something simpler here like we have for `MakeSendFuture` (possibly even collapsing them together)

parasyte: We can, but `pin-project-lite` is already a dependency of `naga`, so I am not sure what we gain by making that change. Merging with `MakeSendFuture` sounds like a good idea!

kvark: I'm curious as to where Naga would need it.
Anyway, Naga is not a wgpu-rs dependency for the `wasm32` target.

parasyte: I misspoke. not Naga. But `tracing`:

```
tracing v0.1.22
├── cfg-if v1.0.0
├── pin-project-lite v0.2.0
├── tracing-attributes v0.1.11 (proc-macro)
│   ├── proc-macro2 v1.0.24
│   │   └── unicode-xid v0.2.1
│   ├── quote v1.0.7
│   │   └── proc-macro2 v1.0.24 (*)
│   └── syn v1.0.54
│       ├── proc-macro2 v1.0.24 (*)
│       ├── quote v1.0.7 (*)
│       └── unicode-xid v0.2.1
└── tracing-core v0.1.17
    └── lazy_static v1.4.0
```

grovesNL: Alright :+1: I think ideally tracing shouldn't be a dependency for wasm (at least by default), so removing the dependency would still be useful to keep wasm builds smaller. But we could probably tackle this in a follow-up PR instead

grovesNL (approved):

grovesNL:
Oh I guess the only other concern is that we have (temporarily) copy-pasted code from `futures-util` – are there any licensing concerns because of that?

src/backend/web.rs
kvark: Agreed, we don't need to log anything with tracing when targeting the Web, but can follow-up.
Although, now, instead of just cfg-blocking `tracing` usage, we'd have to figure out how to rewrite the pin section...
So it would be nicer to deal with it right away.

parasyte:
`futures-util` is dual-licensed under Apache 2.0 and MIT:

- https://github.com/rust-lang/futures-rs/blob/master/LICENSE-APACHE
- https://github.com/rust-lang/futures-rs/blob/master/LICENSE-MIT

I don't know how well these are compatible with MPL. If licensing is an issue, the `Map` and its supporting code will need to be replaced with I-don't-know-what. :)

grovesNL:
APACHE 2/MIT are ok, I was just wondering how to properly attribute the source code to the original authors (e.g. copyright notice in that file or anything else we need). It might be easier to replace it in this PR anyway to avoid the copy-pasted code.

parasyte:
@grovesNL I was able to combine `MakeSendFuture` and `Map` into a single type. This last commit also removes all of the code borrowed from the `Futures` crate.

I do need someone to thoroughly check my work here, however. I'm not entirely familiar with `Pin` and `Unpin`. I used the latter to avoid some extra unsafe blocks. AFAICT this is ok according to the docs, since it only accesses the `map` field non-structurally. I also added an assertion to panic if the future is ever polled after returning `Ready`. If any of this needs to be changed, I'll be glad to.

Last thing to mention is that I can squash merge all of these commits if you would like.

parasyte:
Rebased and force-pushed as a squashed commit.

grovesNL:
Thank you for the update!

I'll try to review the async changes soon – I'm taking a moment to get caught up on some of the futures changes over the past couple years. If anyone else is familiar with futures/pin and able to review it in the meantime, please go ahead :+1:

src/backend/web.rs
```diff
@@ -55,19 +54,46 @@ pub(crate) struct RenderBundleEncoder(web_sys::GpuRenderBundleEncoder);
 // This is safe on wasm32 *for now*, but similarly to the unsafe Send impls for the handle type
 // wrappers, the full story for threading on wasm32 is still unfolding.
 
-pub(crate) struct MakeSendFuture<F>(F);
+pub(crate) struct MakeSendFuture<F, M> {
+    future: F,
+    map: Option<M>,
```
grovesNL, line 59: We probably don't need to worry about the `map` being used multiple times, because the caller should stop polling once `Poll::Ready` happens anyway.  So maybe we could remove the `Option` here to just have `M`, and also remove the `assert`

src/backend/web.rs
```diff
@@ -55,19 +54,46 @@ pub(crate) struct RenderBundleEncoder(web_sys::GpuRenderBundleEncoder);
...
+impl<F, M> MakeSendFuture<F, M> {
+    fn new(future: F, map: M) -> Self {
+        Self {
+            future,
+            map: Some(map),
+        }
+    }
+}
+
+impl<F, M> Unpin for MakeSendFuture<F, M> {}
```
grovesNL, line 95: We shouldn't need `Unpin` here

src/backend/web.rs
```diff
@@ -55,19 +54,46 @@ pub(crate) struct RenderBundleEncoder(web_sys::GpuRenderBundleEncoder);
...
+        assert!(
+            self.as_mut().get_mut().map.is_some(),
+            "MakeSendFuture must not be polled after returning Ready"
+        );
 
-    fn poll(self: Pin<&mut Self>, cx: &mut task::Context) -> Poll<Self::Output> {
         // This is safe because we have no Drop implementation to violate the Pin requirements and
         // do not provide any means of moving the inner future.
-        unsafe { self.map_unchecked_mut(|s| &mut s.0) }.poll(cx)
+        let future = unsafe { self.as_mut().map_unchecked_mut(|s| &mut s.future) };
```
grovesNL, line 73: We could probably simplify the access to `T` a little, e.g.:
```rust
fn poll(self: Pin<&mut Self>, cx: &mut task::Context) -> Poll<Self::Output> {
    // This is safe because we have no Drop implementation to violate the Pin requirements and
    // do not provide any means of moving the inner future.
    unsafe {
        let this = self.get_unchecked_mut();
        match Pin::new_unchecked(&mut this.future).poll(cx) {
            task::Poll::Ready(value) => task::Poll::Ready((this.map)(value)),
            task::Poll::Pending => task::Poll::Pending,
        }
    }
}
```

src/util/belt.rs
```diff
@@ -133,18 +132,17 @@ impl StagingBelt {
...
-                })
-        }))
-        .map(|_| ())
+        let sender = self.sender.clone();
+        let chunks = self.closed_chunks.drain(..).collect::<Vec<_>>();
+
+        async move {
+            for chunk in chunks {
+                // The result is ignored
+                chunk.buffer.slice(..).map_async(MapMode::Write).await.ok();
```
grovesNL, line 141: If I'm understand correctly it seems like this would wait for every chunk in order. Instead I think we'd want to start running each future (e.g. `map_async`) and `await` them in parallel/poll them all in a loop (but not blocking on any of them), which is what `join_all` seems to do (https://docs.rs/futures-util-preview/0.3.0-alpha.19/src/futures_util/future/join_all.rs.html#115-122)

Cargo.toml
```diff
@@ -60,7 +59,10 @@ rand = { version = "0.7.2", features = ["wasm-bindgen"] }
 bytemuck = { version = "1.4", features = ["derive"] }
 noise = "0.6"
 ddsfile = "0.4"
-futures = { version = "0.3", default-features = false, features = ["std", "executor"] }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
+async-executor = "1.0"
```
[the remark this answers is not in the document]
parasyte: `async_executor` is used in the example framework: https://github.com/gfx-rs/wgpu-rs/pull/685/files#diff-e3c88eb31749039885929e401e55a90a5f69a0b8cbbcf46cb2df072aafbff069R294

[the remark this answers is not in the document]
grovesNL: Ah ok, I missed that. Sounds good

src/util/belt.rs
parasyte: @grovesNL Glad you caught that. The latest commit rewrites this one to return a future that polls its inner futures similarly to `join_all`.

I've only tested it with the `skybox` demo, but it is properly allowing chunks to be reused in the `StagingBelt`, and I don't see any obvious flaws with the implementation.

src/util/belt.rs
```diff
@@ -2,9 +2,42 @@ use crate::{
...
     CommandEncoder, Device, MapMode,
 };
-use futures::{future::join_all, FutureExt};
+use std::pin::Pin;
+use std::task::{self, Poll};
 use std::{future::Future, sync::mpsc};
 
+// Given a vector of futures, poll each in parallel until all are ready.
+struct Join<F> {
+    futures: Vec<Option<F>>,
```
grovesNL, line 11: nit: technically I guess we could just remove items directly from `futures` instead of setting them to `None`, but either seems alright

grovesNL (approved):
The updated PR looks great, thank you! Could you squash the commits once more before we merge?

src/util/belt.rs
parasyte: I thought about that, but IIUC, that constitutes a move which would violate `Pin` requirements. Am I wrong?

parasyte:
@grovesNL Thanks for the reviews. This should be ready to go.

grovesNL:
🎉 
bors r+

src/util/belt.rs
grovesNL: Yeah I think that's right – it would probably have to be behind another pointer type, so maybe not as straightforward as the `Option` approach anyway

PR #690, opened 2020-12-24
Add `RenderEncoder`

0x182d4454fb211940:
It seems like `RenderPass` and `RenderBundleEncoder` share many methods in common (`set_pipeline`, `draw_indexed`, etc). By creating a shared trait, it allows more reusable code. For instance:

```rust
impl RedCircle {
    fn new(...) -> Self { .. }
    fn draw<'a>(&'a self, encoder: &mut impl RenderEncoder<'a>) { .. }
}
```

This code would then work for both structs.

Moving the previous methods from their `impl`s to the trait's would break previous code, as to use these methods code would now have to import `RenderEncoder`. To avoid this, it is implemented by calling the direct methods.

It may be worth considering moving this to `wgpu::util`.


kvark:
Thank you for experimenting with this!
I do see the benefit of allowing generic code over bundles. At the same time, most users aren't going to benefit from it, and will be burdened by the `use` statement. Hence, your PR essentially makes those function signatures to be duplicated. So the main question is - where should this trait live: either in our `utils` module, or in a separate wgpu-utils of sorts...

src/lib.rs
```diff
@@ -2575,6 +2575,183 @@ impl<'a> RenderBundleEncoder<'a> {
...
+        &mut self,
+        indirect_buffer: &'a Buffer,
+        indirect_offset: BufferAddress,
+    );
+}
+
+impl<'a> RenderEncoder<'a> for RenderPass<'a> {
+    #[inline(always)]
+    fn set_bind_group(&mut self, index: u32, bind_group: &'a BindGroup, offsets: &[DynamicOffset]) {
+        (self as &mut RenderPass<'a>).set_bind_group(index, bind_group, offsets);
```
kvark, line 2665: let's use a different form of this: `RenderPass::set_bind_group(self, index, bind_group, offsets)`

cwfitzgerald: Better yet: `Self::set_bind_group(..)`

src/lib.rs
```diff
@@ -2575,6 +2575,183 @@ impl<'a> RenderBundleEncoder<'a> {
...
+    fn draw_indexed_indirect(
+        &mut self,
+        indirect_buffer: &'a Buffer,
+        indirect_offset: BufferAddress,
+    ) {
+        (self as &mut RenderPass<'a>).draw_indexed_indirect(indirect_buffer, indirect_offset);
+    }
+}
+
+impl<'a> RenderEncoder<'a> for RenderBundleEncoder<'a> {
```
cwfitzgerald, line 2708: Could we add the features (like push constants) that are in common between renderbundles and renderpasses?

src/lib.rs
```diff
@@ -2575,6 +2575,183 @@ impl<'a> RenderBundleEncoder<'a> {
     }
 }
 
+/// Methods shared by `RenderPass` and `RenderBundleEncoder`
+pub trait RenderEncoder<'a> {
```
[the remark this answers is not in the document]
0x182d4454fb211940: Done.

src/lib.rs
0x182d4454fb211940: Done

src/lib.rs
0x182d4454fb211940: I think the only one is `set_push_constants` (which I've added).

kvark (approved):
bors r+

PR #693, opened 2021-01-01
Expose get_swap_chain_preferred_format

### LINKED ISSUE #691: Add 'swap_chain_get_preferred_format' to wgpu-rs (available in wgpu-core but not exposed)
> we actually have `swap_chain_get_preferred_format` in `wgpu-core`. We should expose this in `wgpu-rs`
> _Originally posted by @kvark in https://github.com/gfx-rs/wgpu-rs/pull/511#discussion_r473203364_

This functionality was commited in https://github.com/gfx-rs/wgpu/commit/6f1d6145f555ebbe89b0cca94288990cc0368223 but never exposed to wgpu-rs.

Problem: Currently, examples make use of `wgpu::TextureFormat::Bgra8UnormSrgb` but that's unreliable and [according to the API ](https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-getswapchainpreferredformat)this method is the way of properly obtaining an optimal texture format for the swapchain based on the platform.

manugildev:
Fixes #691

This functionality was committed in gfx-rs/wgpu@6f1d614 but never exposed to `wgpu-rs`.

Problem: Examples made use of `wgpu::TextureFormat::Bgra8UnormSrgb` but that's unreliable and according to the [API](https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-getswapchainpreferredformat) this method is the way of properly obtaining an optimal texture format for the `SwapChain` based on the platform.

- [x]  Expose `get_swap_chain_preferred_format()` on Device
- [x]  Use `TextureFormat::Bgra8Unorm` for [web.rs](https://github.com/gfx-rs/wgpu-rs/blob/e00ef5d5842a1fcfb46bb531789b1c340bdee3ab/src/backend/web.rs#L970)
- [x]  Update [framework.rs](https://github.com/gfx-rs/wgpu-rs/blob/e00ef5d5842a1fcfb46bb531789b1c340bdee3ab/examples/framework.rs#L227) and examples to call this new method

### DIFF
--- a/src/backend/web.rs
+++ b/src/backend/web.rs
@@ -962,6 +962,14 @@ impl crate::Context for Context {
         wgt::Limits::default()
     }
 
+    fn device_get_swap_chain_preferred_format(
+        &self,
+        device: &Self::DeviceId,
+    ) -> wgt::TextureFormat {
+        // TODO: web-sys bindings need to be updated to not return a promise
+        wgt::TextureFormat::Bgra8Unorm
+    }
+
     fn device_create_swap_chain(
         &self,
         device: &Self::DeviceId,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -196,6 +196,7 @@ trait Context: Debug + Send + Sized + Sync {
 
     fn device_features(&self, device: &Self::DeviceId) -> Features;
     fn device_limits(&self, device: &Self::DeviceId) -> Limits;
+    fn device_get_swap_chain_preferred_format(&self, device: &Self::DeviceId) -> TextureFormat;
     fn device_create_swap_chain(
         &self,
         device: &Self::DeviceId,
@@ -1426,6 +1427,11 @@ impl Device {
     /// If any of these limits are exceeded, functions may panic.
     pub fn limits(&self) -> Limits {
         Context::device_limits(&*self.context, &self.id)
+    }
+
+    /// Returns an optimal texture format to use for the [`SwapChain`] with this device.
+    pub fn get_swap_chain_preferred_format(&self) -> TextureFormat {
+        Context::device_get_swap_chain_preferred_format(&*self.context, &self.id)
     }
 
     /// Creates a shader module from either SPIR-V or WGSL source code.
--- a/examples/framework.rs
+++ b/examples/framework.rs
@@ -224,12 +224,7 @@ fn start<E: Example>(
 
     let mut sc_desc = wgpu::SwapChainDescriptor {
         usage: wgpu::TextureUsage::RENDER_ATTACHMENT,
-        // TODO: Allow srgb unconditionally
-        format: if cfg!(target_arch = "wasm32") {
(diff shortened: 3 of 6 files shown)


src/backend/web.rs
```diff
@@ -962,6 +962,16 @@ impl crate::Context for Context {
         wgt::Limits::default()
     }
 
+    fn device_get_swap_chain_preferred_format(
+        &self,
+        device: &Self::DeviceId,
+    ) -> wgt::TextureFormat {
+        let device_promise = surface.0.get_swap_chain_preferred_format(&device.0);
```
kvark, line 969: This doesn't need to be a promise. We can for now leave it as `todo!` until the web-sys bindings are updated with it being non-promise.

src/lib.rs
```diff
@@ -196,6 +196,7 @@ trait Context: Debug + Send + Sized + Sync {
 
     fn device_features(&self, device: &Self::DeviceId) -> Features;
     fn device_limits(&self, device: &Self::DeviceId) -> Limits;
+    fn device_get_swap_chain_preferred_format(&self, device: &Self::DeviceId) -> TextureFormat;
```
kvark, line 199: we still need an exposed method of the `Device` to access this

src/backend/web.rs
manugildev: Should we make it return the default texture format (`Bgra8UnormSrgb` according to [wgpu/src/device/mod.rs](https://github.com/gfx-rs/wgpu/blob/2287ae3f8a7246e8b47b73a1de6fe2410c42b8c8/wgpu-core/src/device/mod.rs#L3853)) like this:
```rust
fn device_get_swap_chain_preferred_format(
    &self,
    device: &Self::DeviceId,
) -> wgt::TextureFormat {
    // TODO: web-sys bindings need to be updated to not return a promise
    wgt::TextureFormat::Bgra8UnormSrgb
}
```

Or just leave it as `todo!` for future work?

src/lib.rs
manugildev: Resolved on https://github.com/gfx-rs/wgpu-rs/pull/693/commits/f343c2384096c4a855f27a4232b9120441fe9af2

src/backend/web.rs
manugildev: Resolved on https://github.com/gfx-rs/wgpu-rs/pull/693/commits/45dd8e17b4ca06b68ce45dfdad911113ecd2fb44

kvark (commented):
Nice, thank you!
Could you squash the commits to proceed?
Also, could you adjust the example framework and the other examples to call this function instead of hard-coding the format?

kvark (approved):
Nice work!
It just occurred to me that this method needs to be on an adapter, not the device. But we'll update after `wgpu-core`, which will update after WebGPU upstream, so no need to block this now.
bors r+

PR #701, opened 2021-01-06
Fix cargo doc warnings due to unresolved links

manugildev:
This PR fixes 13 warnings caused by unresolved internal and external(wgt) links in `cargo doc`.

> Screenshot 2021-01-06 at 18 33 00


kvark:
Very nice, thank you!
Could we make it a required CI check, perhaps?
bors r+

PR #707, opened 2021-01-10
Fix compilation on aarch64-apple-darwin

andykilroy:
wgpu-rs couldn't compile on the Apple M1 (arm64) macs due to a compilation
problem in winit.  winit 0.24.0 has the fix so upgrade the dependency.

Link to the fix's pull request in winit:
   https://github.com/rust-windowing/winit/pull/1752


kvark (approved):
Thank you!
note: it's not wgpu-rs that failed to compile, it's the examples
bors r+

PR #710, opened 2021-01-11
Implement read_buffer

### LINKED ISSUE #694: Enable mapping buffers via Encoder
Currently if you want to download a GPU resource produced while rendering a scene, you have to do the following steps in order:

1. Record a buffer copy command onto an active CommandEncoder
2. Finish any other rendering commands and submit the CommandEncoder
3. Call BufferSlice::map_async
4. Poll until the future completes

Because steps (1) and (2) can happen in very different parts of the code, it would be nice if I could run them in (1), (3), (2) order so that the buffer copy and asynchronous buffer map request could happen together. Unfortunately, this reordering doesn't work today: the queue submission will panic because it notices that the buffer is already mapped.

What I propose is to add an additional way of mapping buffers via a method on CommandEncoder. Buffer mapping is already asynchronous, but this method could delay the map operation all the way until the prior command recorded to the CommandEncoder had completed (which would be the copy... exactly what we want). Does this make sense / is it feasible to provide?

fintelia:
I'm not sure about the exact interface that makes sense, but this PR sketches out how read_buffer (and eventually also read_texture) could be implemented. 

Resolves #694

### DIFF
--- a/src/util/mod.rs
+++ b/src/util/mod.rs
@@ -54,3 +54,11 @@ pub fn make_spirv<'a>(data: &'a [u8]) -> super::ShaderSource<'a> {
     );
     super::ShaderSource::SpirV(words)
 }
+
+struct ReadBuffer(super::BufferMappedRange);
+impl std::ops::Deref for ReadBuffer{
+  type Target = [u8];
+  fn deref(&self) -> &[u8] {
+    super::BufferMappedRangeSlice::slice(&self.0)
+  }
+}
--- a/src/util/mod.rs
+++ b/src/util/mod.rs
@@ -6,6 +6,7 @@ mod encoder;
 
 use std::{
     borrow::Cow,
+    future::Future,
     mem::{align_of, size_of},
     ptr::copy_nonoverlapping,
 };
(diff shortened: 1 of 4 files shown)


src/util/mod.rs
```diff
@@ -54,3 +59,25 @@ pub fn make_spirv<'a>(data: &'a [u8]) -> super::ShaderSource<'a> {
     );
     super::ShaderSource::SpirV(words)
 }
+
+/// Asynchronously read the contents of buffer.
+pub fn read_buffer(device: &super::Device, queue: &super::Queue, buffer: &super::Buffer, offset: super::BufferAddress, size: super::BufferAddress) -> impl Future<Output=Result<Vec<u8>, super::BufferAsyncError>> + Send {
```
kvark, line 64: should accept a `BufferSlice`, makes things simpler

Cargo.toml
```diff
@@ -49,6 +49,7 @@ smallvec = "1"
 tracing = { version = "0.1", default-features = false, features = ["std"] }
 typed-arena = "2.0.1"
 serde = { version = "1", features = ["derive"], optional = true }
+futures = "0.3.9"
```
kvark, line 52: would be quite unfortunate to have this dependency just for this tiny little bit. Can we work around it?

kvark (commented):
Looks usable! Got a few notes.

src/util/mod.rs
fintelia: Is there a way to go from a BufferSlice to the size that's simpler than this?
```
slice.size.map(Into::into).unwrap_or(slice.buffer.map_context.lock().total_size)
```

kvark: We could place `total_size` on the buffer itself instead of the `map_context` to simplify this.
Also this code needs to take into account the offset:
```rust
match slice.size {
  Some(size) => size.into(),
  None => slice.buffer.map_context.lock().total_size - slice.offset,
}
```

src/util/mod.rs
```diff
@@ -54,3 +55,31 @@ pub fn make_spirv<'a>(data: &'a [u8]) -> super::ShaderSource<'a> {
...
+
+    let mut encoder = device.create_command_encoder(&super::CommandEncoderDescriptor { label: None });
+    encoder.copy_buffer_to_buffer(buffer.buffer, buffer.offset, &download, 0, size);
+    let command_buffer: super::CommandBuffer = encoder.finish();
+    queue.submit(Some(command_buffer));
+
+    let fut = download.slice(..).map_async(super::MapMode::Read);
+    async move {
+        fut.await?;
+        let data = download.slice(..).get_mapped_range().to_vec();
```
kvark, line 81: Would it make sense to return something like:
```rust
struct ReadBuffer {
  buffer: crate::Buffer,
}
impl Deref for ReadBuffer{
  type Item = [u8];
  fn deref(&self) -> &[u8] {
    self.buffer.slice(..).get_mapped_range()
  }
}
```
This wouldn't make it any less usable, but would avoid an unnecessary heap allocation and the data copy for many users of this new API.

fintelia: That might work. You are thinking that unmapping the buffer would just be handled when ReadBuffer was dropped?

kvark: yep

kvark: technically, `wgpu` should allow dropping a mapped buffer and unmap it internally, too

kvark:
Also, if we go with `ReadBuffer`, it would be great to make the method you are adding to basically be a constructor for it, i.e. user would call `ReadBuffer::new()`, they'd only need to `use wgpu::util::ReadBuffer`, etc.

src/util/mod.rs
fintelia: I tried some variations on your version, but I couldn't find any that have workable lifetimes. The problem was that the return value from Buffer::slice() is valid only for the current scope, but deref() needs to return a slice that references it.

kvark: I played a bit with it, and indeed you are right. A slight change is needed in order to expose this.
See https://github.com/kvark/wgpu-rs/commit/f5ea067cdb3629454cd0c4dcd17103d6bfef7dfa

fintelia: What are you thinking on this? Do you want to pull that change into wgpu-rs and then I can rebase this PR on it?

kvark: If you can just cherry-pick this commit into your PR, that would be fine.

kvark (commented):
I think it looks good now, thank you for addressing all of the concerns!
My last question is - was it intentional to hide `struct ReadBuffer`? Would there be any downside from having it as `pub` and implementing `ReadBuffer::new()` instead of `read_buffer`?

fintelia:
Hiding that struct wasn't for any particular reason. To leave open a future addition of reading back textures, does this seem reasonable to you:
```rust
pub struct DownloadBuffer(..);
impl DownloadBuffer{
    pub fn from_buffer() -> Self { ... }
    pub fn from_texture() -> Self { ... }
}

kvark:
That sounds great!

kvark (approved):
Great, thank you!
bors r+

PR #720, opened 2021-01-20
Remove typed-arena dependency

kvark:
Also bumps the wgpu dependency to include https://github.com/gfx-rs/wgpu/pull/1160

### DIFF
--- a/src/backend/direct.rs
+++ b/src/backend/direct.rs
@@ -20,7 +20,6 @@ use std::{
     slice,
     sync::Arc,
 };
-use typed_arena::Arena;
 
 const LABEL: &str = "label";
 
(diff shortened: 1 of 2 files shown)


src/backend/direct.rs
```diff
@@ -867,7 +868,23 @@ impl crate::Context for Context {
         wgc::span!(_guard, TRACE, "Device::create_bind_group wrapper");
         use wgc::binding_model as bm;
 
-        let texture_view_arena: Arena<wgc::id::TextureViewId> = Arena::new();
+        let mut arrayed_texture_views = Vec::new();
```
cwfitzgerald, line 871: It kinda pains me to allocate this without a `with_capacity` on the known max size, but it's good enough for now.

cwfitzgerald (approved):

src/backend/direct.rs
kvark: Agreed! There are things to try, but they involve `unsafe` one way or the other.... Let's leave for a follow-up

kvark:
bors r=cwfitzgerald

PR #727, opened 2021-01-23
RFC: Continuously run poll from a background thread on native backend

grovesNL:
One of the parts that's currently a bit confusing about async buffer mapping is how the device has to be manually polled after futures are created but before being awaited.

The device polling also doesn't exist in the upstream web API, so it would be nice to at least make it optional somehow (possibly opt-in if finer-grained control is needed).

In this PR we poll all devices continuously in a background thread. Futures will resolve automatically whenever the buffer mapping is ready. For example, we don't to use `force_wait` in order to read data back in the compute example or spin because the background thread is doing this for us already.

I'm interested in feedback about this approach and whether this is a good idea. I see a few pros and cons but hopefully some of the cons can be addressed later without API changes.

Pros:
- Slightly better alignment with web backend, since [this is similar to the approach gecko is using too](https://github.com/mozilla/gecko-dev/blob/6638e1d/dom/webgpu/ipc/WebGPUParent.cpp#L168)
- It's not necessary to manually call `device.poll` anymore because it will happen in the background
- We could probably remove `device.poll` entirely unless we want to allow for finer-grained control
- `Maintain::Wait` isn't portable to the web (e.g. it currently no-ops there) so this fixes a minor inconsistency which might not match expectations when targeting the web

Cons:
- A background thread is spinning and burning CPU cycles even though it might not be necessary (e.g. if no buffer mapping ever happens)
- Multiple threads are required for now when running on native, so we can't run on a single thread even for debugging
- A default interval of 100ms might cause longer waits for buffer mapping than using `wgpu::Maintain::Wait` (i.e. `force_wait` in wgpu)

Some future possibilities:
- We could open up more control over the thread (thread affinity, sleep interval/yielding continuously without waiting) or integrate with other threadpools if they're being used
- We could spin up/down threads if no futures/buffer mappings are currently pending
- We could probably make the spinloop a bit more smart by waiting different intervals depending on how much work is pending
- Maybe we could still make it work on a single thread for a (slow) debug by running with `force_wait`/spinning immediately whenever a future is created (i.e. basically blocking on buffer mapping whenever it happens)
- Maybe we'd be able to eventually eliminate most or all of the spinning internally without later API changes, and maybe even avoid creation of a thread in some cases (e.g. possibly epoll on Linux+Vulkan could be used, or timeline semaphores on Vulkan and similar concepts in other backends could be used)

We could also expose some way to opt-in or opt-out of the background thread/automatic polling to give users control over how polling should happen. This would allow people to have a choice about how polling should happen for the native backend (their own thread, blocking, etc.). For example, if we wanted to expose the ability to opt-out, we could have a function on the `Adapter` to disable it. In contrast, if we wanted to expose the ability to opt-in, we wouldn't launch a thread by default and instead expose a helper function that's able to create one (we could use the helper function in our examples).

### DIFF
--- a/examples/capture/main.rs
+++ b/examples/capture/main.rs
@@ -5,7 +5,7 @@ use std::env;
 use std::fs::File;
 use std::io::Write;
 use std::mem::size_of;
-use wgpu::{Buffer, Device};
+use wgpu::Buffer;
 
 async fn run(png_output_path: &str) {
     let args: Vec<_> = env::args().collect();
@@ -20,14 +20,14 @@ async fn run(png_output_path: &str) {
             return;
         }
     };
-    let (device, buffer, buffer_dimensions) = create_red_image_with_dimensions(width, height).await;
-    create_png(png_output_path, device, buffer, &buffer_dimensions).await;
+    let (buffer, buffer_dimensions) = create_red_image_with_dimensions(width, height).await;
+    create_png(png_output_path, buffer, &buffer_dimensions).await;
 }
 
 async fn create_red_image_with_dimensions(
     width: usize,
     height: usize,
-) -> (Device, Buffer, BufferDimensions) {
+) -> (Buffer, BufferDimensions) {
     let adapter = wgpu::Instance::new(wgpu::BackendBit::PRIMARY)
         .request_adapter(&wgpu::RequestAdapterOptions::default())
         .await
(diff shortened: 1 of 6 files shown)


kvark:
Thank you for making this!

One thing that's puzzling me a bit: there is always some kind of event loop going on that can be associated with polling. I.e. when targeting the web, there is an event loop. When rendering to `winit`, there is an event loop. The only case where there is no event loop are the compute and capture example son native, and that's precisely where we have the issue.

So I'm wondering if the way to go would be to heavily rely on those event loops:
  - convert the capture and compute examples to have some event loop, maybe `winit::event_loop::EventLoop`?
  - find a way to hook up our polling into the event loop. I.e. maybe `Instance::new()` should get a reference to the event loop, and plug into that?

From this perspective, your PR looks like just adding another domain-specific event loop/runtime to the native targets, on top of the existing ones.

kvark:
Project moved to `wgpu`

PR #744, opened 2021-02-05
Demonstrate obj loading in the skybox example

kvark:
I wanted to see which example is best suited for `typed-arena` demo... and here I am, implementing a completely different unrelated feature. But I think it's useful, still ;)
Model loading is something everybody needs, and for us depending on `obj` in dev-dependencies cost nothing, especially since I'm basically maintaining it.

screenshot


kvark:
@grovesNL should we force the `obj` to be considered binary for git?

PR #755, opened 2021-02-10
Add Instance::poll_all

lachlansneff:
`Instance::poll_all` polls all devices. This will be useful for integrating into the winit event loop from a third-party crate.﻿


kvark (approved):
Thank you!
bors r+

PR #759, opened 2021-02-12
Correct FragmentState targets comment

Systemcluster:
Previously the comment was copied verbatim from the `VertexState` `buffers` field.


grovesNL (approved):
Great, thank you!

PR #764, opened 2021-02-14
Change the texture sample type for the cube example from Float to Uint

igowen:
also move the mandelbrot->color mapping functionality from the texel generation code to the fragment shader.

this demonstrates the use of a non-float texture/sampler, which is not covered by any current example.

### DIFF
--- a/src/util/device.rs
+++ b/src/util/device.rs
@@ -25,7 +25,7 @@ pub trait DeviceExt {
     /// each layer and its mips to be tightly packed.
     ///
     /// Example:
-    /// Layer0Mip0 Layer0Mip1 Layer0Mip2 ... Layer1Mip0 Layer1Mip1 Layer1Mip2 ...  
+    /// Layer0Mip0 Layer0Mip1 Layer0Mip2 ... Layer1Mip0 Layer1Mip1 Layer1Mip2 ...
     fn create_texture_with_data(
         &self,
         queue: &crate::Queue,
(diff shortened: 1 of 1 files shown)


kvark (approved):
Wonderful, thank you!
bors r+

kvark:
actually, wait a bit
bors r-

kvark (changes requested):
We aren't validating this yet (soon!) but WGSL doesn't actually allow `textureSample` from integer textures. We need to change this to `textureLoad` now.

igowen:
hmm, in that case, we (a) don't need a sampler at all, and (b) need to convert texture coordinates to pixel coordinates ourselves, correct?

additionally, doing this results in the following error:

```
[0.117727 ERROR]()(no module): unimplemented ImageLoad { image: Handle(1), coordinate: Handle(10), array_index: None, index: Some(Handle(11)) }
[0.118114 ERROR]()(no module): unimplemented ImageLoad { image: Handle(1), coordinate: Handle(10), array_index: None, index: Some(Handle(11)) }
[0.118177 ERROR]()(wgpu_core::device): Shader error: Shader module failed to compile: unimplemented expression
wgpu error: Validation Error
```

kvark:
> don't need a sampler at all

correct!

> need to convert texture coordinates to pixel coordinates ourselves, correct

Yes, that's true. We may need to pass the size in uniforms, or query it on the texture itself.

>  unimplemented ImageLoad 

Let's disable `EXPERIMENTAL_TRANSLATION` flag for now. The image loads are fully supported in latest Naga, but it will arrive here when `gfx-11` tag is propagated (soon). So we'll re-enable the flag later.

igowen:
summarizing offline discussion -

- this is not actually caused by `EXPERIMENTAL_TRANSLATION`
- once [this change in Naga](https://github.com/gfx-rs/naga/pull/465) is released `textureLoad()` should work

so i'll sit on this PR for now and update it once Naga supports the necessary functionality.

PR #765, opened 2021-02-16
Update naga to gfx-11

kvark:
Gets us https://github.com/gfx-rs/wgpu/pull/1220

Note that "cross" feature is not optional here. We could lift it up and add to `default = []` but that would make the Web backend to always unconditionally depend on `wgpu-core` (since enabling a feature on it automatically enables it). So we ideally need a way for Cargo to allow platform-specific default features...

About the shadow example - the experimental translation works on the main pipeline, but fails on the baking pipeline because of https://github.com/gfx-rs/naga/issues/483. Fortunately, it falls back to SPIRV-Cross gracefully here.

The new validation detected a flaw in our shader (comparison sampler mismatch), yay!


cwfitzgerald (approved):
LGTM
bors r+

PR #771, opened 2021-02-22
Fix create_texture_with_data for 3d textures

expenses:
Currently, `create_texture_with_data` doesn't take depth into account for `TextureDimension::D3` textures.


kvark (approved):
Thank you!

PR #773, opened 2021-02-24
Panic on empty buffer slices

### LINKED ISSUE #735: Empty buffer slice's mapped range isn't empty
Code:

```rust
fn main() {
    let (device, _queue) = futures::executor::block_on(async {
        let instance = wgpu::Instance::new(wgpu::BackendBit::PRIMARY);
        instance.request_adapter(&wgpu::RequestAdapterOptions::default())
            .await
            .unwrap()
            .request_device(&wgpu::DeviceDescriptor::default(), None)
            .await
            .unwrap()
    });

    let buffer = device.create_buffer(&wgpu::BufferDescriptor {
        label: None,
        size: 4,
        usage: wgpu::BufferUsage::empty(),
        mapped_at_creation: true,
    });

    assert_eq!(
        buffer.slice(..0).get_mapped_range().len(),
        0,
    );
}
```

Expected outcome: program exits successfully, or panics in `slice`/`get_mapped_range` if empty buffer slices are not allowed
Actual outcome:

```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `4`,
 right: `0`', src/main.rs:19:5
```

Creating an empty slice at any starting point always maps the whole remainder of the buffer (so if the starting point is the end of the buffer the the mapped range is empty as expected).

Looks like the problem is here:

https://github.com/gfx-rs/wgpu-rs/blob/f882abcf4b3d1e9bfbfc95b61874344661956479/src/lib.rs#L1682-L1688

If `offset` is equal to `bound`, `BufferSize::new` (which is `NonZeroU64::new`) returns `None`, producing the same result as if the range was unbounded.

Wumpf:
Fixes #735 
Can't really pass through zero sizes, so panic it is.

I'm really just here to update to latest wgpu version to get my recent fix there in a more convenient manner, but doing a PR with just that felt a bit boring ;-). Now I'm of course guilty of conflating things 🤔

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1682,12 +1682,40 @@ fn range_to_offset_size<S: RangeBounds<BufferAddress>>(
         Bound::Unbounded => 0,
     };
     let size = match bounds.end_bound() {
-        Bound::Included(&bound) => BufferSize::new(bound + 1 - offset),
-        Bound::Excluded(&bound) => BufferSize::new(bound - offset),
+        Bound::Included(&bound) => Some(bound + 1 - offset),
+        Bound::Excluded(&bound) => Some(bound - offset),
         Bound::Unbounded => None,
-    };
+    }
+    .map(|size| BufferSize::new(size).expect("Buffer slices can not be empty"));
 
     (offset, size)
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::BufferSize;
+
+    #[test]
+    fn range_to_offset_size_works() {
+        assert_eq!(crate::range_to_offset_size(0..2), (0, BufferSize::new(2)));
+        assert_eq!(crate::range_to_offset_size(2..5), (2, BufferSize::new(3)));
+        assert_eq!(crate::range_to_offset_size(..), (0, None));
+        assert_eq!(crate::range_to_offset_size(21..), (21, None));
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1682,14 +1682,48 @@ fn range_to_offset_size<S: RangeBounds<BufferAddress>>(
...
     };
     let size = match bounds.end_bound() {
-        Bound::Included(&bound) => BufferSize::new(bound + 1 - offset),
-        Bound::Excluded(&bound) => BufferSize::new(bound - offset),
+        Bound::Included(&bound) => Some(bound + 1 - offset),
+        Bound::Excluded(&bound) => Some(bound - offset),
         Bound::Unbounded => None,
-    };
+    }
+    .map(|size| {
```
kvark, line 1689: why not just `BufferSize::new(size).expect("Buffer slices can not be empty")`?

kvark (commented):
Thank you! There is one thing I think we can improve

src/lib.rs
Wumpf: Because I still don't have all the Option utility functions in my head 😊 
Sooo much better :)

kvark (approved):
Thank you!
bors r+

PR #778, opened 2021-03-01
Return None from get_swap_chain_preferred_format if the adapter does not support the surface

### LINKED ISSUE #777: Add method to check compatibility between adapter and surface
Hi! The `enumerate_adapters` method is very useful for desktop users which can have multiple GPUs since it allows choosing the preferred GPU by name instead of relying on GPU type (integrated/discrete) which I feel is more intuitive.

However, there is one minor problem with choosing the GPU by name: it can potentially be incompatible with the given surface. If I understand correctly, the way to check if the compatibility is calling `Adapter::get_swap_chain_preferred_format` and catching panics.

Maybe a separate method like `Adapter::is_compatible_surface` or `Adapter::try_get_swap_chain_preferred_format` (returning `Result`) could be added?

OlegOAndreev:
Fixes #777 

Almost every other method in wgpu panics on failure, so I decided to make the return value an `Option<TextureFormat>`, not a `Result<Option<TextureFormat>>`

### DIFF
--- a/examples/framework.rs
+++ b/examples/framework.rs
@@ -202,7 +202,7 @@ fn start<E: Example>(
     let spawner = Spawner::new();
     let mut sc_desc = wgpu::SwapChainDescriptor {
         usage: wgpu::TextureUsage::RENDER_ATTACHMENT,
-        format: adapter.get_swap_chain_preferred_format(&surface),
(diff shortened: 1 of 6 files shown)


kvark (approved):
Thank you!

PR #781, opened 2021-03-04
Mark BufferMappedRange Send/Sync on native

cwfitzgerald:
This fixes an issue reported on the matrix where this was causing things to unnecessarily be !Send. There's no reason I can see that this couldn't be Send/Sync as we only use it like one would a slice. We might also consider backporting this to 0.7.


kvark (approved):
bors r+

PR #791, opened 2021-03-14
Added example for conservative rasterization

Wumpf:
Uses new conservative raster feature https://github.com/gfx-rs/wgpu/pull/1263 to implement a comprehensive demo


examples/README.md
```diff
@@ -10,40 +10,41 @@ Notably, `capture` example shows rendering without a surface/window. It reads ba
...
-| - queries              |        |        | :star: |           |        |        |                |        |
-| *integrations*         |        |        |        |           |        |        |                |        |
-| - staging belt         |        |        |        |           |        |        |                |        |
-| - typed arena          |        |        |        |           |        |        |                |        |
-| - obj loading          |        |        |        |           |        | :star: |                |        |
+| Feature                      | boids  | cube   | mipmap | msaa-line | shadow | skybox | texture-arrays | water  | conservative-raster |
+| ---------------------------- | ------ | ------ | ------ | --------- | ------ | ------ | -------------- | ------ | ------------------- |
+| WGSL shaders                 | :star: | :star: | :star: | :star:    | :star: | :star: |                |        | :star:              |
+| vertex attributes            | :star: | :star: |        | :star:    | :star: | :star: | :star:         | :star: | :star:              |
```
kvark, line 16: does the example need vertex attributes? I think we can get away without them

examples/README.md
```diff
@@ -10,40 +10,41 @@ Notably, `capture` example shows rendering without a surface/window. It reads ba
...
+| instancing                   | :star: |        |        |           |        |        |                |        |                     |
+| lines and points             |        |        |        | :star:    |        |        |                |        | :star:              |
```
kvark, line 18: conservative doesn't apply to lines and points, so wonder why using them in the example?

examples/conservative-raster/README.md
```diff
@@ -0,0 +1,19 @@
+# conservative-raster
+
+This example shows how to render with conservative rasterization (native extension with limited support).
+
+When enabled, any pixel touched by a triangle primitive is rastered.
```
kvark, line 5: "rasterized"

examples/conservative-raster/main.rs
```diff
@@ -0,0 +1,333 @@
...
+                }),
+                bind_group_layout,
+            )
+        };
+
+        let (low_res_target, bind_group_upscale) =
+            Self::create_low_res_target(sc_desc, device, &bind_group_layout_upscale);
+
+        println!("------------------------------------------------------------------------");
+        println!("Press Space to toggle between regular and conservative rasterization");
```
kvark, line 226: Would it be easier to just draw both conservative, non-conservative (with different color), and lines on top, to show the difference? I think non-interactive examples are easier to test.

examples/conservative-raster/triangle_and_lines.wgsl
```diff
@@ -0,0 +1,17 @@
+[[stage(vertex)]]
+fn vs_main([[builtin(vertex_index)]] vertex_index: u32) -> [[builtin(position)]] vec4<f32> {
+    var i: i32 = i32(vertex_index % 3u);
```
kvark, line 3: let's use `const` here in other places instead of `var`

examples/conservative-raster/triangle_and_lines.wgsl
```diff
@@ -0,0 +1,17 @@
+[[stage(vertex)]]
+fn vs_main([[builtin(vertex_index)]] vertex_index: u32) -> [[builtin(position)]] vec4<f32> {
```
kvark, line 2: ok, so we aren't really using vertex attributes here, good! Let's remove the star from the table.

examples/conservative-raster/upscale.wgsl
```diff
@@ -0,0 +1,24 @@
+struct VertexOutput {
+    [[builtin(position)]] position: vec4<f32>;
+    [[location(0)]] tex_coords: vec2<f32>;
+};
+
+[[stage(vertex)]]
+fn vs_main([[builtin(vertex_index)]] vertex_index: u32) -> VertexOutput {
+    var x: f32 = f32(i32(vertex_index & 1) << 2) - 1.0;
```
kvark, line 8: similarly, let's use `const`

kvark (commented):
Looking great! Have a few suggestions.

examples/README.md
Wumpf: I'm using it to draw "the real triangle" over the pixelated one
https://github.com/Wumpf/wgpu-rs/blob/conservative-raster/examples/conservative-raster/screenshot.png
I'd really like to keep that because it's a nice visualization. But with a bit of effort I can make it optional

examples/conservative-raster/triangle_and_lines.wgsl
Wumpf: ah. Haven't done much of this fancy new wglsl :) fixing!

examples/conservative-raster/main.rs
Wumpf: image
tried it here. Don't like it as much but I get your point - much easier to go through them and maybe even implement an image comparison test later.

kvark: looks great!

Wumpf: okay. I hate the colors, but it will do :D

kvark: we can change the colors :)

kvark (approved):
Thank you!

PR #795, opened 2021-03-16
Fix excessive buffer padding

HalfVoxel:
Previously buffers sizes which were already aligned would be unnecessarily padded with `COPY_BUFFER_ALIGNMENT` (4) bytes.

### DIFF
--- a/src/util/device.rs
+++ b/src/util/device.rs
@@ -37,8 +37,10 @@ pub trait DeviceExt {
 impl DeviceExt for crate::Device {
     fn create_buffer_init(&self, descriptor: &BufferInitDescriptor<'_>) -> crate::Buffer {
         let unpadded_size = descriptor.contents.len() as crate::BufferAddress;
-        let padding = crate::COPY_BUFFER_ALIGNMENT - unpadded_size % crate::COPY_BUFFER_ALIGNMENT;
-        let padded_size = padding + unpadded_size;
+        // Valid vulkan usage is
+        // 1. buffer size must be a multiple of COPY_BUFFER_ALIGNMENT.
+        // 2. buffer size must be greater than 0.
+        let padded_size = ((unpadded_size + crate::COPY_BUFFER_ALIGNMENT - 1) / crate::COPY_BUFFER_ALIGNMENT).max(1) * crate::COPY_BUFFER_ALIGNMENT;
 
         let wgt_descriptor = crate::BufferDescriptor {
             label: descriptor.label,
--- a/src/util/device.rs
+++ b/src/util/device.rs
@@ -1,5 +1,7 @@
 use std::convert::TryFrom;
 
+use crate::COPY_BUFFER_ALIGNMENT;
+
 /// Describes a [Buffer](crate::Buffer) when allocating.
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct BufferInitDescriptor<'a> {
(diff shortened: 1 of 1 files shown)


src/util/device.rs
```diff
@@ -37,8 +39,10 @@ pub trait DeviceExt {
 impl DeviceExt for crate::Device {
     fn create_buffer_init(&self, descriptor: &BufferInitDescriptor<'_>) -> crate::Buffer {
         let unpadded_size = descriptor.contents.len() as crate::BufferAddress;
-        let padding = crate::COPY_BUFFER_ALIGNMENT - unpadded_size % crate::COPY_BUFFER_ALIGNMENT;
-        let padded_size = padding + unpadded_size;
+        // Valid vulkan usage is
+        // 1. buffer size must be a multiple of COPY_BUFFER_ALIGNMENT.
+        // 2. buffer size must be greater than 0.
+        let padded_size = ((unpadded_size + COPY_BUFFER_ALIGNMENT - 1) / COPY_BUFFER_ALIGNMENT).max(1) * COPY_BUFFER_ALIGNMENT;
```
cwfitzgerald: We can be a smidge cleverer here:
```suggestion
        let align_mask = COPY_BUFFER_ALIGNMENT - 1;
        let padded_size = ((unpadded_size + align_mask) & align_max).max(1);
```
But otherwise looks good!

cwfitzgerald (changes requested):

cwfitzgerald:
With those changes, please squash and we'll be good to go.

src/util/device.rs
HalfVoxel: Hmm, yes if we assume the alignment is a power of two (which I guess it will always be) then that will work. Thanks!
Though your `max` call should be `max(COPY_BUFFER_ALIGNMENT)` in this case. I'll make that change separately.

HalfVoxel: Actually I checked the assembly, and llvm generates the exact same output for both cases. Should I still change this? I suppose it just comes down to what is more readable.

cwfitzgerald: Ah yes, good catch :)

cwfitzgerald: IMO it's clearer this way as the whole divide then multiply thing always takes me a minute to properly parse what's going on.

HalfVoxel:
The bot says it has found warnings and errors... but it doesn't seem to show them? Is it configured correctly?

src/util/device.rs
```diff
@@ -1,5 +1,7 @@
 use std::convert::TryFrom;
 
+use crate::COPY_BUFFER_ALIGNMENT;
```
kvark, line 3: let's follow the existing style please, where `crate::COPY_BUFFER_ALIGNMENT` is fully qualified where it's useed

kvark (commented):
Thank you for the PR!
One nit plus squashing, and we'll be good to go.
Don't mind the monocodus, please. If CI is green, we are happy.

HalfVoxel:
@kvark, done. Squashing I believe can be done by github directly when merging.
image

Btw, locally I'm getting a ton of clippy errors for other parts of the code, am I using the wrong clippy config?

kvark:
> Squashing I believe can be done by github directly when merging.

Wee use a merge bot which doesn't know how to squash.
I can squash manually though if I see no other PRs are in flight

kvark:
> Btw, locally I'm getting a ton of clippy errors for other parts of the code, am I using the wrong clippy config?

We aligned naga and wgpu to respect clippy warnings/errors, but not gfx-rs or wgpu-rs yet. This would be nice to get!

PR #796, opened 2021-03-16
Update naga to gfx-15

kvark:
Depends on https://github.com/gfx-rs/wgpu/pull/1267
Puts an end to our DX12 leakage


kvark:
bors r-
oops, water example fails

kvark:
> thread 'main' panicked at 'index out of bounds: the len is 2 but the index is 2', /Users/dmalyshau/.cargo/git/checkouts/naga-dbb2b19faed49210/196523d/src/proc/analyzer.rs:633:33

PR #797, opened 2021-03-16
Move from `tracing` to `log` + `profiling`

cwfitzgerald:
Continuation of https://github.com/gfx-rs/wgpu/pull/1268 into wgpu-rs

This also includes the naga gfx-15 train.

### DIFF
--- a/src/backend/direct.rs
+++ b/src/backend/direct.rs
@@ -397,7 +397,7 @@ mod pass_impl {
             &mut self,
             render_bundles: I,
         ) {
-            wgc::span!(_guard, TRACE, "RenderPass::execute_bundles wrapper");
+            profiling::scope!("RenderPass::execute_bundles wrapper");
             let temp_render_bundles = render_bundles.cloned().collect::<SmallVec<[_; 4]>>();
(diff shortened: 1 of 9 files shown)


src/backend/direct.rs
```diff
@@ -1313,7 +1313,7 @@ impl crate::Context for Context {
             sub_range.start,
             wgt::BufferSize::new(size)
         )) {
-            Ok(ptr) => BufferMappedRange {
+            Ok((ptr, size)) => BufferMappedRange {
```
cwfitzgerald, line 1316: Is this right? I took a stab at fixing this compile error.

Systemcluster: @kvark's PR included the same change: https://github.com/gfx-rs/wgpu-rs/pull/796/files#diff-8d6e6a150805924081efd1ebe6a37810a66a9a1f3b712d0523429f1d65610bceR1316

kvark (approved):
Note: this will break the water example temporarily.
The fixes in Naga gfx-16 branch are following up shortly.
bors r+

PR #798, opened 2021-03-17
Fix all the clippy lints

HalfVoxel:
Clippy is now happy

### DIFF
--- a/examples/skybox/main.rs
+++ b/examples/skybox/main.rs
@@ -363,6 +363,7 @@ impl framework::Example for Skybox {
         }
     }
 
+    #[allow(clippy::single_match)]
     fn update(&mut self, event: winit::event::WindowEvent) {
         match event {
             winit::event::WindowEvent::CursorMoved { position, .. } => {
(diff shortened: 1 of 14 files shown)


examples/skybox/main.rs
```diff
@@ -364,14 +364,11 @@ impl framework::Example for Skybox {
...
     fn update(&mut self, event: winit::event::WindowEvent) {
-        match event {
-            winit::event::WindowEvent::CursorMoved { position, .. } => {
-                let norm_x = position.x as f32 / self.camera.screen_size.0 as f32 - 0.5;
-                let norm_y = position.y as f32 / self.camera.screen_size.1 as f32 - 0.5;
-                self.camera.angle_y = norm_x * 5.0;
-                self.camera.angle_xz = norm_y;
-            }
-            _ => {}
+        if let winit::event::WindowEvent::CursorMoved { position, .. } = event {
```
kvark, line 367: hmm, that's not desired. if/let is not scaleable: once we need to add a new variant, we'll have to go back to a match

kvark (commented):
Could we also have a CI job checking for this? See https://github.com/gfx-rs/wgpu/blob/bb9a5a85d2bc1ee555739c646976f992a48127b5/.github/workflows/ci.yml#L117

kvark (approved):
Thank you!
bors r+

PR #809, opened 2021-03-23
Add overflow detection to hello-compute

danwilhelm:
Description
The `hello-compute` example sometimes gives incorrect results if an intermediate Collatz value overflows. 

Proposed Solution
We can detect overflows before they occur, since the `3n+1` step is the only possible overflow. (WGSL only supports u32.)

Example
`cargo run --example hello-compute 77031 837799 8400511 63728127` 
> Incorrect Current Output: `[350, 524, 312, 346]`
> Correct Output: `[350, 524, 685, 949]`
> Proposed Solution: `[350, 524, OVERFLOW, OVERFLOW]`

Notes
- The `hello-compute` example appears to be [borrowed from naga](https://github.com/gfx-rs/naga/blob/master/tests/in/collatz.wgsl), so this would cause the implementations to differ.
- Hexadecimal literals are not yet implemented in naga for WGSL shaders, so the shader literals are not as pretty as they could be.
- Revised terminology `times` -> `steps` to be consistent with [Wikipedia](https://en.wikipedia.org/wiki/Collatz_conjecture).

### DIFF
--- a/examples/hello-compute/shader.wgsl
+++ b/examples/hello-compute/shader.wgsl
@@ -23,6 +23,11 @@ fn collatz_iterations(n_base: u32) -> u32{
             n = n / 2u;
         }
         else {
+            // Overflow? (i.e. 3*n + 1 > 0xffffffffu?)
+            if (n >= 1431655765u) {   // 0x55555555u
+                return 4294967295u;   // 0xffffffffu
+            }
+
             n = 3u * n + 1u;
         }
         i = i + 1u;
--- a/examples/hello-compute/main.rs
+++ b/examples/hello-compute/main.rs
@@ -1,6 +1,9 @@
 use std::{borrow::Cow, convert::TryInto, str::FromStr};
 use wgpu::util::DeviceExt;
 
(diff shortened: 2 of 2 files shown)


examples/hello-compute/shader.wgsl
```diff
@@ -23,6 +23,11 @@ fn collatz_iterations(n_base: u32) -> u32{
             n = n / 2u;
         }
         else {
+            // Overflow? (i.e. 3*n + 1 > 0xffffffffu?)
+            if (n >= 1431655765u) {   // 0x55555555u
+                return 4294967295u;   // 0xffffffffu
```
kvark, line 28: we could write this simpler, i.e. `~0u` or `u32(-1)`

kvark (approved):
Thank you!
bors r+

examples/hello-compute/shader.wgsl
danwilhelm: `~0u` would have been nice! I'll update this when hex is working to `0xffffffff` to emphasize it is the max value and to make it consistent with the Rust constant.

PR #810, opened 2021-03-23
Passthrough cross feature to wgpu-core

Uriopass:
Hopefully it can even be removed when `naga` is more mature. :-)


kvark:
I'll update the README to say that one has to disable the cross feature when building for WASM WebGPU

PR #813, opened 2021-03-25
Fix dropping for command encoders

kvark:
We weren't dropping them properly, and since the ID is just an obfuscated `Box<RealStuff>` that made the contents leak.


grovesNL (approved):
Thanks!
bors r+

PR #818, opened 2021-03-29
Update wasm-bindgen dependencies.

VincentFTS:
Because of https://github.com/rustwasm/wasm-bindgen/issues/2508

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,7 +89,7 @@ test = true
 #gfx-backend-metal = { path = "../gfx/src/backend/metal" }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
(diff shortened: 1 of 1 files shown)


kvark (approved):
thank you!
bors r+

kvark:
Oh snap, this is 0.7, I didn't realize.
Not sure if we can update these without breaking downstream depenencies

kvark:
@VincentFTS if this is going to break the downstream projects, let's back this out

kvark:
@VincentFTS I've just backed this out, until we figure out how to land this nicely

VincentFTS:
Do you prefer to add syn dependency to fix its version ?

grovesNL:
Yeah at least temporarily we might want to pin the old version of syn for 0.7 on wasm32 (`syn = "=1.0.65"`)

PR #819, opened 2021-03-29
Boost texels alpha channel in cube example.

VincentFTS:
Permits to have a good rendering for WebGL.


kvark (approved):
Thank you!
bors r+

kvark:
oh wow, what's happened to our wasm build? cc @grovesNL

Gordon-F:
We need to update wasm deps in master too https://github.com/gfx-rs/wgpu-rs/pull/818

VincentFTS:
> We need to update wasm deps in master too #818

Btw, is a specific a MR is needed for master or #818 is OK ?

PR #821, opened 2021-03-30
[0.7] Require syn 1.0.65 for web backend

grovesNL:
Alternative to #818

`syn` is already included by a transitive dependency that also has a fixed version here, so it doesn't need to be `optional`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -185,7 +185,7 @@ web-sys = { version = "=0.3.46", features = [
 ]}
 js-sys = "0.3.46"
 wasm-bindgen-futures = "0.4.19"
-syn = { version = "=1.0.65", optional = true }
+syn = "=1.0.65"
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -185,6 +185,7 @@ web-sys = { version = "=0.3.46", features = [
 ]}
 js-sys = "0.3.46"
 wasm-bindgen-futures = "0.4.19"
+syn = { version = "=1.0.65", optional = true }
```
CryZe, line 188: I don't believe you can have target specific dependencies like that, and it being optional probably doesn't do anything either.

grovesNL: Yeah I guess the `optional` here isn't useful since we already know syn is included upstream.

We do use target specific dependencies with fixed versions already though and they seem to work fine

kvark (approved):
So we'll end up with multiple versions of `syn` in the lockfile? I suppose it's the best we can do here.
bors r+

VincentFTS:
Can we have a new release with this ?

PR #822, opened 2021-03-30
Update wasm-bindgen and web-sys to latest

grovesNL:
#821 for `master`

On `master` it should be ok to update wasm-bindgen and web-sys, so we don't need to fix syn's version


kvark (approved):
Thank you!
bors r+

PR #823, opened 2021-03-30
Fix crash-on-resize in skybox example

danwilhelm:
Problem: Skybox example currently crashes on resize. 
- Solution: Now, recreates depth texture on resize event.
---
Problem: Shadow example currently duplicates texture creation code on resize. 
- Solution: Now, shadow is refactored identically to skybox & water examples -- minimizes code duplication, makes examples more similar.


kvark (approved):
Very nice, thank you!
bors r+

PR #824, opened 2021-03-30
make all labels use `wgpu::Label`

LU15W1R7H:
This PR does a very minor change.

It replaces all occurences of `label: Option<'a str>` with `label: crate::Label`,
making the usage consistent all across the API.

This doesn't bring any benefit other than consistency.

I hope to have catched every occurance.


LU15W1R7H:
yeah 🥳

PR #826, opened 2021-03-31
Disable SPIRV-Cross by default

kvark:


cwfitzgerald (approved):
It doesn't compile, but r=me when it does! Less c++ here we goooo!

PR #834, opened 2021-04-01
frame capture

adamnemecek:


src/backend/web.rs
```diff
@@ -2021,6 +2021,14 @@ impl crate::Context for Context {
     fn queue_get_timestamp_period(&self, _queue: &Self::QueueId) -> f32 {
         1.0 //TODO
     }
+
+    fn start_capture(&self, device: &Self::DeviceId) {
+        unimplemented!()
```
kvark, line 2026: let's just not do anything there.
Now that I think of it, we shouldn't have `unimplemented!` for this in gfx-rs either. Doing a capture start/stop should be a no-op on backends that don't support it.

adamnemecek: I think that crashing is better, like the user is expecting something to happen, we should indicate that it didn't happen.

kvark: Crashing is unacceptable in wgpu in general

PR #847, opened 2021-04-07
Don't pin web-sys and wasm-bindgen versions

RichoDemus:
Pinning them forces all downstream dependencies to use exactly this version

The latest release of wgpu has pinned an old version of wasm-bindgen, forcing users to downgrade wasm-bindgen-cli in order to build

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -28,7 +28,7 @@ jobs:
         run: cargo build --release --target wasm32-unknown-unknown --examples
 
       - name: Install wasm-bindgen-cli
-        run: cargo install --force wasm-bindgen-cli --version 0.2.73
+        run: cargo install wasm-bindgen-cli
 
       - name: Generate JS bindings for the examples
         run: |
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -116,7 +116,7 @@ test = true
 #wasm-bindgen = { path = "../wasm-bindgen" }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-wasm-bindgen = "=0.2.73" # remember to change version in wiki as well
+wasm-bindgen = "0.2.73" # remember to change version in wiki as well
 web-sys = { version = "=0.3.50", features = [
     "Document",
     "Navigator",


grovesNL:
Hi! :wave: The WebGPU WebIDL is unstable, so the API often has breaking changes in web-sys between releases (i.e. not following semver). This also means we can't automatically use new versions of web-sys, because it would break this crate and any crates depending on it. Until it's stabilized we keep web-sys pinned and try to bump it regularly.

RichoDemus:
I see, does the same apply for wasm-bindgen or can I modify the PR to just remove the pin there?

grovesNL:
Yeah it's probably ok to unpin wasm-bindgen, but we'd need to change our [examples publish action](https://github.com/gfx-rs/wgpu-rs/blob/master/.github/workflows/publish.yml#L31) (e.g. possibly updating wasm-bindgen to latest before running the CLI) which needs the versions of wasm-bindgen and wasm-bindgen-cli to match exactly.

.github/workflows/publish.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         run: cargo build --release --target wasm32-unknown-unknown --examples
 
       - name: Install wasm-bindgen-cli
-        run: cargo install --force wasm-bindgen-cli --version 0.2.73
+        run: cargo install wasm-bindgen-cli
```
grovesNL, line 31: How do we know that `wasm-bindgen-cli` version matches the one used during `cargo build` above?

RichoDemus: Since wgpu-rs doesn't have a lock file, shouldn't both pick up the latest one? (apart from differences in major version)
Maybe there's a better way of doing this that I don't know

grovesNL: Yeah generally I think that's true. It's possible they might not match in some cases (e.g. if a dependency or transitive dependency requires a particular version or version range) although those cases are probably unlikely, and if/when that happens we can pin CI to a particular version or workaround it. So we can probably proceed with this

grovesNL (approved):
bors r+

PR #850, opened 2021-04-08
Add uniform and storage buffer arrays

csnewman:
**Connections**
https://github.com/gfx-rs/wgpu/pull/1315

**Description**
Adds uniform and storage buffer arrays

Usage:
```
wgpu::BindGroupEntry {
    binding: 0,
    resource: BindingResource::BufferArray(&[
        buffer1.as_entire_buffer_binding(),
        buffer2.as_entire_buffer_binding(),
    ]),
}
```

### DIFF
--- a/src/backend/direct.rs
+++ b/src/backend/direct.rs
@@ -1,6 +1,6 @@
 use crate::{
     backend::{error::ContextError, native_gpu_future},
-    AdapterInfo, BindGroupDescriptor, BindGroupLayoutDescriptor, BindingResource,
+    AdapterInfo, BindGroupDescriptor, BindGroupLayoutDescriptor, BindingResource, BufferBinding,
     CommandEncoderDescriptor, ComputePassDescriptor, ComputePipelineDescriptor,
     DownlevelProperties, Features, Label, Limits, LoadOp, MapMode, Operations,
     PipelineLayoutDescriptor, RenderBundleEncoderDescriptor, RenderPipelineDescriptor,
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -1834,7 +1845,12 @@ impl Drop for BufferViewMut<'_> {
 impl Buffer {
     /// Return the binding view of the entire buffer.
     pub fn as_entire_binding(&self) -> BindingResource {
-        BindingResource::Buffer {
+        BindingResource::Buffer(self.as_entire_buffer_binding())
+    }
+
+    /// Return the binding view of the entire buffer.
+    pub fn as_entire_buffer_binding(&self) -> BufferBinding {
```
csnewman, line 1852: Can be merged with `as_entire_binding` if we don't mind breaking changes.

src/backend/direct.rs
```diff
@@ -875,21 +875,47 @@ impl crate::Context for Context {
         }
         let mut remaining_arrayed_texture_views = &arrayed_texture_views[..];
 
+        let mut arrayed_buffer_bindings = Vec::new();
+        if device.features.contains(Features::BUFFER_BINDING_ARRAY) {
+            // gather all the buffers first
+            for entry in desc.entries.iter() {
+                match entry.resource {
```
kvark, line 882: nit: `if let` would work better here, if you don't want to match exhaustively

src/lib.rs
kvark: We don't mind, but also it seems useful to have this one now.
We can either keep the old one, or even do `impl From<BufferBinding> for BindingResource {..}`

kvark (approved):

src/lib.rs
cwfitzgerald: `From` is definitely more idiomatic, but is harder to discover in docs so we should call it out specifically. I have received a suggestion to have From<> for BindingResource for each type of binding.

kvark: I'm concerned about deriving `From<>` for references though. I believe this is a bad practice, since at the call site you'd see `xxx.into()` and assume it's converted by consumption, while it's not.

cwfitzgerald: Hmm, yeah it also won't auto-convert for you so if you have a Vec<BufferBinding> you need to call `xxx.as_ref().into()` which is kinda lame. I definitely feel like we should have a consistent abstraction here somehow, but maybe that's just making every resource have an `as_binding()`...

kvark: Yeah, that doesn't sound too bad.

PR #857, opened 2021-04-14
updated wgpu with more copyable types

adamnemecek:


kvark:
Let's include https://github.com/gfx-rs/wgpu/pull/1329 in here

PR #859, opened 2021-04-15
Fix typos - it's -> its

ashpil:
As I was reading the documentation, I found an instance where "it's" should actually be "its", as it is used as a possessive, not a contraction. I fixed that and `grep`ed for other instances, of which I found one and fixed as well.


cwfitzgerald:
English is hard! Thanks!

bors r+

PR #860, opened 2021-04-15
Drops the Sync requirement from UncapturedErrorHandler

adamnemecek:


kvark (approved):
Appears safe to me because we wrap it into `ErrorSinkRaw`, which is inside a `Mutex`. So all the access to the error handler is going to be synchronized externally.
bors r+

PR #861, opened 2021-04-15
Remove glslangvalidator requirements from water example

Gordon-F:
`glslangvalidator` is no longer needed because the example uses WGSL.


kvark (approved):
Thank you!
bors r+

PR #870, opened 2021-04-21
Remove dead code

dhardy:
Also, why bother setting the padding bytes to 0?

### DIFF
--- a/src/util/device.rs
+++ b/src/util/device.rs
@@ -52,10 +52,6 @@ impl DeviceExt for crate::Device {
             mapped_at_creation: true,
         };
 
-        let mut map_context = crate::MapContext::new(padded_size);
-
-        map_context.initial_range = 0..padded_size;
-
         let buffer = self.create_buffer(&wgt_descriptor);
         {
             let mut slice = buffer.slice(..).get_mapped_range_mut();


kvark (approved):
IIRC there is a way to observe the padding, so we zero it out just to be sure.
bors r+

grovesNL:
@kvark should we expect the padded range to be zeroed already though? If not, could we move this zeroing down to wgpu-core?

kvark:
@grovesNL in wgpu-core you can't create a mapped buffer that is not rounded to 4 bytes already. So it doesn't have a problem. Only wgpu-rs utils have this issue, since they try to be more flexible than wgpu-core itself.

grovesNL:
Alright yeah that makes sense to me. In that case though how could we end up with non-zero values in the extra padded range (i.e. the last  `unpadded_size % 4` bytes)? I thought zeroing it here would be redundant, because it should be zeroed already (in wgpu-core/Gecko)

kvark:
Indeed, you are right, wgpu-core recently got this functionality. We can remove that code now from utils.

dhardy:
Already merged or I would adjust this PR. I'll let you fold into another.

But good to know that the zeroing is redundant. :+1:

PR #872, opened 2021-04-21
Remove zero padding from `create_buffer_init`

grovesNL:
Follow-up to #870 (https://github.com/gfx-rs/wgpu-rs/pull/870#issuecomment-824333926)

### DIFF
--- a/src/util/device.rs
+++ b/src/util/device.rs
@@ -53,14 +53,10 @@ impl DeviceExt for crate::Device {
         };
 
         let buffer = self.create_buffer(&wgt_descriptor);
-        {
-            let mut slice = buffer.slice(..).get_mapped_range_mut();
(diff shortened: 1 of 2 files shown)


src/util/device.rs
```diff
@@ -56,10 +56,6 @@ impl DeviceExt for crate::Device {
         {
             let mut slice = buffer.slice(..).get_mapped_range_mut();
```
kvark, line 57: maybe pass `unpadded_size` here and then use `slice` without brackets?

kvark (commented):
let's squash and possibly make the proposed change

src/util/device.rs
grovesNL: Ah yeah I guess we don't even need `slice` here at all

kvark (approved):
Thank you!
bors r+

PR #873, opened 2021-04-22
Boids example: cleaner rand code

dhardy:
... for a certain definition of cleaner. Is this clearer or just more confusing?


kvark (approved):
beautiful!
bors r+

PR #900, opened 2021-05-07
Fix initializing buffer with unaligned data

Uriopass:
In wgpu 0.7, this was a valid operation, the zero padding was done by hand. 
In #872 this logic was removed as an optimization since buffers are already zero initialized, however the slice end was taken from buffer.slice which has to be aligned when mapped.

This is causing issues when trying to update imgui-wgpu-rs (https://github.com/Yatekii/imgui-wgpu-rs/issues/54) to  wgpu 0.8 since they  use U16 index buffers with odd lengths.

### DIFF
--- a/src/util/device.rs
+++ b/src/util/device.rs
@@ -65,9 +65,7 @@ impl DeviceExt for crate::Device {
 
             let buffer = self.create_buffer(&wgt_descriptor);
 
-            buffer
-                .slice(..unpadded_size)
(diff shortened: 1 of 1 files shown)


grovesNL (approved):
Nice, thank you!
bors r+

Uriopass:
That would be great!

kvark:
It's on crates!

PR #905, opened 2021-05-09
Fill Out Issue Templates

cwfitzgerald:
This adds issue templates, as well as links to the wgpu-core tracker and the discussion board. Hopefully this should reduce the amount of issues we have to transfer.


kvark (approved):
Nice, thank you!
bors r+

PR #910, opened 2021-05-14
WIP: OpenXR integration

blaind:
I've made a proof-of-concept for OpenXR rendering on Bevy engine, using wgpu & underlying stack. See https://github.com/blaind/xrbevy

The code in this pull request is not even close to merge-able, but I thought its good idea to open it for further discussion.

See also https://github.com/gfx-rs/gfx/pull/3761 and https://github.com/gfx-rs/wgpu/pull/1387

For architecture of the pull request, see https://github.com/blaind/xrbevy/blob/main/docs/architecture.md (evolving document)


Cargo.toml
```diff
@@ -48,6 +49,7 @@ raw-window-handle = "0.3"
 smallvec = "1"
 tracing = { version = "0.1", default-features = false, features = ["std"] }
 serde = { version = "1", features = ["derive"], optional = true }
+openxr = { version = "0.14", features = ["loaded"], default-features = false, optional = true }
```
blaind, line 52: Is  it okay to pull openxr as optional dependency?

Cargo.toml
```diff
@@ -22,6 +22,7 @@ replay = ["serde", "wgc/replay"]
 # Make Vulkan backend available on platforms where it is by default not, e.g. macOS
 vulkan-portability = ["wgc/gfx-backend-vulkan"]
 webgl = ["wgc"]
+use-openxr = ["wgc/use-openxr", "openxr"]
```
blaind, line 25: Should probably rename the feature `use-openxr` (across crates)

src/lib.rs
```diff
@@ -1586,6 +1603,25 @@ impl Device {
         }
     }
 
+    /// Creates a new [`Texture`] from a raw (Vulkan) image
+    #[cfg(feature = "use-openxr")]
+    pub fn create_openxr_texture_from_raw_image(
```
blaind, line 1608: See https://github.com/Ralith/openxrs/blob/5283daa09ddbe7170069ff3ea9e42e8a8a9ff4bc/openxr/examples/vulkan.rs#L537

lachlansneff:
Fantastic! OpenXR support for wgpu-rs is something I really would like to see.

blaind:
> Fantastic! OpenXR support for wgpu-rs is something I really would like to see.
> 
> I do wonder if it could be implemented by instead exposing internal WebGPU, Vulkan, DX12, etc primitives in wgpu-rs through platform dependent traits, and then having an wgpu-openxr crate that uses those.

For context, I've updated the architecture document. There's quite a bit of different things going on (i/o, swapchain, device handling, etc.) so its a bit of puzzle to find the right places in ecosystem for each. Openxr itself supports vulkan and opengl at least.

kvark:
Project moved to `wgpu`

PR #918, opened 2021-05-29
Specify latest gfx release branch in README

grovesNL:
Specify latest gfx release branch in README, to hopefully avoid people checking out `hal-0.2` then wondering why the patch doesn't seem to do anything.


kvark:
Thanks!
bors r+

PR #920, opened 2021-06-01
Bunnymark example

kvark:


examples/bunnymark/main.rs
```diff
@@ -0,0 +1,359 @@
...
+    fn resize(
+        &mut self,
+        _sc_desc: &wgpu::SwapChainDescriptor,
+        _device: &wgpu::Device,
+        _queue: &wgpu::Queue,
+    ) {
+        //empty
+    }
+
+    /// render is called each frame, dispatching compute groups proportional
```
grovesNL, line 291: Looks like these comments are leftover from boids

examples/bunnymark/main.rs
```diff
@@ -0,0 +1,359 @@
...
+            primitive: wgpu::PrimitiveState {
+                topology: wgpu::PrimitiveTopology::TriangleStrip,
+                ..wgpu::PrimitiveState::default()
+            },
+            depth_stencil: None,
+            multisample: wgpu::MultisampleState::default(),
+        });
+
+        let texture = {
+            let img_data = include_bytes!("icon.png");
```
grovesNL, line 135: We could use the wgpu logo or something here instead

kvark:
This is technically ready. Interestingly, using wgpu logo made it run slower for me.
I checked, and we appear to be texture unit bound on GPU, so the logo is probably just bigger than the other picture.

kvark:
Let's add it with an idea that:
  1. we'll potentially move the examples out into a separate repo. I think this has a benefit of allowing users to copy stuff out of the examples, easier.
  2. we'd extend this example with more features, such as `TypedArena` use.

bors r=grovesNL

